PR #112, opened 2014-12-03
Object-Oriented Geometric Intrinsics Refactorization + OpenCG Compatibility

wbinventor:
This pull request provides a more flexible object-oriented implementation of geometric intrinsics, including ``Materials``, ``Surfaces``, ``Cells``, ``Universes`` and ``Lattices`` than that in the current release version of OpenMOC. This PR is a full departure from the original XML input specification originally used in OpenMOC prior to the current SWIG-generated Python interface. All of the example input files have been updated to reflect the changes made to the geometric intrinsics and are fully functional. The online documentation will be updated soon in another pull request to the ``gh-pages`` branch.

In particular, this refactorization of the geometric intrinsics no longer requires the user to  implicitly define the constructive solid geometry "tree" using the "IDs" of various object. Instead, geometric intrinsics are registered with each other to build up the CSG tree (*e.g.*, adding ``Surfaces`` to ``Cells``, adding ``Cells`` to ``Universes``, etc.). Furthermore, the ``Universes`` are now instantiated directly by the user as an object rather than implicitly defined with an ID given to each ``Cell``. This makes the Python input specification slightly more lengthy, but much more transparent and scalable to complex, large reactor geometries.

In addition, some additions were made to permit the user to extract geometric intrinsics from the CSG tree in Python using some SWIG magic. For example, users may now retrieve Python dictionaries of ``Cells``, ``Universes``, etc. using new routines, a subset of which are below:

* ``Geometry.getAllMaterials()``
* ``Universe.getAllCells()``
* ``Universe.getAllMaterials()``
* ``Universe.getAllUniverses()``
* ``Cell.getSurfaces()``
* ``Cell.getAllCells()``
* ``Cell.getAllUniverses()``

These are helper routines which traverse the CSG tree and return geometric intrinsics to the user in Python. The intrinsics can be modified by the user with the changes being directly reflected in the underlying ``Geometry``. This permits easy parametric optimization studies over various geometries and/or material compositions all from within Python.

This PR also implements a fully functional compatibility module with OpenCG ("Open Combinatorial Geometry"), a tool which will be made open source in the coming months. This is implemented in ``openmoc/compatible/opencg_compatible.py`` and is installed as the ``openmoc.compatible.opencg_compatible`` Python module. In particular, the compatibility module allows users  to build 3D reactor geometries in OpenCG and to easily export them to OpenMOC, and vice versa. This functionality is useful for V&V between reactor physics codes, parametric optimization studies over geometric parameters, etc. The compatiblity module is a corollary to the one currently being reviewed for release with OpenMC in this [pull request](https://github.com/mit-crpg/openmc/pull/331).

A final note is that ``Universe`` at the root node of the CSG tree is no longer required to have ``ID=0``. This constraint on the ID was simply a legacy requirement based on the original input XML specifications when they were modeled based on OpenMC. With this geometry refactorization, this constraint is no longer necessary since a user now "registers" the root node ``Universe`` object with the ``Geometry``.


sample-input/simple-lattice/simple-lattice.py
```diff
@@ -132,10 +156,11 @@
...
-#plotter.plot_flat_source_regions(geometry, gridsize=500)
-#plotter.plot_fluxes(geometry, solver, energy_groups=[1,2,3,4,5,6,7])
+plotter.plot_materials(geometry, gridsize=500)
+plotter.plot_cells(geometry, gridsize=500)
+plotter.plot_flat_source_regions(geometry, gridsize=500)
+plotter.plot_fluxes(geometry, solver, energy_groups=[1,2,3,4,5,6,7])
 
 log.py_printf('TITLE', 'Finished')
 
+u
```
PrezNattyGibbs: Is this a typo?

sample-input/tiny-lattice/tiny-lattice.py
```diff
@@ -56,10 +53,13 @@
 
 log.py_printf('NORMAL', 'Creating cells...')
 
-cells = []
-cells.append(CellBasic(universe=1, material=uo2_id))
-cells.append(CellBasic(universe=1, material=water_id))
-cells.append(CellFill(universe=0, universe_fill=2))
+cells = list()
```
PrezNattyGibbs: For these super simple sample input files, with these changes, there isn't really a need to put cells and surfaces in a list.  It would be more intuitive for new users to just see the cell objects have their own names.

That said, this probably doesn't need to be included in the current PR, I might even volunteer to revamp a few of these in another PR.

PrezNattyGibbs:
I'm just starting to look at this, went through the simple sample input files first.  I just want to say that this change is definitely a substantial improvement and greatly increases the intuitive nature of our CSG implementation.  My work on the geometry clone is almost entirely obsolete now, but redoing it should be much simpler now that we are passing around pointers rather than the nightmare of indices and lists.

Bravo!

sample-input/benchmarks/LRA/LRA-cmfd.py
```diff
@@ -30,12 +30,6 @@
 
 materials = materialize.materialize('LRA-materials.py')
 
-region1 = materials['region_1'].getId()
-region2 = materials['region_2'].getId()
-region3 = materials['region_3'].getId()
-region4 = materials['region_4'].getId()
-region5 = materials['region_5'].getId()
-region6 = materials['region_6'].getId()
```
PrezNattyGibbs: Getting rid of these kinds of lines is awesome.

sample-input/benchmarks/LRA/LRA-cmfd.py
```diff
@@ -82,98 +83,189 @@
...
+            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
+            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
+            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
+            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
+            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
+            [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]]
+
+for i in range(10):
+  for j in range(10):
+    template[i][j] = universes[template[i][j]-1]
```
PrezNattyGibbs: I agree with Sam that the 0-based indexing in the lists compared to the 1-based indexing in the template is a little weird.  I don't see a wildly better option regardless.

sample-input/benchmarks/c5g7/c5g7-cmfd.py
```diff
@@ -46,12 +38,12 @@
...
-planes.append(YPlane(y=32.13))
+circles = list()
+planes = list()
+planes.append(XPlane(x=-32.13, name='left'))
+planes.append(XPlane(x=32.13, name='right'))
+planes.append(YPlane(y=-32.13, name='bottom'))
+planes.append(YPlane(y=32.13, name='top'))
 circles.append(Circle(x=0., y=0., radius=0.54))
 circles.append(Circle(x=0., y=0., radius=0.58))
 circles.append(Circle(x=0., y=0., radius=0.62))
```
PrezNattyGibbs: Why only name the planes?

sample-input/benchmarks/c5g7/c5g7-cmfd.py
```diff
@@ -174,143 +174,226 @@
...
+universes[5].addCell(cells[22])
+universes[5].addCell(cells[23])
+universes[6].addCell(cells[24])
+universes[7].addCell(cells[25])
+universes[8].addCell(cells[26])
+universes[9].addCell(cells[27])
+universes[10].addCell(cells[28])
+universes[11].addCell(cells[29])
+universes[12].addCell(cells[30])
+root.addCell(cells[31])
```
PrezNattyGibbs: My opinion would be to add cells as you create universes, rather than this unintelligible and long list of `addCell` calls after the fact.

Even something like
```
universes.append(Universe(name='uo2 pin cell'))
universes[-1].addCell(cells[0])
universes[-1].addCell(cells[1])
```

sample-input/benchmarks/c5g7/c5g7-cmfd.py
```diff
@@ -174,143 +174,226 @@
...
-     [12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 12, 7, 7, 7, 7, 7, 7],
-     [7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7],
-     [7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7],
-     [7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7],
-     [7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7],
-     [7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7],
-     [7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7]])
-
+lattices.append(Lattice(name='bottom corner reflector'))
+lattices[-1].setWidth(width_x=1.26, width_y=1.26)
```
PrezNattyGibbs: What's the rationale behind moving the width specification to a separate function call rather than leaving it in the constructor?  Does this allow something like changing the pitch to be done parametrically?

sample-input/bundled-lattice/bundled-lattice.py
```diff
@@ -38,13 +35,13 @@
 
 log.py_printf('NORMAL', 'Creating surfaces...')
 
-circles = []
-planes = []
+circles = list()
+planes = list()
```
PrezNattyGibbs: Is `list()` recommended over `[]` in general by some Python standard?

sample-input/bundled-lattice/bundled-lattice.py
```diff
@@ -93,57 +95,100 @@
...
+    [u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5],
+    [u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4],
+    [u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5],
+    [u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4],
+    [u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5],
+    [u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4],
+    [u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5],
+    [u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4],
+    [u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5, u6, u5],
+    [u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4, u5, u4]])
```
PrezNattyGibbs: Are you intentionally doing some of these differently from others or is this just a result of changing your mind after not liking one of them?

There is certainly an argument that can be made for non-uniformity, as users could see multiple options...

src/Cell.cpp
```diff
@@ -91,11 +111,116 @@ cellType Cell::getType() const {
...
+  return _max_z;
+}
+
+
+/**
+ * @brief Return the boundary condition (REFLECTIVE, VACUUM, or INTERFACE) at
+ *        the minimum reachable x-coordinate in the Cell.
+ * @return the boundary condition at the minimum x-coordinate
+ */
+int Cell::getMinXBoundaryType() {
```
PrezNattyGibbs: Why the move from "left" to "minX"?  Are we ramping up for BCs on weird shaped cells?  Or are we just trying to sound more technical?

src/Cell.cpp
```diff
@@ -119,16 +244,23 @@ std::map<int, surface_halfspace> Cell::getSurfaces() const {
...
+ * @brief Sets the name of the Cell
+ * @param name the Cell name string
  */
-void Cell::setUniverse(int universe) {
-  _universe = universe;
+void Cell::setName(const char* name) {
+  int length = strlen(name);
+
+  /* Initialize a character array for the Cell's name */
+  _name = new char[length+1];
```
PrezNattyGibbs: What if the `_name` already exists?  Shouldn't we delete the old contents?

src/Cell.h
```diff
@@ -73,28 +77,96 @@ class Cell {
...
   int _id;
 
+  /** A user-defined name for the Surface */
+  char* _name;
+
   /** The type of Cell (ie MATERIAL or FILL) */
   cellType _cell_type;
 
   /** The ID for the Universe within which this cell resides */
+  // FIXME: Remove this
```
PrezNattyGibbs: Fix the `FIXME` :smiley_cat:

src/Cmfd.cpp
```diff
@@ -45,7 +45,7 @@ Cmfd::Cmfd() {
   _group_indices_map = NULL;
 
   /* Initialize boundaries to be reflective */
-  _boundaries = new boundaryType[4];
+  _boundaries = new int[4];
```
PrezNattyGibbs: I must have missed where this change happened.  Why the change from enum to int?

src/Geometry.cpp
```diff
@@ -110,8 +77,8 @@ double Geometry::getXMin() {
  * @brief Return the maximum x-coordinate contained by the Geometry.
  * @return the maximum x-coordinate (cm)
  */
-double Geometry::getXMax() {
-  return _x_max;
+double Geometry::getMaxX() {
+  return _root_universe->getMaxX();
```
PrezNattyGibbs: Why add this change rather than just storing in the `Geometry` like before?

PrezNattyGibbs:
Alright, I'm done with my review.  I mostly gave a bunch of "Why did you do this?" comments.  Once you enlighten me, I'm ready to give my thumbs up.

sample-input/benchmarks/c5g7/c5g7-cmfd.py
wbinventor: If there was a reason, I don't recall it. Do you want me to name the ```Circles`` as well?

sample-input/benchmarks/c5g7/c5g7-cmfd.py
wbinventor: I like that, I'll make this change in my next commit.

sample-input/benchmarks/c5g7/c5g7-cmfd.py
wbinventor: Yes, the main rationale was to make it easier to parametrically change the lattice pitch. Plus, I think it is best to only set things like IDs, names, etc. in the constructors and require the user to set more substantive attributes through setter routines. This is certainly debatable - if you'd like me to add the ``width_y`` and ``width_x`` parameters back to the constructor, let me know.

sample-input/bundled-lattice/bundled-lattice.py
wbinventor: I've never seen it recommended, but it is more explicit and transparent. I've started using ``list()``, ``dict()``, etc. rather than ``[]`` and ``{}`` in all of my code for that reason.

sample-input/bundled-lattice/bundled-lattice.py
wbinventor: Mainly for non-uniformity, not other reason at all.

sample-input/simple-lattice/simple-lattice.py
wbinventor: Yes, thanks! Just removed it.

sample-input/tiny-lattice/tiny-lattice.py
wbinventor: I agree. I'll revamp the lattice cell indexing, and let you revamp this in your own PR.

src/Cell.cpp
wbinventor: I (eventually) plan to extend our geometry formulation to 3D to allow for full compatibility with OpenCG. The user will just designate some 2D "slice" of the geometry they wish to model with OpenMOC.  In this case, "minX," "minY," and "minZ" will be more explicit and transparent than "left," "bottom," and whatever else we might come up with for z.

src/Cell.cpp
wbinventor: Good catch! I just committed a fix for the ``Cell``, ``Surface``, ``Universe``, ``Lattice``, and ``Material`` classes.

src/Cell.h
wbinventor: Haha, got it.

src/Cmfd.cpp
wbinventor: Actually, this was just me being lazy and not wanting to deal with a circular header dependency that was wreaking havoc. I've dealt with this in my latest commit and reverted everything back to using the ``boundaryType`` enum.

Just for future reference, an enumeration is a distinct type whose value is restricted to one of several explicitly named constants. The values of the constants are values of an integral type known as the underlying type of the enumeration (*e.g.*, an unsigned int).

src/Geometry.cpp
wbinventor: The problem with storing it in the ``Geometry`` is that the user may change a ``Surface`` parametrically - hence changing the bounding box around various ``Cells``, ``Universes``, etc. under the hood. The ``Geometry`` will not know about this unless it traverses the CSG tree to find the appropriate bounding box. This allows geometric parameters to be varied with multiple simulations in a single Python script.

wbinventor:
All of the input files have been refactored to address some of the concerns that came up in the comments. The new versions of the input files for the C5G7 and LRA benchmarks showcase some of the niceties of dynamic Pythonic input files, and as a result, are now over 100 lines shorter than before. What do you think of these changes @PrezNattyGibbs and @samuelshaner?

I also merged in the changes made in PR #115 which changed the way k-effective is computed to account for nu-scatter.

Finally, in the absence of a test suite, I think that @PrezNattyGibbs and @samuelshaner should download this branch and run all of the example input files for a couple of different input parameters just to confirm that everything is running as it should.

PrezNattyGibbs:
I just ran all the sample input files, and everything seems to behave as expected.  I give my thumbs up to accept this PR.  Seeing as @samuelshaner is assigned to the task, I'll let him pull the trigger.

For what it's worth, I found these things in running the sample inputs, but these don't seem to be related to the current PR.  If we care about any of them, I can open an issue.
- simple-lattice, cmfd has plots active, none of the other ones do
- First time I ran homogeneous-two-group, it threw a NaN, but I can't replicate it.  Might be my computer hiccuping, I don't know.
- LRA doesn't converge in 1000 iterations.  It seems silly to have a sample input that doesn't converge.
- c5g7 is 60pcm or so below c5g7-cmfd.  Are we cool with that bias?

wbinventor:
I will remove the CMFD plots from simple-lattce since it does not make since for that line to be there (copy and paste issue).

As for your second two bullets referring to homogeneous-two-group, this is an issue that I've noticed as well. I need to solve this before we merge this PR.

I'm not sure about the LRA problem - @samuelshaner, is this something you know about? I know that with CMFD this problem converges just fine.

Finally, how many angles did you use for C5G7 @PrezNattyGibbs? I don't think that the bias between CMFD and no-CMFD is as large when more angles are used, but I may be wrong. If not, this is certainly not an issue that this PR introduced since it mostly just affects how the geometry is created.

samuelshaner:
@PrezNattyGibbs thanks for pointing those things out and looking over the PR. 

For the LRA problem, I have known about that. You guys are right, we should probably change the LRA input file to have more iterations. Perhaps set the `max_iters` variables to 10,000 in the LRA input files. Either that or increase the `_max_iters` default value in `openmoc/options.py`. I would vote for just setting `max_iters` in the LRA input files.  

The bias between C5G7 with and without CMFD is due to CMFD crushing the convergence of the source while the non-CMFD run is still converging. If you decrease the convergence criteria to 1e-7, the discrepancy goes below 1 pcm. 

A few things I'll point out:
- SWIG gives the following warning during compilation. I'm not sure why, but it might be work checking out.

`/usr/share/swig/1.3.40/python/pyiterators.swg:389: Warning(511): Can't use keyword arguments with overloaded functions (swig::SwigPyIterator::operator -(ptrdiff_t) const).
/usr/share/swig/1.3.40/python/pyiterators.swg:390: Warning(511): Can't use keyword arguments with overloaded functions (swig::SwigPyIterator::operator -(swig::SwigPyIterator const &) const).`
- C5G7 seg faults when the file is run with to `tracks` directory present. I think this might have something to do with setting the directory permissions.

I'm going to go through running all the input files tomorrow.

wbinventor:
I just pushed some code which fixes the NaN issue for the homogeneous-one-group and homogeneous-two-group problems that were pointed out yesterday. This is a really interesting bug which was only seen on the very first run for those two sample problems. Once a track file had been created after the first run, all subsequent runs which loaded tracks from that file converged to the correct answer instead of NaN.

The issue occurred when some track segment had an optical path length equal (or very nearly equal) to the maximum optical path length stored in the exponential interpolation table. In some cases, this allowed for an index to be computed which was just outside of the table. Although I have verified this with my testing, I am still unsure why the code worked just fine after storing all of the tracks to a file.

In any case, my current fix is to make the exponential table just slightly larger than it needs to be to accomodate track segments which may have the max optical path length:

https://github.com/wbinventor/OpenMOC/blob/opencg-compat/src/CPUSolver.cpp#L319

What do you think @samuelshaner and @PrezNattyGibbs?

Note that his appears to have solved the issue for the segmentation fault in C5G7. Frankly, I'm frankly not sure why a segmentation fault was not thrown for homogeneous-two-groups because it should have since the bug allowed the ``Solver`` to read from the outside the bounds of the exponential interpolation table. Perhaps for such a small problem, this happened to be within one of other arrays allocated for scalar or boundary fluxes and so no segmentation fault was thrown. In any case, C5G7 seems to be working on the first run w/o a segmentation fault.

Also, I merged in @PrezNattyGibbs' code from the closed source fork. This included a few patches as well as some hooks for the user to change the maximum optical path length for the exponential interpolation table.

I hope that one or both of you will be willing to run all of the sample input files once more to verify that all is well. In addition, if you would be willing to compile for a GPU machine and run each problem with the ``GPUSolver``, that would be nice QA.

wbinventor:
@samuelshaner - I've just pushed a commit to address the SWIG warning you mentioned in your comment yesterday.

samuelshaner:
@wbinventor - I pulled the changes but I'm getting an error when trying to compile. This shows up on both my laptop and a fresh cloned copy on nsecluster. It does not show up every time I try to compile - sometimes it will compile just fine and others it will throw an error.

The error message on nsecluster is:

byte-compiling /home/shaner/.local/lib/python2.7/site-packages/openmoc/openmoc.py to openmoc.pyc
SyntaxError: ('invalid syntax', ('/home/shaner/.local/lib/python2.7/site-packages/openmoc/openmoc.py', 74, 41, 'def castCellToCellFill(*args, **kwargs) -> "CellFill *" :\n'))
running install_egg_info

The error message on my laptop is:

byte-compiling /Users/sam/Library/Python/2.7/lib/python/site-packages/openmoc/openmoc.py to openmoc.pyc
  File "/Users/sam/Library/Python/2.7/lib/python/site-packages/openmoc/openmoc.py", line 77
    def castCellToCellFill(*args, **kwargs) -> "CellFill *" :
                                            ^
SyntaxError: invalid syntax
running install_egg_info

Have you encountered any compile issues or have any ideas why this would just now be showing up?

config.py
```diff
@@ -266,7 +269,7 @@ class configuration:
   ###########################################################################
 
   # A list of the flags for SWIG
-  swig_flags = ['-c++', '-python', '-keyword'] #, '-keyword', '-py3']
+  swig_flags = ['-c++', '-python', '-keyword', '-py3']
```
samuelshaner: The addition of the `-py3` flag seems to be causing the compile error. Perhaps the `py3` flag could be added only if python 3.0 is being used?

wbinventor:
Thanks for pointing this out @samuelshaner, I definitely didn't notice it before but I did indeed reproduce the issue. I've fixed it with a switch to use the ``-py3`` SWIG flag only when Python 3 is used. In addition, I revised a few small things in the Python ``openmoc.process`` and ``openmoc.materialize`` modules such that they are compatible with Python 3, and confirmed that all of our input files work with both Python 2 and 3.

samuelshaner:
I tested all the input files with a variety of inputs and ran C5G7 on the GPU. Everything looks great so it's time to merge. Thanks for all the work @wbinventor and @PrezNattyGibbs!

PR #115, opened 2014-12-18
Accounting for Nu-Scattering in Keff Calculation

wbinventor:
This pull requests modifies the ``Solver::computeKeff()`` routine to account for (n,2n), (n,3n), etc. reactions through nu-scatter. This functionality allows the user to assign nu-scatter as the scattering matrix for one more more ``Materials``, and for this to be accounted for in the calculation for keff. The old formulation for computing keff was as follows:

equation

The new formulation uses total-nu-scatter instead of absorption in the denominator:

equation

This new formulation was implemented for the ``CPUSolver``, ``VectorizedSolver`` and ``GPUSolver`` subclasses of the abstract ``Solver`` class. 

One item to consider before merging this PR is that the precise value for keff is no longer strictly reproducible for parallel calculations. This seems to only  true in the sub-pcm level. Interestingly enough, it appears to be more of an issue for small problems than for large problems. I only notice this issue when compiling OpenMOC with single precision - I observed no discrepancies when using double precision using the ``--fp=double`` compile time command line argument.

I believe this issue is the result of roundoff error propagation during summations of large arrays. This is the reason we began using [pairwise sum](http://en.wikipedia.org/wiki/Pairwise_summation) throughout the code a year and a half ago. Although the new code still uses pairwise summations of the total, fission and scattering reaction rates, the roundoff error issue is exacerbated due to the fact that I am summing across the group-to-group scattering rates.

One potential fix might be to break up the ``_sigma_s`` attribute for the ``Material`` class into ``_sigma_s`` and ``_sigma_matrix`` attributes. The user would simply need to pass in the scattering matrix to the ``Material::setSigmaS(...)`` routine to be stored in ``_sigma_s_matrix``. Under the hood, the rows of the matrix would be summed up and stored in ``_sigma_s``. Then, when ``Solver::computeKeff()`` is called, the only ``_sigma_s`` would be used rather than the matrix. This *should* reduce roundoff error propagation by implicitly embedding a pairwise sum in into the keff calculation. In addition, it would improve performance, particularly for problems with >100 energy groups.

### DIFF
--- a/config.py
+++ b/config.py
@@ -385,8 +385,6 @@ class configuration:
                 define_macros = self.macros[self.cc][self.fp],
                 swig_opts = self.swig_flags + ['-D' + self.cc.upper()]))
 
-    print('swig opts = {0}'.format(self.swig_flags + ['-D' + self.cc.upper()]))
-
     # The openmoc.cuda extension if requested by the user at compile
     # time (--with-cuda)
     if self.with_cuda:
--- a/config.py
+++ b/config.py
@@ -385,6 +385,8 @@ class configuration:
                 define_macros = self.macros[self.cc][self.fp],
                 swig_opts = self.swig_flags + ['-D' + self.cc.upper()]))
 
+    print('swig opts = {0}'.format(self.swig_flags + ['-D' + self.cc.upper()]))
+
     # The openmoc.cuda extension if requested by the user at compile
     # time (--with-cuda)
     if self.with_cuda:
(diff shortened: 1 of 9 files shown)


wbinventor:
I actually made a mistake in believing that we could use a simple ``_sigma_s`` 1D array to resolve this issue. This actually isn't possible - we need the full scattering matrix to compute the scattering source in each group. I'm going to go ahead and complete this merge now, and we can consider other methods to preserve parallel reproducibility in the future.

PR #116, opened 2015-01-15
OpenMOC Test Suite

luisarose:
This pull request adds the early stages of a Python test suite to OpenMOC.


wbinventor:
Closed since this is now far out-of-date. This may serve as a starting point for a formal test suite if and when we find the time to create one to host here on GitHub.

PR #120, opened 2015-01-26
Removed id from Material object

samuelshaner:
This PR removes the uid from the Material object as it is not currently needed. In the future, we may choose to reincorporate it to be consistent with OpenCG or for other reasons. However, to avoid confusion in the current version of the code, it has been removed.

Minor changes needed to be made in the CUDA code for the GPUSolver since CUDA does not support std::map data types on the device (GPU). A std::map was used in `GPUSolver.cu` to map material IDs to indices in the `_FSR_materials` array of material pointers on the device. Segments on the GPU then identify their corresponding material by an index into the `_FSR_materials` array, and not a material uid.

CPU and GPU input files were tested on an AWS EC2 g2.2xlarge instance with a slightly modified version of the HVM StarCluster AMI, ami-6b211202.

### DIFF
--- a/src/Geometry.cpp
+++ b/src/Geometry.cpp
@@ -739,26 +739,16 @@ void Geometry::subdivideCells() {
 
 /**
  * @brief Compute the number of flat source regions in the Geometry and
- *        initialize arrays for FSR IDs and maps.
+ *        initialize CMFD.
  * @details This method is intended to be called by the user before initiating
  *          source iteration. This method first subdivides all Cells by calling
- *          the Geometry::subdivideCells() method. Then it computes the total
- *          number of FSRs in the Geometry and initializes integer arrays of
- *          maps to Cells and Materials UIDs/IDs indexed by FSR IDs.
+ *          the Geometry::subdivideCells() method. Then it initializes the CMFD
+ *          object. 
  */
 void Geometry::initializeFlatSourceRegions() {
 
   /* Subdivide Cells into sectors and rings */
   subdivideCells();
-
-  /* Assign UIDs to materials */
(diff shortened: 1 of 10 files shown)


src/Geometry.cpp
```diff
@@ -739,27 +739,17 @@ void Geometry::subdivideCells() {
 
 /**
  * @brief Compute the number of flat source regions in the Geometry and
```
wbinventor: Looks like this comment wasn't updated from PR #98 - this routine does not calculate the number of FSRs any longer. I know this isn't related to this PR, but let's go ahead and fix it since it is front and center.

wbinventor:
This PR looks good to me! I can't test it out since both my laptop and workstation are down at the moment, but can do so if you want me to in a day or two once I'm up and running again.

wbinventor:
I just ran all of the test cases and everything looks fine. The only issue I see is that the code stalls for ~1 minute for all of the C5G7 cases when it reports "Converging the source..." but prior to performing power iteration. This is reminiscent of an issue I solved in PR #112 in commit add0c995ccbe56c745b792501ea6b0f594015302

@PrezNattyGibbs - have you had a chance to look this one over?

samuelshaner:
Looks like the C5G7 cases were stalling because I did not initialize `_all_materials` in `Geometry::initializeFlatSourceRegions`. I just readded the initialization of `_all_materials` and the code no longer stalls.

wbinventor:
Is there anything stalling this PR from being merged? @PrezNattyGibbs have you given it a look?

PrezNattyGibbs:
I just looked at it, and ran the input files.  Looks fine to me.

PR #121, opened 2015-01-28
Isotopes

PrezNattyGibbs:
This PR splits the `Material` class into two subclasses, `MacroMaterial` and `IsoMaterial`.  The `Material` class is now an abstract class.

- `MacroMaterial` contains all the same functionality as the previous `Material` class
- `IsoMaterial` defines a `Material` as a set of `Isotopes`
- `Isotopes` are introduced, which contain microscopic cross sections
- Only major change to old functionality is `getSigmaS` has been removed from `Material`
- You should now use `getSigmaSByGroup` and the new `getScatterSource` methods instead

This PR allows the definition of `Materials` to be defined on an isotopic basis, which enables self-shielding methods to be implemented and could even make comparisons to Monte Carlo easier.  Note that `Isotope` is intended to be subclassed in the future for various self-shielding implementations and scattering physics.


samuelshaner:
The build for this PR fails when trying to compile the CUDA code. @PrezNattyGibbs - haver you ever worked with the CUDA code or have an interest in learning? If not, @wbinventor or I can make the changes and PR them into your fork's Isotope branch.

PrezNattyGibbs:
I have no idea how to use CUDA, so I won't be able to do it on my own.  
But I'd be happy to work with you and @wbinventor on this one so I get a 
taste of it.  I'll reach out to find a time that works for us all.

sample-input/benchmarks/homogeneous-one-group/HOG-isotope.py
```diff
@@ -0,0 +1,118 @@
...
+iso.setNumEnergyGroups(1)
+iso.setSigmaA(numpy.array([0.069389522]))
+iso.setSigmaF(numpy.array([0.0414198575]))
+iso.setNuSigmaF(numpy.array([0.0994076580]))
+iso.setSigmaS(numpy.array([0.383259177]))
+iso.setChi(numpy.array([1.0]))
+iso.setSigmaT(numpy.array([0.452648699]))
+
+infinite_medium = IsoMaterial(name='1-group infinite medium')
+infinite_medium.addIsotope(iso,1.0)
```
wbinventor: you might just want to add a comment clarifying what the ``1.0`` represents - a/o, w/o, etc.

sample-input/benchmarks/homogeneous-one-group/HOG-two-isotopes.py
```diff
@@ -0,0 +1,124 @@
...
+iso1.setSigmaT(numpy.array([0.442648699]))
+
+iso2 = Isotope(name='isotope #2')
+iso2.setNumEnergyGroups(1)
+iso2.setSigmaA(numpy.array([0.02]))
+iso2.setSigmaT(numpy.array([0.02]))
+
+infinite_medium = IsoMaterial(name='1-group infinite medium')
+infinite_medium.addIsotope(iso1,1.0)
+infinite_medium.addIsotope(iso2,0.5)
```
wbinventor: ditto

sample-input/pin-cell/PC-isotope.py
```diff
@@ -0,0 +1,274 @@
...
+log.set_log_level('NORMAL')
+
+
+###############################################################################
+###########################   Creating Isotopes    ############################
+###############################################################################
+
+log.py_printf('NORMAL', 'Importing isotope data from HDF5...')
+
+mgxs = h5.File('mgxs-8-groups.h5')
```
wbinventor: just add a comment clarifying that you are importing microscopic cross-sections from HDF5 to populate your ``Isotopes``

sample-input/pin-cell/PC-isotope.py
```diff
@@ -0,0 +1,274 @@
...
+mgxs = mgxs['material']
+
+# O-16 in Fuel
+o16f = Isotope(name='O-16 in Fuel')
+o16f.setNumEnergyGroups(8)
+xs = mgxs['material 10000']['O-16']
```
wbinventor: You should use a stripped down version of the HDF5 data file. In particular, rename the 'material 10000' -> '1' or something simpler. In addition, remove the 'average' and 'std. dev.' groups in place of a single group for each xs.

wbinventor: You can either do this with a simple python script with h5py, or with hdfview.

sample-input/pin-cell/PC-isotope.py
```diff
@@ -0,0 +1,274 @@
...
+
+###############################################################################
+###########################   Creating Materials   ############################
+###############################################################################
+
+log.py_printf('NORMAL', 'Creating materials...')
+
+# Fuel
+fuel_matl = IsoMaterial(name='Fuel')
+fuel_matl.addIsotope(u235,3.7503e-4)
```
wbinventor: ditto a comment specifiying a/o or w/o

sample-input/pin-cell/PC-isotope.py
```diff
@@ -0,0 +1,274 @@
...
+fuel_matl.addIsotope(u238,2.2625e-2)
```
wbinventor: It's pretty nitpicky, but our Style Guidelines do call for a space between function parameters ([link](https://mit-crpg.github.io/OpenMOC/devguide/style_guide.html#whitespace))

src/CPUSolver.cpp
```diff
@@ -113,7 +113,7 @@ FP_PRECISION CPUSolver::getFSRSource(int fsr_id, int energy_group) {
   FP_PRECISION* nu_sigma_f = material->getNuSigmaF();
   FP_PRECISION* chi = material->getChi();
   FP_PRECISION fission_source = 0.0;
-  FP_PRECISION scatter_source = 0.0;
```
wbinventor: Why don't you initialize ``scatter_source`` to zero?

wbinventor: Oh I see now.

src/CPUSolver.cpp
```diff
@@ -652,16 +649,9 @@ FP_PRECISION CPUSolver::computeFSRSources() {
...
-      scatter_source = 0;
-
-      for (int g=0; g < _num_groups; g++)
-        _scatter_sources(tid,g) = material->getSigmaSByGroupInline(g,G)
-                      * _scalar_flux(r,g);
-
-      scatter_source=pairwise_sum<FP_PRECISION>(&_scatter_sources(tid,0),
-                                                _num_groups);
+      /* Compute total scattering source for group G */
+      scatter_source = material->getScatterSource(G,&_scalar_flux(r,0));
```
wbinventor: For uniformity, might it be a good idea to do the same thing for the fission source - add a ``Material::getFissionSource(group, flux)`` routine? Even if this isn't something you foresee using in the future for your self-shielding methods, it might make the code easier to follow.

src/Isotope.cpp
```diff
@@ -0,0 +1,607 @@
...
+int Isotope::getNumEnergyGroups() const {
+  return _num_groups;
+}
+
+
+/**
+ * @brief Return the array of the isotope's total cross-sections.
+ * @return the pointer to the isotope's array of total cross-sections
+ */
+FP_PRECISION* Isotope::getSigmaT() {
```
wbinventor: Are the ``Isotope's`` cross-sections micros or macros? Based on what I saw in your new sample input files, I assume micros. You might want to add a ``@details`` doxygen comment block to these setter methods to explicitly specify what the routine expects.

src/Isotope.cpp
```diff
@@ -0,0 +1,607 @@
...
+ * @param xs the total cross-section (\f$ \Sigma_t [cm^1] \f$)
+ * @param group the energy group
+ */
+void Isotope::setSigmaTByGroup(double xs, int group) {
+
+  if (group <= 0 || group > _num_groups)
+    log_printf(ERROR, "Unable to set sigma_t for group %d for isotope "
+              "%d which contains %d energy groups", group, _uid, _num_groups);
+
+  _sigma_t[group-1] = xs;
```
wbinventor: Any reason why you cast to ``FP_PRECISION`` in the ``setSigmaT(...)`` routine above but do not do so here? You shouldn't need to cast here as the compiler will do it implicitly.

wbinventor: And if you were simply copying from my original ``Material`` class implementation, then I apologize for causing a fuss :-)

src/Isotope.cpp
```diff
@@ -0,0 +1,607 @@
...
+         group, _uid, _num_groups);
+
+  _chi[group-1] = xs;
+}
+
+
+/**
+ * @brief Get the scatter source for an isotope
+ * @param group The energy group
+ * @param flux Pointer to the array of flux values
```
wbinventor: no capitalization in comments

src/Isotope.cpp
```diff
@@ -0,0 +1,607 @@
...
+ * @return The scatter source
+ */
+FP_PRECISION Isotope::getScatterSource(int group, FP_PRECISION* flux) {
+  
+  FP_PRECISION scatter_source = 0;
+  
+  /* loop over origin groups */
+  /* Note: could switch to pairwise sum if needed */
+  for (int g=0; g < _num_groups; g++)
+    scatter_source += getSigmaSByGroup(g+1, group+1) * flux[g];
```
wbinventor: Use a pairwise sum here instead.

src/Isotope.h
```diff
@@ -0,0 +1,118 @@
...
+  void setChiByGroup(double xs, int group);
+  
+  bool isFissionable();
+
+  
+};
+
+
+
+#endif
```
wbinventor: Just add a little comment ``/* ISOTOPE_H */`` here to indicate that this is part of the header guard - see other header files for examples of this standard practice.

src/Material.h
```diff
@@ -43,6 +45,13 @@
 int material_id();
 void reset_material_id();
 
+enum materialType {
+  GENERIC_MATERIAL,
```
wbinventor: Perhaps simplify these as ``GENERIC``, ``MACRO``, ``ISO`` (or ``MICRO``?) and ``SUBGROUP``?

src/Material.cpp
```diff
@@ -49,14 +51,11 @@ Material::Material(int id, const char* name) {
 
   _sigma_t = NULL;
   _sigma_a = NULL;
-  _sigma_s = NULL;
   _sigma_f = NULL;
   _nu_sigma_f = NULL;
   _chi = NULL;
-  _dif_coef = NULL;
```
wbinventor: I assume you remove these since they only make sense for macro materials?

src/Material.cpp
```diff
@@ -663,38 +435,400 @@ void Material::setNumEnergyGroups(const int num_groups) {
...
+/**
+ * @brief Destructor deletes all cross-section data structures from memory.
+ */
+MacroMaterial::~MacroMaterial() {
+  if (_data_aligned) {
+    if (_sigma_s != NULL)
+      MM_FREE(_sigma_s);
+  }
+  
+  else {
```
wbinventor: You also need to deallocate memory for the ``_dif_coeff``, ``_dif_hat``, ``_dif_tilde``, and ``_buckling`` arrays.

src/Material.cpp
```diff
@@ -663,38 +435,400 @@ void Material::setNumEnergyGroups(const int num_groups) {
...
+ */
+void MacroMaterial::setNumEnergyGroups(const int num_groups) {
+  Material::setNumEnergyGroups(num_groups);
+  
+  /* Free old data arrays if they were allocated for a previous simulation */
+
+  /* If data is vector aligned */
+  if (_data_aligned) {
+    if (_sigma_s != NULL)
+      MM_FREE(_sigma_s);
```
wbinventor: You need to deallocate ``_dif_coef``, ``_dif_hat``, ``_dif_tilde`` and ``_buckling`` using ``MM_FREE`` here as well.

src/Material.cpp
```diff
@@ -663,38 +435,400 @@ void Material::setNumEnergyGroups(const int num_groups) {
...
+    delete [] _dif_hat;
+
+  if (_dif_tilde != NULL)
+    delete [] _dif_tilde;
+
+  if (_buckling != NULL)
+    delete [] _buckling;
+
+  /* Allocate memory for data arrays */
+  _sigma_s = new FP_PRECISION[_num_groups*_num_groups];
```
wbinventor: Shouldn't his allocate memory for the diffusion coefficient and buckling? Or was that done elsewhere in the old version?

src/Material.cpp
```diff
@@ -816,7 +951,7 @@ void Material::setSigmaS(double* xs, int num_groups_squared) {
 
   for (int dest=0; dest < _num_groups; dest++) {
     for (int orig=0; orig < _num_groups; orig++)
-      _sigma_s[dest*_num_groups+orig] = xs[orig*_num_groups+dest];
+      setSigmaSByGroupInline(xs[orig*_num_groups+dest],orig,dest);
```
wbinventor: Was there a ``Material::setSigmaSByGroupInline(...)`` routine in the previous implementation? I don't think so, since it doesn't make much sense to have one (it only made sense to have an inline routine for the getter, but even then, we might do away with it to simplify the code).

src/Material.cpp
```diff
@@ -1259,13 +1342,69 @@ void Material::setDifTildeByGroup(double xs, int group, int surface) {
...
+
+/**
  * @brief Checks if the total cross-section for this Material is equal to the
  *        absorption plus scattering cross-sections for all energy groups.
  * @details If the total cross-section does not equal the absorption plus
  *          scattering cross-section within SIGMA_T_THRESH then this method
  *          exits OpenMOC.
  */
-void Material::checkSigmaT() {
+void MacroMaterial::checkSigmaT() {
```
wbinventor: Do you think we should make the message in ``MacroMaterial::checkSigmaT()`` a ``WARNING`` message instead of an ``ERROR`` message? This seems to be more inline with where you and I are taking the code, as well was with PR #115's inclusion of code to account for nu-scatter.

src/Material.cpp
```diff
@@ -1313,7 +1452,7 @@ void Material::checkSigmaT() {
  *          scattering cross-sections and chi for all energy groups.
  * @return character array of this Material's attributes
  */
-std::string Material::toString() {
+std::string MacroMaterial::toString() {
```
wbinventor: Perhaps some of the ``MacroMaterial::toString()`` routine should be implemented in the ``Material`` parent class? Since some of the arrays are attributes of ``Material``, it should have its own ``toString()`` routine to print those items. The subclasses can all call the parent class and  then concatenate their own substrings to the end to create a super string.

src/Material.cpp
```diff
@@ -1394,28 +1533,17 @@ void Material::printString() {
...
   if (_data_aligned)
     return;
-
-  if (_num_groups <= 0)
-    log_printf(ERROR, "Unable to align Material %d data since the "
-               "cross-sections have not yet been set\n", _id);
-
-  _num_vector_groups = (_num_groups / VEC_LENGTH) + 1;
+    
+  Material::alignData();
```
wbinventor: It doesn't appear that this routine is doing what it is intended to do. Am I missing something?

src/Material.cpp
```diff
@@ -1518,3 +1619,230 @@ Material* Material::clone(){
...
+  if (isotope->isFissionable())
+    _fissionable = true;
+  
+  _num_isotopes++;
+}
+
+
+/**
+ * @brief Get specific number density from Material
+ * @param index The index of the material to be retrieved
```
wbinventor: I think this is a typo: "material" -> "isotope". Also, the isotope isn't "retrieved", only its number density is, correct?

src/Material.cpp
```diff
@@ -1518,3 +1619,230 @@ Material* Material::clone(){
...
+ * @return the number density
+ */
+FP_PRECISION IsoMaterial::getNumberDensity(int index) {
+  if (index > _num_isotopes-1)
+    log_printf(ERROR,"Cannot return Isotope with index %d because there are "
```
wbinventor: Ditto typo in the print statement.

src/Material.cpp
```diff
@@ -1518,3 +1619,230 @@ Material* Material::clone(){
...
+
+/**
+ * @brief Get vector of Isotopes from Material
+ * @return vector of Isotopes
+ */
+std::vector<Isotope*> IsoMaterial::getIsotopes() {
+  return _isotopes;
+}
+  
+
```
wbinventor: I think you should add a ``IsoMaterial::getIsotopeIndex(Isotope* isotope)`` routine - this will be helpful in scenarios where the index for each isotope is not being kept track of by the user (e.g., automated parameter sweeps).

src/Material.cpp
```diff
@@ -1518,3 +1619,230 @@ Material* Material::clone(){
...
+/**
+ * @brief Checks if the total cross-section for this Material is equal to the
+ *        absorption plus scattering cross-sections for all energy groups.
+ * @details If the total cross-section does not equal the absorption plus
+ *          scattering cross-section within SIGMA_T_THRESH then this method
+ *          exits OpenMOC.
+ */
+void IsoMaterial::checkSigmaT() {
+  if (_num_groups == 0)
+    log_printf(ERROR, "Unable to verify Material %d's total cross-section "
```
wbinventor: Ditto my earlier comment - perhaps we want to make these ``WARNING`` rather than ``ERROR`` messages.

src/Material.cpp
```diff
@@ -1518,3 +1619,230 @@ Material* Material::clone(){
...
+  
+  return Q;
+}
+
+
+/**
+ * @brief Get the group to group scatter cross section for given groups
+ * @details Sums the contribution from each isotope
+ * @param origin the column index of the scattering matrix
+ * @param destination the row index of the scatteirng matrix
```
wbinventor: typo: scatteirng -> scattering

src/Material.cpp
```diff
@@ -1518,3 +1619,230 @@ Material* Material::clone(){
...
+
+
+/**
+ * @brief Create a duplicate of the IsoMaterial.
+ * @return a pointer to the clone
+ */    
+IsoMaterial* IsoMaterial::clone() {
+  IsoMaterial* clone = new IsoMaterial(getId(),getName());
+  
+  /* Add each isotope in original Material to cloned Material */
```
wbinventor: The ``IsoMaterial`` inherits a number of items from ``Material``. Why are they not cloned here?

src/Material.h
```diff
@@ -153,29 +193,30 @@ class Material {
...
   void setDifHat(double* xs, int num_groups);
   void setDifTilde(double* xs, int num_groups);
+  void setBuckling(double* xs, int num_groups);
 
   void setSigmaTByGroup(double xs, int group);
   void setSigmaAByGroup(double xs, int group);
   void setSigmaFByGroup(double xs, int group);
   void setNuSigmaFByGroup(double xs, int group);
   void setSigmaSByGroup(double xs, int origin, int destination);
+  void setSigmaSByGroupInline(double xs, int origin, int destination);
```
wbinventor: As I mentioned before, I'm not sure how relevant an inline setter routine is here.

src/Material.h
```diff
@@ -190,9 +231,72 @@ class Material {
...
+  std::vector<Isotope*> _isotopes;
+  
+  /** The vector of number densities associated with the Isotopes */
+  std::vector<FP_PRECISION> _num_dens;
+  
+  /** The number of Isotopes in the Material */
+  int _num_isotopes;
+  
+  /** A flag for whether the fission spectrum has been set */
+  bool _chi_set;
```
wbinventor: What is this used for? I saw that it is set in the ``IsoMaterial::addIsotope(...)`` routine but didn't see it used anywhere else.

wbinventor:
@PrezNattyGibbs - I just finished looking over this PR and it looks pretty good. I think the separation into different ``Material`` subclasses with ``Isotopes`` may not only be useful for your self-shielding work but my work as well. There are a few items I think that we need to address for QA before merging:

* We should merge PR #120 first since it is simpler - it should be easy to merge into this one
* Test ``GPUSolver`` - we can discuss this further per our earlier conversation
* Test ``VectorizedSolver`` with the Intel compiler. You will need to make the same (or similar) modifications that you made to ``CPUSolver`` to the ``VectorizedSolver``. Although we do not typically use this ``Solver`` subclass, we've kept it around as inspiration and a "launchpad" for anyone who decides to come along and properly vectorize OpenMOC, since my implementation from two years ago never achieved any speedup from auto-vectorization. The vectorization relies on the ``MacroMaterial`` to properly align its data (i.e., the ``MacroMaterial::alignData()`` routine), hence we need to ensure that your changes did not affect this process. The ``VectorizedSolver`` must be built with Intel's ``icpc`` compiler and MKL. This can be done as follows:
  * compile with ``--with-icpc`` command line flag
  * Change one or more benchmarks by importing ``openmoc.intel.single`` rather than ``openmoc``
  * Use the ``VectorizedSolver`` rather than ``CPUSolver``
  * Verify the result with respect to the ``CPUSolver`` - it will differ a slight bit since the order of operations is different with SIMD'ization
* Test with both single and double precision - *e.g.*, use the ``--with-sp`` and ``--with-dp`` build options.

src/Material.cpp
samuelshaner: The diffusion coefficients and buckling are only allocated if their getter or setter is called. Materials for the MOC solver don't necessarily need a diffusion coefficient or buckling. It they do have one, they are used to modify the absorption cross section. Furthermore the materials used for MOC don't need a `_dif_hat` or `_dif_tidle`; the materials used in the CMFD solver will need a `_dif_coef`, `_dif_hat`, and `_dif_tilde`.

One option to avoid any confusion would be to subclass `MacroMaterial` as `CMFDMaterial`, but that should probably be done in another PR.

src/Isotope.cpp
PrezNattyGibbs: I just added microscopic all over the place in the `@brief` blocks.

sample-input/benchmarks/homogeneous-one-group/HOG-isotope.py
PrezNattyGibbs: I added that.  When I make documentation, I'll make sure to note it there.  Right now, I'm just supporting number density in a/b-cm.  Someone (probably me) can implement other units in a future PR.

sample-input/pin-cell/PC-isotope.py
PrezNattyGibbs: Yes, sir.

src/CPUSolver.cpp
PrezNattyGibbs: I like the thought, but the implementation is a little tricky, as we precompute the total fission source and then multiply by chi.  So maybe the routine should be to compute the total fission source (not group specific) and then just multiply by chi in the solver function?

src/Isotope.cpp
PrezNattyGibbs: Fixed it.  But yeah, it was your fault :stuck_out_tongue_winking_eye:

src/Material.cpp
PrezNattyGibbs: Yeah, they are in the MacroMaterial constructor now.

src/Material.cpp
PrezNattyGibbs: I stuck with what was previously in the code using `delete []`.  Should I switch to `MM_FREE`?  I don't really know the difference.

src/Material.cpp
PrezNattyGibbs: No, there wasn't.  I can get rid of it.  The inline certainly isn't needed.  I was just trying to avoid putting the logic in too many places.  But I can call the non-inline version.

src/Material.cpp
PrezNattyGibbs: I like this idea.

src/Material.cpp
PrezNattyGibbs: This seems reasonable, but I'd prefer to leave it for a future PR, unless you think it's really important.

src/Material.cpp
PrezNattyGibbs: I certainly haven't tested if it works, but I don't see what you mean.

src/Material.cpp
PrezNattyGibbs: Good call.

src/Material.cpp
PrezNattyGibbs: All of the items it inherits are populated by adding the 'Isotopes'.  The only questionable part is the aligned data, which we can talk about when we fix the `GPUSolver` and test the SIMD solver.

src/Material.h
PrezNattyGibbs: Chi is an issue when you are using more than one isotope in multigroup methods.  My (admittedly not wonderful) solution is to just use the chi from the first fissionable isotope added.  This flag makes that possible.

I'm considering moving chi to be an `IsoMaterial` property rather than pulling from the individual `Isotopes`, but I'll leave it as is for now.

sample-input/pin-cell/PC-isotope.py
PrezNattyGibbs: I clearly was lazy before, but I agree this is a much better call.

src/CPUSolver.cpp
wbinventor: I like that idea. Is that too much to ask for this PR?

src/Isotope.cpp
wbinventor: That's what I thought, thanks for keeping me honest (and fixing it!).

src/Material.cpp
wbinventor: You only need to use ```MM_FREE`` if the data is aligned (*e.g.*, within the ``if(_data_aligned) { ... }`` block. Otherwise, it is fine to used ``delete []``. The constructor has to be able to handle both cases - aligned or not - hence the two different blocks here.

src/Material.cpp
wbinventor: Great!

src/Material.cpp
wbinventor: I think this should be pretty easy to implement. This will fall under the rug after this PR, so I think it's best to get the implementation for it right (as generalized as possible) with this PR.

src/Material.cpp
wbinventor: After looking at this a second time, I think I see what you are doing. You've abstracted most of this routine to the ``Material::alignData()`` routine, and are simply memory aligning the scattering matrix here. This should work fine. You'll confirm if this is indeed the case when you implement the new material implementation in the ``VectorizedSolver``, since it is the only entity which calls the ``Material::alignData()`` routine.

src/Material.cpp
wbinventor: Ok I see.

src/Material.h
wbinventor: Why would you move chi from the ``Isotopes`` to the ``IsoMaterial``? It seems logical to me to keep it isotopic. And related to one of my other comments, perhaps instead of a ``Material:getFissionSource()`` routine you could use a ``Material::getChiFissionSource()`` routine. In the case of ``MacroMaterials``, this would use the macroscopic chi, or in the case of ``IsoMaterials``, it would use each ``Isotopes`` chi separately. This would have the effect of abstracting chi from the ``Solver's`` and making its routines simpler and more generalizable for different material formulations.

src/CPUSolver.cpp
wbinventor: See my related comment on the ``IsoMaterial``.

wbinventor:
I just merged in PR #120, so make sure to merge it in here. Let @samuelshaner and I know when you want some help integrating these changes into the ``GPUSolver`` and ``VectorizedSolver``.

src/Material.h
PrezNattyGibbs: Using a single chi for an isotopic mixture is a pain is what I meant.  But actually computing isotopic fission rates is reasonable and avoids approximation, and we can always make subclasses if we don't want to do that.

PrezNattyGibbs: Actually, this is going to be tricky given the needs of CMFD.  Let's discuss this when we fix up the `GPUSolver`, etc.

geogunow:
What is the status of this PR?

PrezNattyGibbs:
It's on the back burner for now.  In my more recent work, I've found that there are a few other options for `Isotope` implementation, and I'd like to spend a bit more time doing some research before finalizing it.  The stuff in the PR is fine, and would just be expanded upon, but it needs some work to be consistent with `GPUSolver` and `VectorizedSolver`.  I'm not sure what the best thing to do with it now, whether it should be closed and reopened at a later date, if someone should convince me to worry about the `GPUSolver` stuff sooner than later, or if we're happiest just leaving this floating here for now.

wbinventor:
Before we lose you @PrezNattyGibbs, is this a PR that should die an early death?

PrezNattyGibbs:
This is something that should be worked on at some point by me, but it 
won't be before I graduate, and who knows if it'll happen afterwards.  
Can you archive PRs?  We could just kill it and hope I try again later.  
It's SUPER out of date at this point.

wbinventor:
Yeah it is way out-of-date. If you can see yourself returning to this with a fresh mind post-PhD, then I'd say let's kill this PR and you can open a new one if and when you are ready. The PR will always be archived, and you can always reopen a PR, it just won't clutter our backlog of open PRs. I'll let you decide what to do, just wanted to see what plans if any remained for an isotopic model.

wbinventor:
I guess you're off the hook for this now @PrezNattyGibbs :-) Closing this for now, perhaps we'll see it come together another day.

PR #125, opened 2015-03-18
Logger module compatibility with IPython Notebook

wbinventor:
This PR presents a few minor modifications to OpenMOC's logger module: 

First, a few changes were made to ``log.h`` and ``log.cpp`` to ensure that log messages are printed to the inline IPython Notebook console.

Second, messages labeled with an ``ERROR`` tag no longer call the ``exit(1)`` routine. This was originally implemented when OpenMOC was a binary executable; now that OpenMOC is a shared library which is intended to be used as a "plug-n-play" Python package, it no longer makes sense to "exit" the program when an error condition is met. A particular case in point is when a user is using OpenMOC in an IPython Notebook - previously, an error message would cause the entire Notebook kernel to crash. Now, the user  will simply see the error message printed to the inline console and may continue to revise the Notebook.

### DIFF
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -462,7 +462,7 @@ void log_printf(logLevel level, const char* format, ...) {
     if (level == ERROR)
       throw std::logic_error(msg_string.c_str());
     else 
-     std::cout << msg_string;
+      printf("%s", msg_string.c_str());
   }
 }
 
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -412,7 +412,7 @@ void log_printf(logLevel level, const char* format, ...) {
         vsprintf(message, format, args);
         va_end(args);
         std::string msg = std::string(message);
-        std::string level_prefix = "[  ERROR  ]  ";
+        std::string level_prefix = "";
 
         /* If message is too long for a line, split into many lines */
         if (int(msg.length()) > line_length)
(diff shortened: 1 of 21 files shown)


src/log.cpp
```diff
@@ -459,13 +459,8 @@ void log_printf(logLevel level, const char* format, ...) {
...
     /* Write the log message to the shell */
-    std::cout << msg_string;
+    printf(msg_string.c_str());
   }
-
-
-  /* If the message was a runtime error, exit the program */
-  if (level == ERROR)
-    exit(1);
 }
```
samuelshaner: It would still be nice for the program to stop execution after a log message with `ERROR` has been called. Have you looked into alternative ways of interrupting the program execution without destructing objects and killing the ipython kernel? I looked around but couldn't find a gracefully way to stop the execution without killing the kernel.

wbinventor: I did look around but ran into the same issue. Perhaps I should post a question about this on StackOverflow or to the SWIG mailing list.

wbinventor: I'll watch my StackOverflow (post)[http://stackoverflow.com/questions/29371292/exit-swig-wrapped-c-c-routine-without-killing-ipython-kernel] over the next few days to see if any SWIG wizards out there can help with this.

samuelshaner: Sounds good! Hopefully a clean solution exists.

wbinventor: I've corrected this issue per a suggestion from the SWIG mailing list. Now when an ``ERROR`` message is printed, it throws a ``logic_error(...)`` exception rather than calling ``exit()``. This allows the IPython Notebook kernel to continue running, but immediately terminates the execution of whatever routine was previously running. Alternatively, it will kill a process running a Python script as is currently done.

wbinventor:
Closing to move to the "develop" branch.

Issue #124, opened 2015-03-18
Universe function getAllUniverses does not include current universe


samuelshaner:
When trying to run a pin cell problem with only one universe (the root universe) and trying to divide the fuel into rings and sectors and moderator into sectors, rings and sectors are not created. In the `Geometry::subdivideCells` method, the `Universe::getAllUniverses` method is called on the root universe. This function will return an empty map so the pin cell will not be broken up into rings and sectors. This could be fixed by adding the following line to the `Universe::getAllUniverses` method:

`universes[_id] = this;`

wbinventor:
@samuelshaner is this a bug you plan to take care of in a future PR?

samuelshaner:
Yes, I'll take care of it. I'll get to it next week.

wbinventor:
Fixed in PR #126.

PR #126, opened 2015-03-31
Force Universe's getAllUniverses method to include current universe

samuelshaner:
Fixed issue #124 to force Universe's getAllUniverses method to include current universe.


wbinventor:
This change looks good to go.

PR #127, opened 2015-03-31
Fix for Spatial Zones with Zero Total Cross-Sections

wbinventor:
This pull request introduces a fix which allows regions with a zero total cross-section to be modeled without leading to NaNs. In the current version of OpenMOC, a zero total cross-section in one or more energy groups in one or more regions will lead to NaNs when evaluating the reduced source and scalar flux terms. On the second power source iteration, the NaNs are propagated throughout, leading to keff to be evaluated to NaN.

There is no perfectly straightforward way to deal with zero total cross-section regions. The quick and dirty approach I came up with was to simply implement conditional checks to assign the reduced source and scalar flux to zero (not physical) in regions / energy groups with a zero total cross-section. Considering that this is a bit of a hacky patch for some problems I've been trying to solve, I'd especially like to hear @samuelshaner and @PrezNattyGibbs thoughts on this since this certainly may be the wrong approach altogether.

This PR also resets the NVIDIA CUDA directory to "/usr/local/cuda" from "/usr/local/cuda-5.5" in ``config.py`` since the latest version of CUDA is 7.0 which is installed by default in "/usr/local/cuda".

### DIFF
--- a/src/Material.h
+++ b/src/Material.h
@@ -17,11 +17,11 @@
 #endif
 
 /** A negligible cross-section value to over-ride user-defined
- *  cross-sectionsvery near zero (e.g., within (-1E-10, 1E-10)) */
+ *  cross-sections very near zero (e.g., within (-1E-10, 1E-10)) */
 #define ZERO_SIGMA_T 1E-10
 
 #ifdef ICPC
-/** Word-aligned memory deallocationallocation for Intel's compiler */
+/** Word-aligned memory allocation for Intel's compiler */
 #define MM_FREE(array) _mm_free(array)
 
 /** Word-aligned memory allocation for Intel's compiler */
--- a/src/Material.h
+++ b/src/Material.h
@@ -16,7 +16,8 @@
 #include "log.h"
 #endif
 
-/** The minimum allowable total cross-section value for any group */
+/** A negligible cross-section value to over-ride user-defined
+ *  cross-sectionsvery near zero (e.g., within (-1E-10, 1E-10)) */
 #define ZERO_SIGMA_T 1E-10
 
 #ifdef ICPC
--- a/src/Material.h
+++ b/src/Material.h
@@ -16,6 +16,9 @@
 #include "log.h"
 #endif
 
+/** The minimum allowable total cross-section value for any group */
+#define ZERO_SIGMA_T 1E-10
+
 #ifdef ICPC
 /** Word-aligned memory deallocationallocation for Intel's compiler */
 #define MM_FREE(array) _mm_free(array)
--- a/config.py
+++ b/config.py
@@ -247,7 +247,7 @@ class configuration:
   library_directories['gcc'] = [usr_lib]
   library_directories['icpc'] = [usr_lib]
   library_directories['bgxlc'] = [usr_lib]
-  library_directories['nvcc'] = [usr_lib, '/usr/local/cuda-5.5/lib64']
+  library_directories['nvcc'] = [usr_lib, '/usr/local/cuda/lib64']
 
 
   #############################################################################
@@ -261,7 +261,7 @@ class configuration:
   include_directories['gcc'] = list()
   include_directories['icpc'] = list()
   include_directories['bgxlc'] = list()
-  include_directories['nvcc'] = ['/usr/local/cuda-5.5/include']
+  include_directories['nvcc'] = ['/usr/local/cuda/include']
 
 
   ###########################################################################
--- a/src/CPUSolver.cpp
+++ b/src/CPUSolver.cpp
@@ -667,8 +667,11 @@ FP_PRECISION CPUSolver::computeFSRSources() {
       fsr_fission_source += fission_source * chi[G];
 
       /* Set the reduced source for FSR r in group G */
-      _reduced_sources(r,G) = (fission_source * chi[G] + scatter_source) *
-                      ONE_OVER_FOUR_PI / sigma_t[G];
+      if (sigma_t[G] > 0)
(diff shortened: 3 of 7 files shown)


samuelshaner:
The one concern I have with this approach is it produces a zero flux in any region and group that has a zero total cross section. If you run a pin cell problem with vacuum (e.g. zero total cross sections for all groups) in the gap, the flux goes to zero in the gap. I would propose treating zero total cross sections as very small pure absorbers. You can set the total cross section to something low like 1.e-10 for the purposes of accumulating the attenuated angular flux to the scalar flux and then dividing by that same small total xs in the `addSourceToScalarFlux` method. I tested that out for a pin cell problem and it seemed to work just fine. If the total xs is sufficiently small, neglecting the contribution of the absorption to keff should not have any effect on keff or the solution to the rest of the problem. Alternatively, the cross sections of the material could just be adjusted. Does that make sense? What do you think about that?

wbinventor:
I recognized that the scalar flux override wasn't physical, but wasn't sure what to do about it. I'm fine with adding 1E-10 to the cross-sections in the ``Material`` cross-section data setters instead, pending @PrezNattyGibbs thoughts.

wbinventor:
I revised this PR such that the ``Material`` now sets the total cross-section to a minimum of ``ZERO_SIGMA_T`` (defined in "src/Material.h) to be 1E-10) in the ``Material.setSigmaT(...)`` routine. This will occur for the cross-section in any energy group with a user-defined value of < ``ZERO_SIGMA_T``. Of course ``ZERO_SIGMA_T`` likely should be defined based on the geometry/materials, so 1E-10 is not necessarily generalizable to any input configuration. Along with this change, I removed my initial implementation which set the reduced sources and scalar fluxes (erroneously) to zero.

PrezNattyGibbs:
Here's my take--
Setting the flux to zero is definitely nonphysical, and I don't really 
like that option.  Setting the reduced source to zero is not a problem 
for areas with very low number densities and thus near zero cross 
sections.  However, if you want to do something crazy like having 
nonzero cross sections in some groups and zero cross sections in other 
groups (I can't think of how this is physical), then the reduced source 
wouldn't actually be zero.  But where we actually use reduced sources in 
computing \Delta \Psi, the limit is still \Delta \Psi = 0 because of the 
exponential term.  So I think setting the reduced source to zero is a 
perfectly fine way to do it, and just let the flux do it's thing as it 
usually does in a calculation.

I don't want to add a small cross section in all regions, but if it's 
just a minimum, and only effects the zero cross section regions, I'm 
fine with that.

--Nathan

wbinventor:
Unfortunately the flux is divided by the total cross-section as well. So you actually can't get away with just setting the reduced source to zero and letting the "flux do it's thing as it usually does in a calculation." I don't see an issue with the minimum zero total cross-section in all regions, so I think that's the only viable option.

wbinventor:
Also fyi, I agree that there *should* be very few situations where this is needed (if the total cross-sections are not erroneously defined by the user). However, I frequently encounter zero total cross-sections when generating my multi-group cross-sections with Monte Carlo (OpenMC). Some of the cross-sections can use efficient tracklength tallies, but others require analog tallies (*e.g.*, transport corrected total xs). As a result, the total cross-section in nearly void regions such as the gap may come out as zero unless a large number of particle histories are modeled. Rather than manually overriding zero cross-sections in my OpenMC-to-OpenMOC cross-section pipeline, I thought it would be useful to simply have the ability to deal with such material data directly in OpenMOC.

PrezNattyGibbs:
I meant that there shouldn't be cases where a group's cross section is 0 
but another group in the same material is nonzero.  The existence of 
zero cross sections makes total sense.  In a real reactor, the gap makes 
the most sense, but I could even envision some other things like 
inserting voided rods into the instrument tubes, etc.

Looking at the equations again, I'm not sure I even know what we want 
OpenMOC to spit out.  In the equations on 
https://mit-crpg.github.io/OpenMOC/methods/method_of_characteristics.html , 
the scalar flux is given in Eq. 40.  I argued that Q = 0, as there was 
no reaction rate in higher groups to generate a scattering source 
there.  But with zero cross sections, \Delta\Psi should also be zero.  
But the real limit is a nonzero flux, so we'd have to do more analysis 
to figure out that analytic limit if we were doing this rigorously.

After this nice refresher in the MOC equations, if the minimum cross 
section solution gives you the expected results, I say go with it.

src/Material.h
```diff
@@ -16,6 +16,9 @@
 #include "log.h"
 #endif
 
+/** The minimum allowable total cross-section value for any group */
```
samuelshaner: Since negative sigma_t are allowed, I would recommend changing this comment to reflect that.

wbinventor: Done

samuelshaner: Sorry to be nit picky, but could you change this comment in `Material.h` as well?

wbinventor: Good call - I revised the comment for ``ZERO_SIGMA_T`` in "Material.h". Let me know if you think it makes sense.

src/Material.h
```diff
@@ -16,6 +16,10 @@
 #include "log.h"
 #endif
 
+/** A negligible cross-section value to over-ride user-defined
+ *  cross-sectionsvery near zero (e.g., within (-1E-10, 1E-10)) */
```
samuelshaner: Sounds great, except for the missing space in `cross-sectionsvery`.

wbinventor:
I updated this PR with a ``WARNING`` message to inform the user if a zero-valued cross-section is overridden with ``ZERO_SIGMA_T`` (1E-10). This is simply to make sure that this adjustment of the cross-sections is done in a transparent way.

PR #128, opened 2015-04-01
Fix for Multi-line Messages in Logger

wbinventor:
This PR introduces a fix for multi-line message printing to the console from the ``openmoc.log`` module. The current implementation had a bug which stripped the leading character from each successive line printed to the console for multiline messages (*e.g.*, messages > 80 characters). In addition, this PR updates the ``Cell.toString()`` routine to include the ``Surface`` halfspace when appending the ``Surface`` IDs to the string representation of each ``Cell``.


samuelshaner:
Testing on Mac and Linux and everything looks good!

PR #129, opened 2015-04-01
Logger module compatibility with IPython Notebook

wbinventor:
This PR is from #125 but on the "develop" branch.

This PR presents a few minor modifications to OpenMOC's logger module:

First, a few changes were made to log.h and log.cpp to ensure that log messages are printed to the inline IPython Notebook console.

Second, messages labeled with an ERROR tag no longer call the exit(1) routine. This was originally implemented when OpenMOC was a binary executable; now that OpenMOC is a shared library which is intended to be used as a "plug-n-play" Python package, it no longer makes sense to "exit" the program when an error condition is met. A particular case in point is when a user is using OpenMOC in an IPython Notebook - previously, an error message would cause the entire Notebook kernel to crash. Now, the user will simply see the error message printed to the inline console and may continue to revise the Notebook.

### DIFF
--- a/openmoc/compatible/opencg_compatible.py
+++ b/openmoc/compatible/opencg_compatible.py
@@ -475,7 +475,10 @@ def get_compatible_opencg_cells(opencg_cell, opencg_surface, halfspace):
 
 def make_opencg_cells_compatible(opencg_universe):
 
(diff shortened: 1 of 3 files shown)


samuelshaner:
Tested in out on Mac and Linux and everything looks good!

PR #130, opened 2015-04-06
Parse convergence data

wbinventor:
This PR introduces the new ``parse_convergence_data(...)`` routine to the ``openmoc.process`` module. This routine will parse an OpenMOC log file and extract the eigenvalues and source residuals at each iteration of an eigenvalue simulation. This data can then be analyzed or plotted to better understand the convergence in a particular simulation.

For example, this routine can be used to plot the eigenvalue convergence for a pure infinite medium problem as follows:

k-eff-conv

Similarly, the source convergence appears as follows:

res-conv


wbinventor:
Closing to open on develop instead of master branch.

PR #131, opened 2015-04-06
Parse convergence data

wbinventor:
This PR introduces the new ``parse_convergence_data(...)`` routine to the ``openmoc.process`` module. This routine will parse an OpenMOC log file and extract the eigenvalues and source residuals at each iteration of an eigenvalue simulation. This data can then be analyzed or plotted to better understand the convergence in a particular simulation.

For example, this routine can be used to plot the eigenvalue convergence for a pure infinite medium problem as follows:

k-eff-conv

Similarly, the source convergence appears as follows:

res-conv

### DIFF
--- a/openmoc/process.py
+++ b/openmoc/process.py
@@ -620,8 +620,8 @@ def parse_convergence_data(filename, directory=''):
               'not an existing OpenMOC log file'.format(filename))
 
   # Compile regular expressions to find the residual and eigenvalue data
-  res = re.compile('res = ([0-9].[0-9]+E[+|-][0-9]+)')
-  keff = re.compile('k_eff = ([0-9]+.[0-9]+)')
+  res = re.compile(b'res = ([0-9].[0-9]+E[+|-][0-9]+)')
+  keff = re.compile(b'k_eff = ([0-9]+.[0-9]+)')
 
   # Parse the eigenvalues
   with open(filename, 'r+') as f:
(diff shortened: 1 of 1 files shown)


openmoc/process.py
```diff
@@ -595,3 +592,57 @@ def restore_simulation_state(filename='simulation-state.h5',
...
+              'not an existing OpenMOC log file'.format(filename))
+
+  # Compile regular expressions to find the residual and eigenvalue data
+  res = re.compile('res = ([0-9].[0-9]+E[+|-][0-9]+)')
+  keff = re.compile('k_eff = ([0-9]+.[0-9]+)')
+
+  # Parse the eigenvalues
+  with open(filename, 'r+') as f:
+    data = mmap.mmap(f.fileno(), 0)
+    eigenvalues = keff.findall(data)
```
samuelshaner: This fails in python 3 with:

>>>  process.parse_convergence_data('openmoc-04-06-2015--13:04:34.log', 'log')
  File "<stdin>", line 1, in <module>
  File "/home/shaner/.local/lib/python3.4/site-packages/openmoc/process.py", line 629, in parse_convergence_data
    eigenvalues = keff.findall(data)
TypeError: can't use a string pattern on a bytes-like object

wbinventor: Thanks for the comment - I'll look into it. I should probably start using Python 3 exclusively for my work to avoid these sorts of issues.

wbinventor: I just pushed what I believe is the fix for this in Python 3. It works on my machine for both Python versions 2.7.3 and 3.4.1.

PR #132, opened 2015-04-06
Plot FSR fluxes vs. energy

wbinventor:
This PR introduces the new ``plot_energy_flux(...)`` routine to the ``openmoc.plotter`` module. This routine can be used to plot the multi-group fluxes vs. energy for one or more flat source regions.  For example, this routine would create a plot of the flux for a 70-group infinite medium problem that might appears as follows:

flux-fsr-0

This PR also breaks backwards compatibility by renaming the ``plot_fluxes(...)`` routine to ``plot_spatial_flux(...)``. This change was made such that the two flux plotting routines would be more explicit as to their domain of interest. In addition, it is no longer necessary to pass in a ``Geometry`` object along with a ``Solver`` object to the ``plot_spatial_flux(...)`` routine (or ``plot_fission_rates(...)``)  since it may be directly accessed from the ``Solver``. Now the user simply needs to pass in a ``Solver`` object to these routines. All of the sample input files have been updated to reflect these changes. Once the review process for this PR has been finalized, I will update the online documentation.


openmoc/plotter.py
```diff
@@ -625,24 +624,22 @@ def plot_cmfd_cells(geometry, cmfd, gridsize=250, xlim=None, ylim=None):
 
 ##
 # @brief This method takes in a Geometry object and plots a color-coded 2D
```
samuelshaner: Change to `This method takes in a Solver object...`

openmoc/plotter.py
```diff
@@ -747,31 +720,174 @@ def plot_fluxes(geometry, solver, energy_groups=[1],
...
 
 
 ##
-# @brief This method takes in a Geometry object and plots a color-coded 2D
-#        surface plot representing the flat source region fission rates in 
-#        the Geometry.
-# @details The geometry object must be initialized with Materials, Cells,
-#          Universes and Lattices before being passed into this method. A user
-#          may invoke this function from an OpenMOC Python file as follows:
+# @brief This method takes in a Geometry object and plots the scalar
```
samuelshaner: ditto on `Geometry` to `Solver`

wbinventor: done

samuelshaner:
Tested it out on Mac and Linux and Python 2 and 3 and everything works great.

What do you think about adding an example in one of the input files of plotting the energy flux? Perhaps plot the fuel and moderator flux for the pin cell?

openmoc/plotter.py
wbinventor: done

wbinventor:
Sure, that sounds interesting. Since this requires one to pass in the FSR IDs, do you think the example should select (x,y) points residing in the fuel/moderator, obtain the FSR IDs from the ``Geometry``, and then plot their energy fluxes?

samuelshaner:
That sounds like a good approach since the FSRs are set somewhat arbitrarily.

wbinventor:
In the ``sample-input/pin-cell/pin-cell.py`` problem we actually only have the fuel and moderator (gap and clad are homogenized into the fuel), so this was a very easy addition to the input file:

```
plotter.plot_energy_fluxes(solver, fsrs=range(geometry.getNumFSRs()))
```

which results in the following plots:

flux-fsr-0

flux-fsr-1

samuelshaner:
Nice - I like that example.

wbinventor:
The online documentation has been updated to reflect the change to the ``plot_fluxes(...)`` routine to ``plot_spatial_fluxes(...)``, as well as the introduction of the new ``plot_energy_fluxes(...)`` routine.

PR #133, opened 2015-04-07
Refactorization of polar angle quadrature 

wbinventor:
This PR refactors the polar angle quadrature implementation in OpenMOC. Previously, the polar angle quadrature was pre-defined to strictly those given by the "Leonard Optimal Quadrature" or the "Tabuchi-Yamamoto Quadrature," each of which are commonly used in other lattice physics codes. With this PR, any polar angle quadrature may be used.

In particular, the general ``PolarQuad`` class is defined in place of the old ``Quadrature`` class. The ``PolarQuad`` may be used with any user-defined polar angle quadrature. In addition, a number of ``PolarQuad`` subclasses are defined based on polar angle quadratures given in "Lattice Physics Computation", Handbook of Nucl. Eng, Knott & Yamamoto, 2010. These subclasses are the following:

* ``TYPolarQuad`` (Tabuch-Yamamoto Quadrature)
* ``LeonardPolarQuad`` (Leonard's Optimal Quadrature)
*  ``GLPolarQuad`` (Gauss-Legendre Quadrature)
* ``EqualAnglePolarQuad``
* ``EqualWeightPolarQuad``

Each of the ``PolarQuad`` subclasses may be assigned some number of polar angles from which it will generate the appropriate quadrature. The user must then assign the ``PolarQuad`` object to the ``Solver``. The ``Solver`` initializes a ``TYPolarQuad`` object with 3 polar angles by default if a ``PolarQuad`` is not assigned to it by the user. An example is given as follows:

```python
# Initialize a Gauss-Legendre polar quadrature with 5 (10) polar angles
polar_quad = opnemoc.GLPolarQuad()
polar_quad.setNumPolarAngles(5)

# Assign the polar quadrature object to the Solver
solver.setPolarQuadrature(polar_quad)

...

solver.convergeSource()
```

Once this PR has been reviewed I will update the online documentation.


src/PolarQuad.cpp
```diff
@@ -0,0 +1,631 @@
...
+    sin_thetas[1] = 0.537707;
+    sin_thetas[2] = 0.932954;
+    weights[0] = 0.046233;
+    weights[1] = 0.283619;
+    weights[2] = 0.670148;
+  }
+
+  /* Set the arrays of sin thetas and weights */
+  PolarQuad::setSinThetas(sin_thetas, _num_polar);
+  PolarQuad::setWeights(weights, _num_polar);
```
samuelshaner: Is there a reason you are using temporary variables and calling these functions instead of directly setting `_sin_thetas` and `_weights` here?

src/PolarQuad.cpp
```diff
@@ -0,0 +1,631 @@
...
+  if (_weights != NULL)
+    delete [] _weights;
+
+  /* Initialize memory for arrays */
+  _weights = new FP_PRECISION[_num_polar];
+
+  /* Extract weights from user input */
+  for (int i=0; i < _num_polar; i++)
+    _weights[i] = FP_PRECISION(weights[i]);
+}
```
samuelshaner: Might consider adding a check to see if the weights sum to 1.0. Could be useful in case users think the weights have to sum to pi.

src/PolarQuad.cpp
wbinventor: Well my intent was to only have allocation/deallocation of the ``_sin_thetas`` and ``_weights`` array in a single place - namely, the ``PolarQuad`` parent class' ``initialize()`` routine. I understand this may seem clunky here since it requires the use of temporary arrays in each subclass' ``initialize()`` routine.

src/PolarQuad.cpp
wbinventor: You mean some sort of routine the users call? Or one that the solver calls during initialization? Perhaps a ``PolarQuad::checkWeightsSum()`` which is called in the ``PolarQuad::initialize()`` routine, which is called by the ``Solver::initializePolarQuadrature()``?

samuelshaner: I was just thinking of a `weights_sum` variable that is initialized before the loop in lines 205-207 and then accumulated in the loop. After the loop, you could just check to see if `weights_sum` is equal to 1 +/- some tolerance.

wbinventor: And it would throw a runtime error if the weights don't sum to one? Or perhaps just a warning?

samuelshaner: I would recommend an error since the quadrature won't be correct if the weights don't sum to one.

wbinventor: Ok, there is now a check that the weights sum to 1.0 in the ``PolarQuad::setWeights(...)`` routine. If the relative difference between the sum of weights and 1.0 is greater than the newly defined ``POLAR_WEIGHT_SUM_TOL`` (in "src/PolarQuad.h") then an error is thrown.

wbinventor:
The refactorization now extends to the ``GPUSolver`` class, which I forgot to include earlier.

wbinventor:
This PR has been updated with an error checking conditional in the ``PolarQuad::setSinThetas(...)`` routine to ensure that all values are in the range [0,1]. In addition, a few bugs were fixed which prevented NumPy arrays of values to be passed in by the user to the ``PolarQuad::setSinThetas(...)`` and ``PolarQuad::setWeights(...)`` routine. All eigenvalues for the sample input problems align with those expected for all of the new polar quadrature subclasses.

openmoc/plotter.py
```diff
@@ -747,31 +720,174 @@ def plot_fluxes(geometry, solver, energy_groups=[1],
     fig = plt.figure()
     plt.imshow(np.flipud(fluxes[index,:,:]), extent=coords['bounds'])
     plt.colorbar()
-    plt.title('Flat Source Region Scalar Flux in Group ' + str(group))
+    plt.title('FSR Scalar Flux (Group {0}'.format(group))
```
samuelshaner: Looks like you are missing a `)` here.

wbinventor: Wow, good catch! I just pushed the fix.

openmoc/plotter.py
```diff
@@ -747,31 +720,174 @@ def plot_fluxes(geometry, solver, energy_groups=[1],
...
+                  'total number of FSRs %d', fsr, num_fsrs)
+
+  else:
+    py_printf('ERROR', 'Unable to plot the flux vs. energy since ' + \
+              'the fsrs is not a Python tuple, list or NumPy array')
+
+  if isinstance(group_bounds, (tuple, list, np.ndarray)):
+
+    if not all(low < up for low, up in zip(group_bounds, group_bounds[1:])):
+      py_printf('ERROR', 'Unable to plot hte flux vs. energy since the ' + \
```
samuelshaner: typo

wbinventor: Fixed

samuelshaner:
The `GPUSolver` currently has a maximum number of polar angles set to 3, but some of the quadrature sets allow for more than 3 angles. It would be nice to have an error message or variable number of polar angles allows for the GPUSolver to avoid any confusion.

wbinventor:
I added an error message in the event that a user has assigned a ``PolarQuad`` object to the ``GPUSolver`` with more angles than were defined in the ``MAX_POLAR_ANGLES`` macro in "GPUSolver.h". At the moment, this maximum must be defined at compile time in order to allocate constant memory on the GPU for the polar quadrature. I updated the ``MAX_POLAR_ANGLES`` macro to 10 polar angles from what was previously 3. Hopefully this is a reasonable workaround until we come up with a better scheme for a more configurable polar quadrature to be defined at runtime.

PR #134, opened 2015-04-08
FSR volume getter / corrector routines

wbinventor:
This PR introduces new routines to the ``Solver`` to get and set flat source region (FSR) volumes. In particular, the ``Solver::getFSRVolume(fsr_id)`` and ``Solver::correctFSRVolume(fsr_id, fsr_volume)`` routines are introduced for the first time. Furthermore, this PR abstracts the FSR volume calculation to the ``Solver`` parent class away from its subclasses to improve code reuse.

The goal of these new routines is to allow users to override the volume calculation performed by OpenMOC during the ``Solver's`` initialization phase. The ``Solver`` computes an approximate volume for each FSR based on the length of each track segment crossing the FSR and the angular quadrature. If a user is concerned about using these approximate volumes, they may now correct the volume. This process which will cause OpenMOC to correct the length of each track segment crossing the FSR such that they integrate to the correct (user-defined) volume. From some preliminary tests with basic pin cell problems, this is only a few pcm (+/-2-3 pcm) issue if 64-128 angles are used with a relatively fine track spacing (0.01-0.005 cm).

This PR builds up on PR #133 for the polar angle quadrature refactorization. This PR should not be reviewed until #133 is finalized since it will significantly reduce the amount of code included in the files changed log.


wbinventor:
I've updated this PR to abstract the FSR volume calculation into the ``TrackGenerator`` and out of the ``Solver``. In addition, the volume correction is now applied separately to ``Track`` segments for each azimuthal angle.

src/Solver.cpp
```diff
@@ -371,6 +390,51 @@ void Solver::useExponentialIntrinsic() {
...
+    delete [] _FSR_materials;
+
+  /* Get an array of volumes indexed by FSR  */
+  _FSR_volumes = _track_generator->getFSRVolumes();
+
+  /* Allocate an array of Material pointers indexed by FSR */
+  _FSR_materials = new Material*[_num_FSRs];
+
+  /* Compute the number of fissionable Materials */
+  std::map<int, Material*> all_materials = _geometry->getAllMaterials();
```
samuelshaner: `all_materials` doesn't seem to be used anywhere so I would recommend taking this line out.

src/Solver.cpp
```diff
@@ -371,6 +390,51 @@ void Solver::useExponentialIntrinsic() {
...
+  _num_fissionable_FSRs = 0;
+
+  /* Loop over all FSRs to extract FSR material pointers */
+  for (int r=0; r < _num_FSRs; r++) {
+
+    /* Assign the Material corresponding to this FSR */
+    _FSR_materials[r] =  _geometry->findFSRMaterial(r);
```
samuelshaner: Very minor point - there are two spaces after the equals sign.

src/TrackGenerator.cpp
```diff
@@ -1424,38 +1544,68 @@ bool TrackGenerator::readTracksFromFile() {
...
+
+  /* Correct volume separately for each azimuthal angle */
+  for (int i=0; i < _num_azim; i++) {
+
+    /* Initialize volume to zero for this azimuthal angle */
+    volume = 0;
+
+    /* Compute effective track spacing for this azimuthal angle */
+    dx_eff = (_geometry->getWidth() / _num_x[i]);
+    dy_eff = (_geometry->getHeight() / _num_y[i]);
```
samuelshaner: `dy_eff` is not needed.

src/Solver.cpp
wbinventor: Got it

src/Solver.cpp
wbinventor: Yeah, that was legacy from the prior implementation, it's removed now

wbinventor:
@samuelshaner - I addressed your comments about unused legacy variables in ``Solver::initilaizeFSRs()`` and ``TrackGenerator::correctFSRVolume(...)``.

src/TrackGenerator.cpp
```diff
@@ -189,6 +162,208 @@ FP_PRECISION* TrackGenerator::getAzimWeights() {
 
 
 /**
+ * @brief Get the maximum allowable optical lenght for a track segment
```
samuelshaner: typo: lenght -> length

samuelshaner:
@wbinventor everything looks good! As soon as the typo is fixed I'll merge it in.

wbinventor:
@samuelshaner I fixed the typo

Issue #135, opened 2015-04-09
Bug when OpenMOC Tries to Handle Non-Square Lattices


geogunow:
When creating a ``Lattice`` with a differing number of cells in the x and y directions, OpenMOC encounters a segmentation fault. 

For instance, if I wish to fill a lattice with the same universe in a rectangular configuration I might initialize a lattice as:
```python
lattice = Lattice(name="lattice")
lattice.setWidth(pitch, pitch)
template = [[universeA]*numcells_x]*numcells_y
lattice.setUniverses(template)
```

If numcells_x (number of cells along the x direction) equals numcells_y (number of cells along the y direction), then the code executes as expected. However, if these numbers differ, then the segmentation fault is encountered.

Sample code is provided below that demonstrates the behavior:

```python
import numpy
from openmoc import *
import openmoc.log as log
import openmoc.plotter as plotter
from openmoc.options import Options


###############################################################################
#######################   Main Simulation Parameters   ########################
###############################################################################

options = Options()

num_threads = options.getNumThreads()
track_spacing = 0.1
num_azim = 32
tolerance = options.getTolerance()
max_iters = options.getMaxIterations()

log.set_log_level('NORMAL')

log.py_printf('TITLE', 'Simulating a two group homogeneous infinite medium...')
log.py_printf('HEADER', 'The reference keff = 1.72...')


###############################################################################
###########################   Creating Materials   ############################
###############################################################################

log.py_printf('NORMAL', 'Creating materials...')

sigT = [1]
sigA = [0.5]
nuSigF = [0.5]
chi = [1]
sigS = [[0.5]]

infinite_medium = Material(name="mat")
infinite_medium.setNumEnergyGroups(1)
infinite_medium.setSigmaA(numpy.array([0.5]))
infinite_medium.setSigmaF(numpy.array([0.5]))
infinite_medium.setNuSigmaF(numpy.array([0.5]))
infinite_medium.setSigmaS(numpy.array([0.5]))
infinite_medium.setChi(numpy.array([1.0]))
infinite_medium.setSigmaT(numpy.array([1.0]))

###############################################################################
###########################   Creating Surfaces   #############################
###############################################################################

log.py_printf('NORMAL', 'Creating surfaces...')

pitch = 30.0
left = XPlane(x=-pitch/2)
right = XPlane(x=pitch/2)
top = YPlane(y=pitch/2)
bottom = YPlane(y=-pitch/2)

left.setBoundaryType(REFLECTIVE)
[...]

PR #136, opened 2015-04-09
Bug fix for irregularly shaped lattices

wbinventor:
This PR is a bug fix for issue #135. The new implementation allows for the creation of ``Lattices`` with different numbers of ``Universes`` along and x, y dimensions. In addition, the x,y width (or pitch) may now be different for the x and y domains.

### DIFF
--- a/src/Universe.cpp
+++ b/src/Universe.cpp
@@ -1079,11 +1079,11 @@ void Lattice::setWidth(double width_x, double width_y) {
  *                                   [u2, u3, u2, u3]])
  * @endcode
  *
- * @param num_x the number of Lattice cells along x
  * @param num_y the number of Lattice cells along y
+ * @param num_x the number of Lattice cells along x
  * @param universes the array of Universes for each Lattice cell
  */
-void Lattice::setUniverses(int num_x, int num_y, Universe** universes) {
+void Lattice::setUniverses(int num_y, int num_x, Universe** universes) {
 
   /* Clear any Universes in the Lattice (from a previous run) */
   for (int i=0; i < _num_x; i++)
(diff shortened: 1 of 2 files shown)


src/Universe.cpp
```diff
@@ -1079,11 +1079,11 @@ void Lattice::setWidth(double width_x, double width_y) {
...
- * @param num_x the number of Lattice cells along x
  * @param num_y the number of Lattice cells along y
+ * @param num_x the number of Lattice cells along x
  * @param universes the array of Universes for each Lattice cell
  */
-void Lattice::setUniverses(int num_x, int num_y, Universe** universes) {
+void Lattice::setUniverses(int num_y, int num_x, Universe** universes) {
 
   /* Clear any Universes in the Lattice (from a previous run) */
   for (int i=0; i < _num_x; i++)
```
samuelshaner: I think this should be `_num_y` to be consistent with lines 804-805.

src/Universe.cpp
```diff
@@ -1104,7 +1104,7 @@ void Lattice::setUniverses(int num_x, int num_y, Universe** universes) {
 
     _universes.push_back(std::vector< std::pair<int, Universe*> >());
 
-    for (int j = 0; j< _num_x; j++){
+    for (int j = 0; j < _num_x; j++){
```
samuelshaner: I feel like having `i` correspond to the x index and `j` correspond to the y index would be more conventional.

src/Universe.cpp
wbinventor: Very good catch.

src/Universe.cpp
wbinventor: I agree and pushed the revised loop counters.

PR #137, opened 2015-04-12
Abstraction of exponential calculations

wbinventor:
This PR abstracts the exponential calculation into a new ``ExpEvaluator`` class (and ``GPUExpEvaluator`` for GPUs). This PR is the first phase to modularize the ``Solver`` and its subclasses and extend them as flexible containers for both fixed source and eigenvalue calculations. This PR builds upon some of my previous ones, and should be reviewed after PR #134.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -374,7 +374,7 @@ def customize_linker(self):
       self.set_executable('linker_so', 'bgxlc++_r')
       self.set_executable('linker_exe', 'bgxlc++_r')
 
-    # If the filename for the extension contains cuda, use g++ to link
+    # If the filename for the extension contains cuda, use nvcc to link
     if 'cuda' in output_filename:
       self.set_executable('linker_so', 'nvcc')
       self.set_executable('linker_exe', 'nvcc')
--- a/setup.py
+++ b/setup.py
@@ -376,8 +376,8 @@ def customize_linker(self):
 
     # If the filename for the extension contains cuda, use g++ to link
     if 'cuda' in output_filename:
-      self.set_executable('linker_so', 'g++')
-      self.set_executable('linker_exe', 'g++')
+      self.set_executable('linker_so', 'nvcc')
+      self.set_executable('linker_exe', 'nvcc')
 
     # Now call distutils-defined link method
     super_link(target_desc, objects,
--- a/src/ExpEvaluator.h
+++ b/src/ExpEvaluator.h
@@ -20,7 +20,9 @@
  * @class ExpEvaluator ExpEvaluator.h "src/ExpEvaluator.h"
  * @brief This is a class for evaluating exponentials.
  * @details The ExpEvaluator includes different algorithms to evaluate
- *          exponentials with varying degrees of accuracy and speed.
+ *          exponentials with varying degrees of accuracy and speed. This
+ *          is a helper class for the Solver and its subclasses and it not
+ *          intended to be initialized as a standalone object.
  */
 class ExpEvaluator {
 
@@ -83,4 +85,4 @@ inline int round_to_int(double x) {
 }
 
 
-#endif /* EXPEVALUATOR_H_ */
+ #endif /* EXPEVALUATOR_H_ */
--- a/src/ExpEvaluator.h
+++ b/src/ExpEvaluator.h
@@ -45,6 +45,12 @@ private:
 
   /** Twice the number of polar angles */
   int _two_times_num_polar;
+
+  /** The maximum optical length a track is allowed to have */
+  FP_PRECISION _max_optical_length;
+
+  /** The maximum acceptable approximation error for exponentials */
+  FP_PRECISION _exp_precision;
   
 public:
 
@@ -52,15 +58,19 @@ public:
   virtual ~ExpEvaluator();
 
   void setPolarQuadrature(PolarQuad* polar_quad);
+  void setMaxOpticalLength(FP_PRECISION max_optical_length);
+  void setExpPrecision(FP_PRECISION exp_precision);
   void useInterpolation();
   void useIntrinsic();
 
+  FP_PRECISION getMaxOpticalLength();
+  FP_PRECISION getExpPrecision();
   bool isUsingInterpolation();
   FP_PRECISION getTableSpacing();
   int getTableSize();
   FP_PRECISION* getExpTable();
 
-  void initialize(double max_tau, double tolerance);
+  void initialize(double tolerance);
   FP_PRECISION computeExponential(FP_PRECISION tau, int polar);
 };
 
--- a/src/ExpEvaluator.h
+++ b/src/ExpEvaluator.h
@@ -1,0 +1,98 @@
+/**
+ * @file ExpEvaluator.h
+ * @brief The ExpEvaluator class.
+ * @date April 9, 2015.
+ * @author William Boyd, MIT, Course 22 (wboyd@mit.edu)
+ */
+
+#ifndef EXPEVALUATOR_H_
+#define EXPEVALUATOR_H_
+
+#ifdef __cplusplus
+#define _USE_MATH_DEFINES
+#include <math.h>
+#include "log.h"
+#include "PolarQuad.h"
+#endif
+
+//FIXME
+#ifdef NVCC
+#define CUDA_CALLABLE __host__ __device__
+#else
+#define CUDA_CALLABLE
+#endif
+
+
+/**
+ * @class ExpEvaluator ExpEvaluator.h "src/ExpEvaluator.h"
+ * @brief This is a class for evaluating exponentials.
+ * @details The ExpEvaluator includes different algorithms to evaluate
+ *          exponentials with varying degrees of accuracy and speed.
+ */
+class ExpEvaluator {
+
+private:
+
+  /** A boolean indicating whether or not to use linear interpolation */
+  bool _interpolate_exponential;
+
+  /** The exponential linear interpolation table */
+  FP_PRECISION* _exp_table;
+
+  /** The spacing for the exponential linear interpolation table */
+  FP_PRECISION _exp_table_spacing;
+
+  /** The inverse spacing for the exponential linear interpolation table */
+  FP_PRECISION _inverse_exp_table_spacing;
+
+  /** The PolarQuad object of interest */
+  PolarQuad* _polar_quad;
+
+  /** Twice the number of polar angles */
+  int _two_times_num_polar;
+  
+public:
+
+  ExpEvaluator();
+  virtual ~ExpEvaluator();
+
+  void setPolarQuadrature(PolarQuad* polar_quad);
+  void useExponentialInterpolation();
+  void useExponentialIntrinsic();
+
+  bool isUsingExponentialInterpolation();
+
+  void initialize(double max_tau, double tolerance);
+  FP_PRECISION computeExponential(FP_PRECISION tau, int polar);
+};
+
+
+/**
+ * @brief Rounds a single precision floating point value to an integer.
+ * @param x a float precision floating point value
+ * @brief the rounded integer value
+ */
+CUDA_CALLABLE inline int round_to_int(float x) {
+#ifdef NVCC
+  return __float2int_rd(x);
+#else
+  return lrintf(x);
+#endif
+}
+
+
+/**
+ * @brief Rounds a double precision floating point value to an integer.
+ * @param x a double precision floating point value
+ * @brief the rounded integer value
+ */
+CUDA_CALLABLE inline int round_to_int(double x) {
+#ifdef NVCC
+  return __double2int_rd(x);
+#else
+  return lrint(x);
+#endif
+}
+
+
+#endif /* EXPEVALUATOR_H_ */
--- a/src/ExpEvaluator.h
+++ b/src/ExpEvaluator.h
@@ -70,7 +70,7 @@ public:
   int getTableSize();
   FP_PRECISION* getExpTable();
 
-  void initialize(double tolerance);
+  void initialize();
   FP_PRECISION computeExponential(FP_PRECISION tau, int polar);
 };
 
--- a/src/ExpEvaluator.h
+++ b/src/ExpEvaluator.h
@@ -15,13 +15,6 @@
 #include "PolarQuad.h"
 #endif
 
-//FIXME
-#ifdef NVCC
-#define CUDA_CALLABLE __host__ __device__
-#else
-#define CUDA_CALLABLE
-#endif
-
 
 /**
  * @class ExpEvaluator ExpEvaluator.h "src/ExpEvaluator.h"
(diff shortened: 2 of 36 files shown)


src/ExpEvaluator.h
```diff
@@ -0,0 +1,88 @@
...
+  void setPolarQuadrature(PolarQuad* polar_quad);
+  void useInterpolation();
+  void useIntrinsic();
+
+  bool isUsingInterpolation();
+  FP_PRECISION getTableSpacing();
+  int getTableSize();
+  FP_PRECISION* getExpTable();
+
+  void initialize(double max_tau, double tolerance);
```
wbinventor: One key question here is whether we truly need to use this tolerance or something else in the construction of the table.

samuelshaner: Can you expand a bit on what you mean here? Are you suggesting we investigate a more accurate approximation of adjusting the table size based on the problem specification?

wbinventor: Yes, I don't think there is a tight relationship between the tolerance on the source convergence and the individual exponential calculations. As a result, I think that this current implementation (greatly) exaggerates the size of the table that is needed, reducing cache performance.

setup.py
```diff
@@ -376,8 +376,8 @@ def link(target_desc, objects, output_filename,
 
     # If the filename for the extension contains cuda, use g++ to link
```
samuelshaner: It looks like you changed it to used nvcc to link.

src/Solver.cpp
```diff
@@ -57,6 +54,7 @@ Solver::Solver(Geometry* geometry, TrackGenerator* track_generator) {
   _polar_quad = new TYPolarQuad();
   _num_polar = 3;
   _two_times_num_polar = 2 * _num_polar;
```
samuelshaner: I think this can be taken out now.

src/Solver.cpp
```diff
@@ -385,7 +352,55 @@ void Solver::useExponentialInterpolation() {
...
+ * @details Deletes memory old PolarQuad if one was previously allocated.
+ */
+void Solver::initializePolarQuadrature() {
+
+  FP_PRECISION azim_weight;
+
+  /* Create Tabuchi-Yamamoto polar quadrature if a
+   * PolarQuad was not assigned by the user */
+  if (_polar_quad == NULL)
+    _polar_quad = new TYPolarQuad();
```
samuelshaner: I don't think this will ever be called since `_polar_quad` is set to `TYPolarQuad` during initialization of the `Solver`.

src/Solver.h
```diff
@@ -11,14 +11,16 @@
...
 #include "Timer.h"
 #include "PolarQuad.h"
 #include "TrackGenerator.h"
 #include "Cmfd.h"
+#include "ExpEvaluator.h"
 #include <math.h>
 #endif
 
-/** Indexing macro for the scalar flux in each FSR and energy group */
+/** Indexing macro for the scalar flux in each FSR and energy grou */
```
samuelshaner: typo

src/accel/cuda/GPUExpEvaluator.h
```diff
@@ -0,0 +1,50 @@
...
+
+#ifdef __cplusplus
+#define _USE_MATH_DEFINES
+#include <math.h>
+#include "../../ExpEvaluator.h"
+#endif
+
+
+/** The maximum number of polar angles to reserve constant memory on GPU */
+#define MAX_POLAR_ANGLES 10
```
samuelshaner: Should this go in constants?

src/accel/cuda/GPUSolver.h
```diff
@@ -33,12 +36,6 @@
...
 #define boundary_flux(t,pe2) (boundary_flux[2*(t)*(*polar_times_groups)+(pe2)])
 
-/** The value of 4pi: \f$ 4\pi \f$ */
-#define FOUR_PI 12.5663706143
-
-/** The values of 1 divided by 4pi: \f$ \frac{1}{4\pi} \f$ */
-#define ONE_OVER_FOUR_PI 0.0795774715
-
 /** The maximum number of polar angles to reserve constant memory on GPU */
 #define MAX_POLAR_ANGLES 10
```
samuelshaner: ditto. What do you think about adding `MAX_POLAR_ANGLES` and `MAX_AZIM_ANGLES` to `constants.h` (and perhaps adding `_GPU` at the end)?

wbinventor:
I just pushed changes to address all of your comments thus far @samuelshaner

samuelshaner:
Have you compiled and tested this on a GPU? When I try to run `c5g7-gpu.py` I get the following error:

  File "c5g7-gpu.py", line 2, in <module>
    from openmoc.cuda import GPUSolver
  File "/home/shaner/.local/lib/python2.7/site-packages/openmoc/cuda/__init__.py", line 6, in <module>
    import _openmoc_cuda
ImportError: /home/shaner/.local/lib/python2.7/site-packages/_openmoc_cuda.so: undefined symbol: __cudaRegisterLinkedBinary_51_tmpxft_000020fd_00000000_9_clone_compute_30_cpp1_ii_d6535865

Any thoughts?

src/accel/cuda/GPUSolver.cu
```diff
@@ -1042,17 +949,18 @@ void GPUSolver::initializePolarQuadrature() {
...
-               "and recompile.", _num_polar, MAX_POLAR_ANGLES);
+               "angles. Update the MAX_POLAR_ANGLES_GPU macro in constants.h "
+               "and recompile.", _num_polar, MAX_POLAR_ANGLES_GPU);
 
   /* Copy the number of polar angles to constant memory on the GPU */
   cudaMemcpyToSymbol(num_polar, (void*)&_num_polar, sizeof(int), 0,
                      cudaMemcpyHostToDevice);
 
   /* Copy twice the number of polar angles to constant memory on the GPU */
+  _two_times_num_polar = 2 * _num_polar * _num_groups;
```
samuelshaner: `* _num_groups` should not be here.

wbinventor:
For the record, as was noticed by @samuelshaner, this PR requires NVIDIA CUDA versions 6.0 or higher - versions 5.5 and lower are no longer supported. The reason is that more recent versions  of the ``nvcc`` compiler allow "device code linking" which is now done between the ``GPUSolver`` and the new``GPUExpEvaluator`` class.

samuelshaner:
I just installed CUDA 7.0 on an AWS AMI and got the following error when compiling:

ccache gcc -DFP_PRECISION=double -DDOUBLE -DCUDA -DCCACHE_CC=nvcc -I/usr/local/cuda/include -I/usr/lib/python2.7/dist-packages/numpy/core/include -I/usr/include/python2.7 -c openmoc/cuda/openmoc_cuda_wrap.cpp -o build/temp.linux-x86_64-2.7/openmoc/cuda/openmoc_cuda_wrap.o -c -O3 -ffast-math -fopenmp -std=c++0x -fpic
In file included from openmoc/cuda/openmoc_cuda_wrap.cpp:3057:0:
openmoc/cuda/../../src/accel/cuda/GPUSolver.h:19:36: fatal error: sm_13_double_functions.h: No such file or directory
 #include <sm_13_double_functions.h>
                                    ^
compilation terminated.
error: command 'ccache' failed with exit status 1

According to the CUDA Toolkit v7.0 release notes (http://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#axzz3ZYL7uEOq) "The CUDA Toolkit and CUDA Drive no longer supports the sm_10, sm_11, sm_12, and sm_13 architectures.". Hence, `sm_13_double_functions.h` is no longer in `/usr/local/cuda/include`.

Do you know if there is a way to get around using `sm_13_double_functions.h`? I briefly tried to figure it out, but couldn't find a solution.

wbinventor:
I don't think this will be an issue - we should just note in our v0.1.5
release that we only support NVIDIA 2.0 and later devices. The
"sm_20_double_functions.h" (which are already included) should provide the
same routines we need as sm_13 (e.g., atomic add).

wbinventor:
I can fix this issue when I return or if you want it in now you can just
submit a PR which removes that include.

samuelshaner:
When I take out `#include <sm_13_double_functions.h>`, it works just fine with CUDA 6.5 on mightywboyd. However, when I take that line out and try to compile on AWS with CUDA 7.0, I get the error shown below. Looks like more of a python related error than a CUDA related error. I'll play around with it a bit and try to get something working.

ccache gcc -DFP_PRECISION=double -DDOUBLE -DCUDA -DCCACHE_CC=nvcc -I/usr/local/cuda/include -I/usr/lib/python2.7/dist-packages/numpy/core/include -I/usr/include/python2.7 -c openmoc/cuda/openmoc_cuda_wrap.cpp -o build/temp.linux-x86_64-2.7/openmoc/cuda/openmoc_cuda_wrap.o -c -O3 -ffast-math -fopenmp -std=c++0x -fpic
In file included from /usr/local/cuda/include/thrust/system/cuda/detail/bulk/detail/throw_on_error.hpp:20:0,
                 from /usr/local/cuda/include/thrust/system/cuda/detail/bulk/future.hpp:21,
                 from /usr/local/cuda/include/thrust/system/cuda/detail/bulk/execution_policy.hpp:19,
                 from /usr/local/cuda/include/thrust/system/cuda/detail/bulk/bulk.hpp:20,
                 from /usr/local/cuda/include/thrust/system/cuda/detail/bulk.h:47,
                 from /usr/local/cuda/include/thrust/system/cuda/detail/for_each.inl:25,
                 from /usr/local/cuda/include/thrust/system/cuda/detail/for_each.h:64,
                 from /usr/local/cuda/include/thrust/system/detail/adl/for_each.h:32,
                 from /usr/local/cuda/include/thrust/detail/for_each.inl:27,
                 from /usr/local/cuda/include/thrust/for_each.h:279,
                 from /usr/local/cuda/include/thrust/system/detail/generic/transform.inl:19,
                 from /usr/local/cuda/include/thrust/system/detail/generic/transform.h:105,
                 from /usr/local/cuda/include/thrust/detail/transform.inl:25,
                 from /usr/local/cuda/include/thrust/transform.h:724,
                 from /usr/local/cuda/include/thrust/system/detail/generic/reduce_by_key.inl:29,
                 from /usr/local/cuda/include/thrust/system/detail/generic/reduce_by_key.h:88,
                 from /usr/local/cuda/include/thrust/detail/reduce.inl:26,
                 from /usr/local/cuda/include/thrust/reduce.h:784,
                 from openmoc/cuda/../../src/accel/cuda/GPUSolver.h:17,
                 from openmoc/cuda/openmoc_cuda_wrap.cpp:3057:
/usr/local/cuda/include/thrust/system/cuda/detail/bulk/detail/terminate.hpp: In function ‘void thrust::system::cuda::detail::bulk_::detail::terminate_with_message(const char*)’:
/usr/local/cuda/include/thrust/system/cuda/detail/bulk/detail/terminate.hpp:45:3: error: ‘PySys_WriteStdout’ is not a member of ‘std’
   std::printf("%s\n", message);
   ^
/usr/local/cuda/include/thrust/system/cuda/detail/bulk/detail/terminate.hpp:45:3: note: suggested alternative:
In file included from /usr/include/python2.7/Python.h:129:0,
                 from openmoc/cuda/openmoc_cuda_wrap.cpp:154:
/usr/include/python2.7/sysmodule.h:17:6: note:   ‘PySys_WriteStdout’
 PyAPI_FUNC(void) PySys_WriteStdout(const char *format, ...)
      ^
error: command 'ccache' failed with exit status 1

wbinventor:
@samuelshaner thanks for digging into this and discovering the compilation issue with CUDA 7.0. This was related to the [Thrust](https://code.google.com/p/thrust/) library's new use of ``std::printf(...)`` rather than simply ``printf(...)``, which SWIG attempts to wrap using the ``PySys_WriteStdout(...)`` routine to channel both Python and C/C++ output to the same I/O stream. I pushed a fix that overrides SWIG's wrapping of ``std::printf`` statements in the CUDA source code. 

Also, I removed the ``<sm_13_atomic_functions.h>`` include from ``GPUSolver`` since these routines are no longer shipped NVIDIA CUDA >= 7.0. This breaks backwards compatibility with older versions of CUDA, but I do not see an issue with this since NVIDIA GPUs with compute capability <2.0 are well on their way out of style anway.

In addition, I added a ``_max_optical_length`` and ``_exp_precision`` class attributes to ``ExpEvaluator``, along with corresponding getter/setter routines. By default, these are assigned to the new ``MAX_OPTICAL_LENGTH`` (*e.g.*, 10.0) and ``EXP_PRECISION`` (*e.g.*, 1E-5) macros defined in "constants.h". The choices for these defaults are based on Yamamotos' analysis in his 2003 paper on the use of interpolation tables to compute exponentials. These changes decouple the size of the exponential interpolation table from the source convergence threshold, which leads to a >30x reduction in the table size overall (it is now 20-30 KB in single precision mode).

wbinventor:
I spoke too soon last night since I forgot to add in handles to adjust the precision (max. approximation error) and max optical path length of the exponential interpolation table. My most recent commits added the ``Solver.setMaxOpticalLength(...)`` and ``Solver.setExpPrecision(...)`` routines. Note that some of these changes are intrinsically linked to PR #141 and hence will not be fully complete until that PR is reviewed.

samuelshaner:
@wbinventor thanks for looking into the issue and glad to hear it is resolved. I'm going to test everything out again and then I think it will be time to merge.

samuelshaner:
@wbinventor can you add `constants.h` to the `.i` files? With the constants moved to `constants.h` and this file not included in the `.i` files, swig does not seem to wrap the constants. The plotter therefore throws an error when it sees `TINY_MOVE`. Adding includes for `constants.h` in the `.i` files seemed to fix it.

wbinventor:
It compiled and ran for you when you added "constants.h" to the interface files? Previously I had it setup in that way, and it was compiling just fine for me, but breaking when I tried to run the code last night. I'll give it another try now to see if I missed something before.

samuelshaner:
Yes, I added includes at the bottom and top of `openmoc.i` and it compiled just fine. I admittedly only tried it in `openmoc.i` though (no other `.i` files).

wbinventor:
I added "constants.h" back to the SWIG interface files and plotting is now working again. In order to make this work, I had to put ``MAX_AZIM_ANGLES_GPU`` and ``MAX_POLAR_ANGLES_GPU`` inside a ``#ifdef CUDA`` block in "constants.h". This is needed since SWIG wraps macro definitions regardless of whether or not they are used in a code package. In our case, the GPU-specific constant macros are only used in the ``openmoc.cuda`` package, so the ``openmoc`` package was failing to import properly since SWIG was trying to access ``MAX_..._ANGLES_GPU`` macros which did not exist in the package. In any case, I think this issue is resolved.

samuelshaner:
Since `_cmfd_surface_fwd` `_cmfd_surface_bwd` are not initialized anywhere, the edges of the segment that come about from chopping are initialized to garbage and CMFD seg faults if segments are cut. I would recommend adding an else statement or finding a way to initialize the surfaces to -1.

src/TrackGenerator.cpp
```diff
@@ -1608,3 +1608,82 @@ void TrackGenerator::correctFSRVolume(int fsr_id, FP_PRECISION fsr_volume) {
...
+
+        /* Split the segment into sub-segments */
+        for (int k=0; k < min_num_cuts; k++) {
+
+          /* Create a new Track segment */
+          segment* new_segment = new segment;
+          new_segment->_material = material;
+          new_segment->_length = length / FP_PRECISION(min_num_cuts);
+          new_segment->_region_id = fsr_id;
+          new_segment->_cmfd_surface_bwd = -1;
```
wbinventor: I'm now initializing ``_cmfd_surface_fwd`` and ``_cmfd_surface_bwd`` to -1 for all split segments. b

wbinventor:

PR #139, opened 2015-04-13
Modular CMFD surface current tallying

wbinventor:
This PR presents some modifications to the CPUSolver::scalarFluxTally(...) routine. In particular, the CMFD surface current calculation is completely abstracted away from the MOC Solver class/subclasses into the Cmfd class. This streamlines the code in the Solver, building upon the changes made in PR #137.


src/CPUSolver.cpp
```diff
@@ -900,44 +698,41 @@ void CPUSolver::scalarFluxTally(segment* curr_segment,
...
-                  _exp_table[index + 2 * p +1]));
-  }
+  FP_PRECISION surf_current;
 
-  /* Evalute the exponential using the intrinsic exp(...) function */
-  else {
-    FP_PRECISION sintheta = _polar_quad->getSinTheta(p);
-    exponential = 1.0 - exp(- tau / sintheta);
+  /* Tally surface currents if CMFD is in use */
+  if (_cmfd != NULL && _cmfd->isFluxUpdateOn()){
```
samuelshaner: Most segments will likely not intersect with a CMFD surface. I think it would be more computationally efficient to check if a segment intersects with a valid cmfd surface (i.e. `curr_segment->_cmfd_surface_fwd != -1` or `curr_segment->_cmfd_surface_bwd != -1` , depending on direction) before `surf_current` is computed, instead of checking in `Cmfd.cpp`.

src/Cmfd.cpp
```diff
@@ -1788,3 +1822,58 @@ void Cmfd::updateBoundaryFlux(Track** tracks, FP_PRECISION* boundary_flux,
...
+ * @brief 
+ * @param curr_segment the current Track segment
+ * @param current the current to tally
+ * @param fwd boolean indicating direction of integration along segment
+ * @param e the MOC energy group index
+ */
+void Cmfd::tallySurfaceCurrent(segment* curr_segment, 
+                               FP_PRECISION current, bool fwd, int e) {
+
+  if (curr_segment->_cmfd_surface_fwd != -1 && fwd){
```
wbinventor: The segments still have a ``_cmfd_surface_fwd`` and ``_cmfd_surface_bwd`` attribute. I think to compromise on computational efficiency and code structure, I should move the loop over energy groups and polar angles from the ``Solver::tallySurfaceCurrent(...)`` routine to here. That way the MOC ``Solver`` class is agnostic to CMFD-specific computations, like which segments intersect mesh surfaces and which do not. What do you think?

samuelshaner:
Sounds good - I like that idea.

src/Cmfd.cpp
wbinventor: I've updated the code to loop over angles and energy groups within ``Cmfd::tallySurfaceCurrent(...)`` rather than in the ``Solver::tallySurfaceCurrent(...)`` routine. This should be much more computationally efficient than before, and hide more of the CMFD-specific computations from the MOC solver.

wbinventor:

PR #140, opened 2015-04-14
OpenMP Mutex for ERROR log messages

wbinventor:
This PR introduces a ``omp_lock_t`` to the ``openmoc.log`` module for use with ``ERROR`` messages. With the current distribution of the code, if two or more threads write an error message at the same time, the log module will attempt to throw a ``std::logic_error`` exception for each thread. This often led to the ``ERROR`` message not being written to the shell at all with the following output instead:

```bash
terminate called recursively
terminate called recursively
Aborted (core dumped)
```

With this PR, a mutual exclusion lock is introduced for throwing exceptions for ``ERROR`` messages. This only allows a single thread to throw an exception, eliminating this issue entirely.


wbinventor:

PR #141, opened 2015-04-15
Deferred segment splitting

wbinventor:
This PR abstracts segment splitting process from the ray tracing phase of a simulation to the point of construction of an exponential linear interpolation table. Currently, the ``TrackGenerator`` contains a maximum optical length (10 by default) which is used to constrain the size of the exponential linear interpolation table for cache efficiency. As a result of this maximum length, if a segment's optical length was found to exceed the maximum during ray tracing, it was split into multiple sub-segments, each with an optical length less than the maximum. In this implementation, any adjustment of the material cross-sections after ray tracing was not reflected in the sub-division of track segments.

The new implementation in this PR introduces a new ``TrackGenerator.splitSegments(...)`` routine which is called by the ``Solver`` during initialization of an ``ExpEvaluator`` object (see PR #137). In this implementation, the "true" (non-subdivided) segment lengths are tabulated during ray tracing. Once a user calls the ``Solver.convergeSource(...)`` routine, the segments will be split based on a current snapshot of the material cross-sections and user-defined maximum optical length. This new formulation allows for a single ray tracing process to be utilized for multiple simulations with varying cross-sections (*e.g.*, for varying fuel enrichments, moderator densities, etc.).

Finally, this new implementation makes the track binary file storage process more robust. In particular, the "true" segment lengths are now stored to the file rather than the sub-divided lengths. This is preferable since it allows the user to re-use utilize the same track file regardless of the runtime configuration for the ``Solver`` (*i.e.*, different maximum optical lengths, or use of the ``exp(...)`` intrinsic rather than the exponential interpolation table).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,18 +5,21 @@ OpenMOC Method of Characteristics Neutral Particle Transport Code
 Welcome to the OpenMOC repository! OpenMOC is a simulation tool for 
 solving for the flux, power distribution, and multiplication factor 
 within a nuclear reactor. The code employs the deterministic method 
-of characteristics using source iteration. The OpenMOC project aims 
-to provide a simple-to-use Python package bound to a back-end of source 
-code written in C/C++ and CUDA. It includes support for constructive 
-solid geometry and 2D ray tracing for fully heterogeneous multi-group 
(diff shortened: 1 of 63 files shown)


wbinventor:
Closed and replaced with PR #147.

PR #142, opened 2015-04-18
Refactored solvers for eigenvalue, sub-criticality and fixed source calculations

wbinventor:
This PR refactors the ``Solver`` and its subclasses to permit eigenvalue, sub-criticality and fixed source calculations. This PR builds upon the revisions introduced in PRs #136, #137, #139, and #141 and should be reviewed after they are finalized.

**External source API**
A number of new routines are introduced to the ``Solver`` to permit easy assignment of fixed external sources to particular flat source regions (FSRs). This fixed source assignment API was initially created by @PrezNattyGibbs and extended in this PR. The user may assign fixed groupwise volume-averaged sources by ``Cell``, ``Material`` or FSR ID to the ``Solver`` through the following routines exposed through the OpenMOC Python API:

* ``Solver::setFixedSourceByCell(cell, group, source)``
* ``Solver::setFixedSourceByMaterial(material, group, source)``
* ``Solver::setFixedSourceByFSR(fsr_id, group, source)``

The first two routines which set sources by ``Cell`` and ``Material`` will assign the same fixed source to each FSR filled by the corresponding ``Cell`` or ``Material``. 

**Simulation modes**
This PR makes large scale modifications to the overall structure and program flow of the ``Solver`` and its subclasses. These changes were made to facilitate modular code reuse for several types of neutron transport simulations. The current release of OpenMOC only provides support for eigenvalue calculations. This PR provides routines to compute the flux, source and eigenvalue:

* ``Solver::computeFlux(...)`` - compute flux distribution for fixed sources
* ``Solver::computeSource(...)`` - compute source distribution from fixed, scatter and fission sources
* ``Solver::computeEigenvalue(...)`` - compute eigenvalue and source distribution

The final eigenvalue simulation mode is equivalent to the functionality provided in the current release, but the program flow had been significantly modified and abstracted into general purpose routines which the other modes may leverage. Each of these simulation modes will use any fixed sources defined by the user as described above.

**Various convergence criteria**
The current release of OpenMOC uses the total volume and energy integrated fission source to determine the convergence of a simulation. This PR makes it easy for the user to select from three different types of convergence criterion to apply to any of the simulation modes described above:

* ``SCALAR_FLUX`` - average relative groupwise scalar flux change in each FSR
* ``FISSION_SOURCE`` - average relative energy-integrated fission source change in each FSR
* ``TOTAL_SOURCE`` - average relative groupwise total source change in each FSR

Default convergence criteria are applied for each of the different simulation modes described above.

**Full support for CPUs and GPUs**
All of the features and modifications in this PR has been verified for the ``CPUSolver``, ``VectorizedSolver`` and ``GPUSolver`` in both parallel and sequential modes.

**Reduced memory footprint**
The number of class attributes for the ``Solver`` and its subclasses has been greatly reduced to minimize the memory footprint on both CPUs and GPUs. This is also intended to make the code for each of the different ``Solver`` subclasses to be more digestible for the reader. As a result of some of these changes, some items are computed on the fly (*e.g.*, the fission source) when needed rather than stored in a class attribute. It is important to note that this decreases the memory footprint at the expense of a large flop count for eigenvalue calculations (~1-2% depending on energy group structure).

**New sample input files**
Two new sample input files were introduced to illustrate the new fixed source API and simulation modes:

* ``sample-input/point-source`` - computes the flux for a fixed point source
* ``sample-input/subcritical`` - computes the source for a fixed point source next to a 2D slab of UO2


sample-input/benchmarks/c5g7/c5g7-cmfd.py
```diff
@@ -33,7 +34,7 @@
 
 
 ###############################################################################
-###########################   Creating Surfaces   #############################
+#                            Creating Surfaces
 ###############################################################################
```
PrezNattyGibbs: I certainly have no problem with changing the output headers like this.  But is there a rationale behind this?  Just curious for consistency going forward.

wbinventor: Yeah I recognize that this is merely cosmetic and not a crucial piece of this PR. I simply changed them to this style to make it easier to pick out the text in the header rather than hide it with "#" symbols. Perhaps we should make a "style guide" for comments in our sample input files, I'm certainly open to suggestions.

sample-input/benchmarks/c5g7/c5g7-cmfd.py
```diff
@@ -318,28 +319,29 @@
...
+#                            Running a Simulation
 ###############################################################################
 
-solver = CPUSolver(geometry, track_generator)
-solver.setSourceConvergenceThreshold(tolerance)
+solver = CPUSolver(track_generator)
+solver.setConvergenceThreshold(tolerance)
 solver.setNumThreads(num_threads)
-solver.convergeSource(max_iters)
+solver.computeEigenvalue(max_iters)
```
PrezNattyGibbs: For what it's worth, this is a much better name than `convergeSource` was.  Does this adequately imply that a whole eigenvalue problem is being solved, not just computation of an eigenvalue for a single iteration, though?

sample-input/point-source/point-source.py
```diff
@@ -0,0 +1,145 @@
...
+###############################################################################
+#                            Creating Lattices
+###############################################################################
+
+num_x = 200
+num_y = 200
+width_x = (root_universe.getMaxX() - root_universe.getMinX()) / num_y
+width_y = (root_universe.getMaxY() - root_universe.getMinY()) / num_x
+universes = [[water_univ]*num_x for _ in range(num_y)]
+universes[15][15] = source_univ
```
PrezNattyGibbs: Maybe change this to `num_x/2` to make changing the size of the problem a little more automatic?

sample-input/subcritical/subcritical.py
```diff
@@ -0,0 +1,155 @@
...
+#                            Creating Lattices
+###############################################################################
+
+num_x = 100
+num_y = 100
+width_x = (root_universe.getMaxX() - root_universe.getMinX()) / num_y
+width_y = (root_universe.getMaxY() - root_universe.getMinY()) / num_x
+universes = [[water_univ]*num_x for _ in range(num_y)]
+
+universes[15][15] = source_univ
```
PrezNattyGibbs: Same.

wbinventor:
I just updated this PR with the merger of #147. I did a brief confirmation that the merger didn't munge the ability for the ``CPUSolver``, ``VectorizedSolver`` and ``GPUSolver`` to properly solve the sample input files.

sample-input/benchmarks/c5g7/c5g7-cmfd.py
wbinventor: Are you asking a question, or stating your opinion that the method signature isn't descriptive enough? If it is not, any ideas on a better name?

sample-input/point-source/point-source.py
wbinventor: Change the placement of ``source_univ``? Do you mean something like:

```python
universes[num_x/2][num_x/2] = source_univ
```

I can see how this would make the input file work seamlessly with more/less mesh. My main intention was to offset the fixed source so that it wouldn't be centered in the lattice. So if the above is what you were recommending, I can come up with a way to combine both.

PrezNattyGibbs: Yeah, that's what I meant.  I just don't like using variables for one thing, but hard coded numbers for the other.  I'd even be fine with `source_x = 15` and such, doesn't have to be mathematically linked.

sample-input/benchmarks/c5g7/c5g7-cmfd.py
PrezNattyGibbs: Asking a question!  I'm not always good at finding the best option that lies between a FORTRAN77 style `DEIGPI` and a Java style `compute_that_eigenvalue_and_associated_eigenvalue_using_power_iteration`.  `computeEigenvalue` is fine with me, just wanted to make sure we adequately happy!

src/CPUSolver.cpp
```diff
@@ -38,132 +38,52 @@ int CPUSolver::getNumThreads() {
...
-               "to %d since it is less than or equal to 0", num_threads);
+  Solver::setFixedSourceByFSR(fsr_id, group, source);
 
-  _num_threads = num_threads;
+  /* Allocate the fixed sources array if not yet allocated */
+  if (_fixed_sources == NULL) {
+    int size = _num_FSRs * _num_groups;
+    _fixed_sources = new FP_PRECISION[size];
+    memset(_fixed_sources, 0.0, sizeof(FP_PRECISION) * size);
+  }
```
PrezNattyGibbs: This is a little simpler than my original implementation, as it keeps only a single fixed source array.  This is of course at the expense of memory, as I only allocated fixed source arrays for regions where it was nonzero.  All in all, I think I agree the simpler approach is the way to go, but we should keep in mind this is what is happening if running with a large problem spatially and in energy.

src/CPUSolver.cpp
```diff
@@ -38,132 +38,52 @@ int CPUSolver::getNumThreads() {
...
+ * @brief Assign a fixed source for a flat source region and energy group.
+ * @details Fixed sources should be scaled to reflect the fact that OpenMOC 
+ *          normalizes the scalar flux such that the total energy- and 
+ *          volume-integrated production rate sums to 1.0.
+ * @param fsr_id the flat source region ID
+ * @param group the energy group
+ * @param source the volume-averaged source in this group
  */
-void CPUSolver::setNumThreads(int num_threads) {
+void CPUSolver::setFixedSourceByFSR(int fsr_id, int group,
```
PrezNattyGibbs: My original implementation was with `addFixedSource...()` routines rather than `setFixedSource...` routines.  My thought process was that you may want to add a baseline source to a material and then an additional external source to a cell that includes that material.  That might be an unusual enough case that your implementation is more reasonable.  But maybe we should throw a warning if a fixed source is being overwritten?  Does that seem reasonable?

src/CPUSolver.cpp
```diff
@@ -226,48 +145,29 @@ void CPUSolver::initializeFluxArrays() {
...
-
-    size = _num_FSRs;
-    _old_fission_sources = new FP_PRECISION[size];
-    _source_residuals = new FP_PRECISION[size];
-
+    /* If no fixed sources were assigned, use a zeroes array */
+    if (_fixed_sources == NULL) {
+      _fixed_sources = new FP_PRECISION[size];
+      memset(_fixed_sources, 0.0, sizeof(FP_PRECISION) * size);
+    }
```
PrezNattyGibbs: This seems like a good solution for fixed source problems, but I don't like it happening for a large eigenvalue calculation.  Could we move this to the fixed source and subcritical solver routines rather than a fully shared routine?

src/CPUSolver.cpp
```diff
@@ -763,7 +725,7 @@ void CPUSolver::transferBoundaryFlux(int track_id,
     for (int p=0; p < _num_polar; p++) {
       track_out_flux(p,e) = track_flux(p,e) * bc;
       track_leakage(p,e) = track_flux(p,e) *
-                            _polar_weights(azim_index,p) * (!bc);
+                           _polar_weights(azim_index,p) * (!bc);
```
PrezNattyGibbs: Clearly wasn't intended to be a part of this PR, but is `!bc` a reasonable thing to put in a mathematical expression?  Wouldn't `1-bc` be more clear?

src/CPUSolver.cpp
```diff
@@ -782,23 +744,20 @@ void CPUSolver::addSourceToScalarFlux() {
...
     sigma_t = _FSR_materials[r]->getSigmaT();
 
-      for (int e=0; e < _num_groups; e++) {
-        _scalar_flux(r,e) *= 0.5;
-        _scalar_flux(r,e) = FOUR_PI * _reduced_sources(r,e) +
-                            (_scalar_flux(r,e) / (sigma_t[e] * volume));
+    for (int e=0; e < _num_groups; e++) {
+      _scalar_flux(r,e) *= 0.5;
+      _scalar_flux(r,e) /= (sigma_t[e] * volume);
+      _scalar_flux(r,e) += (FOUR_PI * _reduced_sources(r,e));
```
PrezNattyGibbs: Why this change?  I feel like this obscures the equation you're encoding.

src/CPUSolver.cpp
```diff
@@ -782,23 +744,20 @@ void CPUSolver::addSourceToScalarFlux() {
...
     }
   }
-
-  return;
 }
 
 
 /**
- * @brief Computes the volume-weighted, energy integrated fission rate in
+ * @brief Computes the volume-averaged, energy-integrated fission rate in
```
PrezNattyGibbs: This change makes these nu-fission rates now.

src/ExpEvaluator.h
```diff
@@ -10,9 +10,9 @@
 
 #ifdef __cplusplus
 #define _USE_MATH_DEFINES
-#include <math.h>
 #include "log.h"
 #include "PolarQuad.h"
+#include <math.h>
```
PrezNattyGibbs: Is this a meaningful change or just something strange that happened? :grinning:

src/Solver.cpp
```diff
@@ -315,16 +391,90 @@ void Solver::setPolarQuadrature(PolarQuad* polar_quad) {
...
+void Solver::setFixedSourceByFSR(int fsr_id, int group, FP_PRECISION source) {
+  
+  if (group <= 0 || group > _num_groups)
+    log_printf(ERROR,"Unable to set fixed source for group %d in "
+               "in a %d energy group problem", group, _num_groups);
+
+  if (fsr_id < 0 || fsr_id >= _num_FSRs)
+    log_printf(ERROR,"Unable to set fixed source for FSR %d with only "
+               "%d FSRs in the geometry", fsr_id, _num_FSRs);
+}
```
PrezNattyGibbs: Shouldn't this do something other than error check?

src/Solver.cpp
```diff
@@ -489,52 +655,212 @@ void Solver::initializeCmfd(){
...
- * @brief Checks that each FSR has at least one Track segment crossing it
- *        and if not, throws an exception and prints an error message.
- * @details This method is for internal use only and is called by the
- *          Solver::convergeSource() method and should not be called
- *          directly by the user.
+ * @brief Computes the scalar flux distribution by performing a series of 
+ *        transport sweeps.
+ * @details This is the main method exposed to the user through the Python
+ *          interface to compute the scalar flux distribution, e.g., for a 
+ *          fixed ource calculation. This routine makes an initial guess for
```
PrezNattyGibbs: typo: ource -> source

src/Solver.cpp
```diff
@@ -489,52 +655,212 @@ void Solver::initializeCmfd(){
...
-    if (FSR_segment_tallies[r] == 0) {
-      log_printf(ERROR, "No tracks were tallied inside FSR id = %d. Please "
-                 "reduce your track spacing, increase the number of "
-                 "azimuthal angles, or increase the size of the FSRs", r);
+/**
+ * @brief Computes the total source distribution by performing a series of 
+ *        transport sweep and source updates.
+ * @details This is the main method exposed to the user through the Python
+ *          interface to compute the source distribution, e.g., for a fixed
+ *          and/or external source calculation. This routine makes an initial
```
PrezNattyGibbs: These comments don't make it clear that `computeFlux` and `computeSource` both fully solve their respective problems.  Back to my previous comment of what to name all these routines, I think I do have an opinion for different names, pretty much change "compute" to "solve" and add some detail:

-`solveFixedSource`
-`solveSubcritical`
-`solveEigenvalue` 
...or something like that.

src/Solver.cpp
PrezNattyGibbs: Never mind, I see what you're doing.  You even explained it in the comment, so just ignore me here!

PrezNattyGibbs:
This looks like great work, and some great additions to the code.  And thanks for cleaning up the mess I started with the fixed source solver!  I still need to run the sample inputs to make sure everything runs, but aside from the minor comments I made, I think this is in good shape.

sample-input/point-source/point-source.py
wbinventor: Ok, see if you like what I did to make the placement of the fixed source more flexible with the mesh refinement.

sample-input/subcritical/subcritical.py
wbinventor: Same here - I made the placement of the fissionable material more dynamic as well.

src/CPUSolver.cpp
wbinventor: Yes, I agree. This should be added as a warning note when we update the docs with info on how to setup fixed source problems. Actually, it affects the memory footprint for both eigenvalue, subcritical and fixed source calculations so it should be a more general note. I do not think that it is something to worry about though since the memory required for track segments will typically dominate that for FSR-based array data, except perhaps for your for ultra-fine energy use cases.

src/CPUSolver.cpp
wbinventor: Ok, that is a good point, I hadn't considered that possibility. So you think that simply over-riding the source in the FSR and printing a ``WARNING`` message is sufficient? I just added the message. If you can think of a good use case which would make it prohibitively difficult to assign sources in this way, we could add an additional method parameter ``increment=False`` which would allow the user to increment any source already assigned to the FSR if they wish.

src/CPUSolver.cpp
wbinventor: Yes, multiplying by a boolean is fine in C/C++. Much simpler (and faster!) than using a conditional statement. Of course we could use ``1-bc`` instead. I'll change this here if both you and @samuelshaner think it necessary.

src/CPUSolver.cpp
wbinventor: Well the main reason was simply to break things up to make it (potentially) easier for the compiler to vectorize the loop.

wbinventor: That said, the compilers don't do a very good job of this so I can revert back to the old form if you wish.

src/ExpEvaluator.h
wbinventor: Yes it actually is a meaningful change. Including any of the standard C libraries prior to including "Python.h" leads to warning messages at compile time (see my discussion in PR #143 for more details). In this case, "log.h" includes "Python.h" first, which is then included here first. Unfortunately, Git didn't do a good job of ordering the includes when it merged together some of the old PRs.

src/Solver.cpp
wbinventor: Yeah, kind of weird I know, maybe my OO nature went a bit too far here.

src/Solver.cpp
wbinventor: Got it

src/Solver.cpp
wbinventor: Ok, I do see your point. I'll leave the method signatures as is for now till we get some more input from @samuelshaner and any others who care to chime in (@geogunow?)

src/CPUSolver.cpp
wbinventor: Yes, good call. I changed "fission rate" in the comments to "nu-fission rate" everywhere.

src/CPUSolver.cpp
PrezNattyGibbs: I like `1-bc` better just for clarity's sake, but it's not the most important issue for me.

src/Solver.cpp
PrezNattyGibbs: My opinion: this OO magic is a bit confusing.  I'd rename the helper routine and make it a private method.

src/Solver.cpp
```diff
@@ -315,16 +391,90 @@ void Solver::setPolarQuadrature(PolarQuad* polar_quad) {
...
-    log_printf(ERROR, "Unable to set the source convergence threshold to %f"
-               "since the threshold must be a positive number", source_thresh);
+  Material* fsr_material;
 
-  _source_convergence_thresh = source_thresh;
+  /* Add the source to all FSRs for this Material */
+  for (int r=0; r < _num_FSRs; r++) {
+    fsr_material = _geometry->findFSRMaterial(r);
+    if (material->getId() == fsr_material->getId())
+      setFixedSourceByFSR(group, r, source);
```
PrezNattyGibbs: Wrong order of the indices here!

src/Solver.cpp
```diff
@@ -315,16 +391,90 @@ void Solver::setPolarQuadrature(PolarQuad* polar_quad) {
...
+    CellBasic* fsr_cell;
+    
+    for (int r=0; r < _num_FSRs; r++) {
+      fsr_cell = _geometry->findCellContainingFSR(r);
+      if (cell->getId() == fsr_cell->getId())
+        setFixedSourceByFSR(r, group, source);
+    }
+  }
+}
+
```
PrezNattyGibbs: I just tried this for a cell with rings, and it looks like the source only made it into the outer ring.  I'm not really sure what went wrong.  I'll look into it more later if you don't first.

wbinventor: Yes, this is a tricky one. The reason that this occurs is that the
``Solver``  is blind to the ``Cell`` "clones" that are created for each
ring/sector during the ``Geometry::initializeFlatSourceRegions()`` routine.
The original "parent" ``Cell``  is kept as one ring/sector, but all other
``Cell`` clones have new IDs which correspond to different FSRs. Hence,
when you set the source for a particular ``Cell``, the ``Solver`` will only
register that source for that ``Cell`` and not all of its clones. This
leaves it to the user to figure out how to register sources for each
ring/sector which is a pain but can be done with some bookkeeping.

Perhaps one good option for resolving this would be some routines like
``Solver::setFixedSource(char* cell_name, FP_PRECISION source)``. This
would permit the user to leverage ``Cell`` string names which are inherited
by the clones. What do you think @PrezNattyGibbs?

PrezNattyGibbs: Seeing as we don't advertise the `Cell` clones to the user, I don't like the user having to figure this out.  It's also one of the most obvious uses of `setFixedSourceByCell` to apply it to a subdivided pin.  So I definitely think we need to make this automatic somehow.

Your solution works, and is probably the easiest way to do it.  The other idea I'm having is in the subdividing cell routine, we can replace the original `CellBasic` with a `CellFill` containing the rings and retaining the original pointer.  That has more applications than just this setting the source, as you can do things like grab all the cells in the pin when computing reaction rates.  But I haven't looked at the `ringify`/`sectorize` implementation enough to know if that'd be hard to pull off.

src/CPUSolver.cpp
PrezNattyGibbs: No strong opinion here, go with what makes you happy.

src/Solver.cpp
wbinventor: Wow, I really like this idea! I agree that this type of functionality would be more broadly useful as boiler-plate stuff to make generic pin cells easier to create / use in OpenMOC. My only question or concern is that I do not believe there is a way to turn a ``CellBasic`` which the user creates into a ``CellFill`` under the hood, but I'll see what I can do.

src/CPUSolver.cpp
wbinventor: Just changed it to ``1-bc``.

src/CPUSolver.cpp
wbinventor: Ok, I'm going to leave this as is unless someone else has a strong opinion against this change.

src/Solver.cpp
wbinventor: Great, thanks for catching that!

src/Solver.cpp
wbinventor: Alright, so it was painful, but I took your idea to heart and overhauled the ``Cell`` implementation in OpenMOC so that this will work now.  You should definitely review the latest code, particularly that in "Cell.h" and "Cell.cpp". 

There are no longer ``CellFill`` and ``CellBasic`` classes, but rather a single ``Cell`` class which can take a ``Material``, ``Universe`` or ``Lattice`` fill. The ``Cell`` class can automatically discretize itself into rings/sectors regardless of its fill type (previously it could only do so if it had a ``Material`` inside). When rings/sectors are created, a new fill ``Universe`` is created and each ring/sector sub-``Cell`` is added to the ``Universe``. 

This new ``Cell`` methodology will allow fixed sources to be assigned to the master/parent ``Cell`` created by the user, and the source will be assigned to each child sub-``Cell`` under-the-hood.

wbinventor: Whew, I have to be careful about taking your comments seriously! That was a lot of work, but it also greatly gutted and simplified the ``Cell`` code. Can't believe it took three years to realize that things could be simplified like that.

src/Solver.cpp
wbinventor: I'm going to leave this as is for now - I kind of like this as it is, but will revise if need be once we get a second opinion.

src/CPUSolver.cpp
wbinventor: You don't like the creation of fixed source arrays regardless of the type of calculation? I can see this, but since it takes no time and negligible memory (compared to everything else), I like the simplicity of having the initialization modularized like this. Do you see any particular reason that this might break some calculations?

src/Solver.cpp
PrezNattyGibbs: I don't know if I should apologize or pat myself on the back!  In any case, this sounds like great work, and I'll check out the details in the morning.

src/Solver.cpp
PrezNattyGibbs: That's fine with me.  I'm on record for calling it a bit confusing, but I admit you commented it well, and it's not totally unreasonable :smile:

src/CPUSolver.cpp
PrezNattyGibbs: Memory was my concern.  A big problem has lot of FSRs, and this seemed to me like a big footprint for carrying around extra zeros.  That said, a rough calculation puts it in the tens of MB on a full core, and that's probably not enough to lose too much sleep over.  Track data will probably dwarf it!

wbinventor:
@PrezNattyGibbs - have you had a chance to look this over with the new ``Cell`` implementation? I'd like to get this merged in soon before I forget it over the summer.

src/Cell.cpp
```diff
@@ -59,6 +51,9 @@ Cell::Cell(int id, const char* name) {
   _name = NULL;
   setName(name);
 
+  _num_rings = 0;
+  _num_sectors = 0;
```
PrezNattyGibbs: Shouldn't these initialize to 1?

src/Cell.cpp
```diff
@@ -250,6 +287,56 @@ std::map<int, surface_halfspace> Cell::getSurfaces() const {
...
+    if (univ_fill->getType() == SIMPLE)
+      nested_cells = univ_fill->getAllCells();
+    else
+      nested_cells = static_cast<Lattice*>(univ_fill)->getAllCells();
+
+    cells.insert(nested_cells.begin(), nested_cells.end());
+  }
+
+  return cells;
+}
```
PrezNattyGibbs: For what used to be a `CellBasic`, this should either return itself or throw an error.  Unless there's a reason to return an empty map?

src/Cell.cpp
```diff
@@ -250,6 +287,56 @@ std::map<int, surface_halfspace> Cell::getSurfaces() const {
...
+
+    std::map<int, Universe*> nested_universes;
+    if (univ_fill->getType() == SIMPLE)
+      nested_universes = static_cast<Universe*>(_fill)->getAllUniverses();
+    else
+      nested_universes = static_cast<Lattice*>(_fill)->getAllUniverses();
+    universes.insert(nested_universes.begin(), nested_universes.end());
+  }
+
+  return universes;
```
PrezNattyGibbs: I think the empty map makes sense here, but seeing as it's a related issue, I thought I'd put a comment here, too!

src/Cell.cpp
```diff
@@ -453,123 +591,21 @@ double Cell::minSurfaceDist(Point* point, double angle,
...
-
-
-/**
- * @brief Returns an empty std::map of Cell IDs and Cell pointers.
- * @return empty std::map of Cell IDs and pointers
- */
-std::map<int, Cell*> CellBasic::getAllCells() {
-  std::map<int, Cell*> cells;
-  return cells;
-}
```
PrezNattyGibbs: I guess you kept the previous functionality.  It just doesn't make much sense to me!

src/Cell.cpp
```diff
@@ -583,8 +619,9 @@ CellBasic* CellBasic::clone() {
 
 /**
  * @brief Subdivides the Cell into clones for fuel pin angular sectors.
+ * @param subcells an empty vector to store all subcells
  */
-void CellBasic::sectorize() {
+void Cell::sectorize(std::vector<Cell*>* subcells) {
```
PrezNattyGibbs: We don't usually pass pointers to vectors, just the vectors themselves.  What's the rationale here?

PrezNattyGibbs:
Looks great to me!  Just had a few OCD comments for you.  Appease me, and then you have my thumbs up!

src/Cell.cpp
wbinventor: Well, the ``sectorize`` and ``ringify`` routine actually treat 0 and 1 the same way. If you think 1 is more intuitive, I'll make that change.

src/Cell.cpp
wbinventor: Ok, so is there anything I need to do to appease you with these empty maps, or do they make sense?

src/Cell.cpp
wbinventor: cool beanz

src/Cell.cpp
wbinventor: Well, passing the vector itself would do an *implicit copy* of the vector which we do not want here. In other words, the vector that is used by the ``sectorize`` routine will be a deep copy of the one that was passed into it as a parameter. The ``sectorize`` routine would then modify / extend the vector copy with ``Cell`` rings, but the caller's vector would be left untouched. There are two ways to ensure that the caller receives the updated vector - 1) pass in a vector pointer as done here, or 2) return the updated vector copy from ``sectorize``. The first option requires less overhead (memory and time to make the vector copy) so I went with it here. But I agree with you that in most places in the code it has not been necessary to use vector / map pointers, but that is simply because our vector / map variables are typically private and only used by a particular class instance (rather than being passed around between instances).

wbinventor:
Great, I think that if this passes your review we're just down to @samuelshaner.

sample-input/bundled-lattice/bundled-lattice.py
```diff
@@ -48,18 +48,18 @@
...
 ###############################################################################
-#############################   Creating Cells   ##############################
+#                             Creating Cells
 ###############################################################################
 
 log.py_printf('NORMAL', 'Creating cells...')
 
-# Create a list of 10 CellBasic instances
-cells = [CellBasic() for i in range(12)]
+# Create a list of 10 Cell instances
```
samuelshaner: Should be `12 Cell instances`

sample-input/subcritical/subcritical.py
```diff
@@ -0,0 +1,172 @@
...
+
+###############################################################################
+#                            Creating Surfaces
+###############################################################################
+
+log.py_printf('NORMAL', 'Creating surfaces...')
+
+left = XPlane(x=-5.0, name='left')
+right = XPlane(x=5.0, name='right')
+bottom = YPlane(y=-5.0, name='top')
```
samuelshaner: `name` attribute for top and bottom surfaces should be switched.

src/CPUSolver.cpp
```diff
@@ -355,111 +262,184 @@ void CPUSolver::normalizeFluxes() {
...
+
+      if (old_total_source > 0.)
+        residuals[r] = pow((new_total_source -  old_total_source) / 
+                            old_total_source, 2);
+    }
+  }
+
+  /* Sum up the residuals from each FSR and normalize */
+  residual = pairwise_sum<double>(residuals, _num_FSRs);
+  residual = sqrt(residual / norm);
```
samuelshaner: I would recommend having `norm` set to the number of FSRs with an `old_total_source` greater than 0.

src/CPUSolver.cpp
```diff
@@ -355,111 +262,184 @@ void CPUSolver::normalizeFluxes() {
...
+    }
   }
 
-  /* Sum up the residuals from each FSR */
-  source_residual = pairwise_sum<FP_PRECISION>(_source_residuals, _num_FSRs);
-  source_residual = sqrt(source_residual \
-                         / (_num_fissionable_FSRs * _num_groups));
+  else if (res_type == TOTAL_SOURCE) {
+
+    norm = _num_FSRs * _num_groups;
```
samuelshaner: I would recommend having `norm` for `TOTAL_SOURCE` be set to the number of FSRs with an `old_total_source` greater than 0. Any reason why you think it should be `_num_FSRs * _num_groups`?

src/CPUSolver.cpp
```diff
@@ -355,111 +262,184 @@ void CPUSolver::normalizeFluxes() {
...
+        if (_old_scalar_flux(r,e) > 0.) {
+          residuals[r] += pow((_scalar_flux(r,e) - _old_scalar_flux(r,e)) / 
+                               _old_scalar_flux(r,e), 2);
+      }
     }
+  }
+
+  else if (res_type == FISSION_SOURCE) {
+
+    norm = _num_fissionable_FSRs * _num_groups;
```
samuelshaner: As in `TOTAL_SOURCE`, I would recommend removing `_num_groups` since the fission source is integrated over groups. This is solved in PR #146, but you might just fix it here.

src/CPUSolver.cpp
```diff
@@ -355,111 +262,184 @@ void CPUSolver::normalizeFluxes() {
...
+double CPUSolver::computeResidual(residualType res_type) {
+
+  int norm;
+  double residual;
+  double* residuals = new double[_num_FSRs];
+  memset(residuals, 0., _num_FSRs * sizeof(double));
+
+  if (res_type == SCALAR_FLUX) {
+
+    norm = _num_FSRs * _num_groups;
```
samuelshaner: ditto on removing `_num_groups`.

src/Cell.cpp
samuelshaner: I'll second @PrezNattyGibbs on initializing to 1.

src/Solver.cpp
samuelshaner: I'm fairly indifferent to the name choice, but if I had to pick I would vote for `solve...` instead of `compute...`.

src/accel/cuda/GPUSolver.cu
```diff
@@ -1476,53 +1335,206 @@ void GPUSolver::addSourceToScalarFlux() {
...
+
+  int norm;
+  double residual;
+  isinf_test inf_test;
+  isnan_test nan_test;
+  thrust::device_vector<double> residuals;
+
+  if (res_type == SCALAR_FLUX) {
+
+    norm = _num_FSRs * _num_groups;
```
samuelshaner: ditto on using `_num_groups` in norms in GPUSolver.

src/CPUSolver.cpp
wbinventor: No, I was just keeping to the existing normalization constant. But I realize that this is erroneous since the groups are integrated out in the residual calculation. I'll revise the norm for all convergence criteria momentarily.

wbinventor:
I've addressed your comments @samuelshaner.

Separately, I wanted to note that I've observed some interesting convergence behaviour for the fixed source and subcritical sample problems. In particular, with the default settings of 4 angles and 0.1 cm track spacing, these problems seem to diverge within ~10-20 iterations. With more angles and / or a finer track spacing, these issues disappear. My hypothesis is that the residual explosion may be related to ray effects, which you can easily see from the flux plots for these problems. If this is indeed the case, then we may consider including some sort of "divergence check" with an error message in a future PR.

wbinventor:
On a related note, I'll point out that I did not get CMFD working for fixed source calculations (actually, I didn't try). But with CMFD, I would not expect to see this divergence issue if my hypothesis above is correct.

src/Solver.cpp
```diff
@@ -489,52 +655,212 @@ void Solver::initializeCmfd(){
...
+  _k_eff = k_eff;
+  FP_PRECISION residual;
+
+  /* Initialize data structures */
+  initializePolarQuadrature();
+  initializeExpEvaluator();
+  initializeFluxArrays();
+  initializeSourceArrays();
+  initializeFSRs();
+
```
samuelshaner: You should add `countFissionableFSRs();` in case the user wants to use `FISSION_SOURCE` for convergence.

src/Solver.cpp
```diff
@@ -489,52 +655,212 @@ void Solver::initializeCmfd(){
...
+   * only fixed sources or b) no previous simulation was performed which
+   * initialized and computed the flux (e.g., an eigenvalue calculation) */
+  if (only_fixed_source || _num_iterations == 0) {
+    initializeFluxArrays();
+    flattenFSRFluxes(0.0);
+  }
+
+  initializeSourceArrays();
+  initializeFSRs();
+  zeroTrackFluxes();
```
samuelshaner: Should add `countFissionableFSRs();` here too.

samuelshaner:
I noticed the same behavior of divergence for the subcritical case. In some cases, I would get NaNs for the first residual and it would stop there. After trying a bunch of things I seem to have found the issue and it is has to do with the exponential calculation. The exponential calculation uses the factor 1.01 to expand the size of the exponential table. The exponential table needs to be expanded because the interpolation uses the nearest value in the table and the value corresponding to the next largest optical path length. Since `tau * _inverse_exp_table_spacing` is being rounded to an int instead of floored (see pics below), half of the exponential interpolations with actually be extrapolations from the interval one larger. If the nearest value in the table is the last value, the exponential will return junk since it tries to use a value that does not exist in the table. The `_exp_precision` is quite large so the table has something like 50 values. Therefore, some of the regions with a large optical path length with give junk for the exponentials. What the `computeExponential` function should actually be doing is flooring `tau * _inverse_exp_table_spacing` and converting that value to an int:

before: `index = rount_to_int(tau * _inverse_exp_table_spacing);`
after: `index = (int) floor(tau * _inverse_exp_table_spacing);`

 Making this change seemed to resolve the divergence issue. The convergence will be nonuniform for the default case (and I believe that is to be expected for the particular geometry being studied), but the problem does converge. Make sense?

exp_calc_round
exp_calc_floor

src/Solver.cpp
wbinventor: Great catch, I didn't think I needed it here at the time I implemented this method. But then I went on to add the different convergence criteria, and I see that now it is necessary to count fissionable FSRs in case the criterion is ``FISSION_SOURCE``.

src/Solver.cpp
```diff
@@ -454,15 +602,34 @@ void Solver::initializeFSRs() {
...
+  std::map<int, Material*> all_materials = _geometry->getAllMaterials();
+  _num_fissionable_FSRs = 0;
+
+  for (int r=0; r < _num_FSRs; r++) {
+    if (_FSR_materials[r]->isFissionable())
+      _num_fissionable_FSRs++;
+  }
 
-  return;
+  if (_num_fissionable_FSRs == 0)
```
samuelshaner: If `countFissionableFSRs` is called for all `compute...` functions, there will be cases where the `_num_fissionable_FSRs` is 0 (e.g. `point-source.py`). `_num_fissionable_FSRs` being 0 only seems to be a problem if the residual type is `FISSION_SOURCE`. Perhaps you could move this check to the `computeResidual` function inside if the `FISSION_SOURCE` if statement?

wbinventor:
Excellent (subtle) point about the exponential table indexing bug! Wow, can't believe that lingered around for so long. When I implemented the ``GPUSolver`` with the fast intrinsic [``__double2int_rd``](http://developer.download.nvidia.com/compute/cuda/4_1/rel/toolkit/docs/online/group__CUDA__MATH__INTRINSIC__CAST_gf91bc76a892f92d470595c3d95703531.html) routine for this, I tried to find something analogous for the ``CPUSolver``. It looks like I didn't do my homework well enough and unfortunately settled on ``lrint``. Although C++11 provides for rounding modes (such as ``FE_DOWNWARD``), I opted instead to go with a simple call to ``floor(...)`` as you recommended. I wanted to avoid this as it can reportedly be quite slow (esp. within our innermost transport sweep loop), but we can let the SimpleMOC developers figure out the fastest implementation and migrate it into OpenMOC, right @geogunow?

wbinventor:
Oh and @samuelshaner I just pushed the fix, plus removed the "1.01" exponential table size multiplier since that was my very hacky attempt to get around this bug at one point (for the life of me I could not track down the culprit of this bug as you did). If you think that we still need to arbitrarily expand the size of the table for some reason, let me know.

src/Solver.cpp
wbinventor: You mean within the ``CPUSolver::computeResidual(...)`` and ``GPUSolver::computeResidual(...)`` routines?

wbinventor:
Also, I just realized that I didn't point out in my description for this PR that I heavily revised the ``GPUSolver`` to not only reflect the new features required by the ``Solver`` interface, but also to make much heavier use of the [Thrust](http://docs.nvidia.com/cuda/thrust/) library for CUDA. The reason for this is that Thrust is heavily optimized (by NVIDIA software engineers) and the more we can use it for book-keeping computations between transport sweeps, the better. In particular, it will allow NVIDIA to "optimize our code for us" as Thrust is updated in parallel with each new generation of GPU boards. 

Plus...I hope that some of the cool stuff that I am now doing with Thrust might serve as an example for how we might construct CMFD matrices on the GPU in a fast and efficient way someday.

samuelshaner:
We still need the exponential table multiplier, but it just needs to increase the table size enough so the maximum optical path length is floored to the second to last value. Something like `1 + 1.e-5` should be sufficient for single and double precision. I removed the multiplier completely and ran `subcritical.py` and saw the same behavior as before.

wbinventor:
Oh, to account for the case when a segment with the max optical length would be floored to one element outside of the table?

samuelshaner:
The GPUSolver stuff sounds awesome!!! I can't say that I understand much of the details, but I'm excited to dig into it and learn from what you did.

wbinventor:
Ok, so I think I've made all of the changes we've discussed to this point.

samuelshaner:
Yes, but the max optical path length segment will be floored to the last element. Then the ExpEvaluator will use the last element and one element outside the table for interpolation.

src/CPUSolver.cpp
```diff
@@ -355,111 +262,188 @@ void CPUSolver::normalizeFluxes() {
...
+                               _old_scalar_flux(r,e), 2);
+      }
     }
+  }
+
+  else if (res_type == FISSION_SOURCE) {
+
+    if (_num_fissionable_FSRs == 0)
+      log_printf(ERROR, "The Solver is unable to compute the "
+                 "eigenvalue without fissionable FSRs");
```
samuelshaner: You might rephrase this error statement (and the one in the GPUSolver) since the problem doesn't have to be computing the eigenvalue.

wbinventor: Wow, embarrassing - coding while waiting on a ride after a redeye is a bad idea. I *think* this is fixed now.

src/accel/cuda/GPUSolver.h
```diff
@@ -14,22 +14,33 @@
...
+#include <thrust/fill.h>
+#include <thrust/reduce.h>
+#include <thrust/replace.h>
+#include <thrust/functional.h>
+#include <thrust/iterator/constant_iterator.h>
 #include <sm_20_atomic_functions.h>
 #include "clone.h"
 #include "GPUExpEvaluator.h"
 
+#define PySys_WriteStdout printf
```
samuelshaner: Can you move this line back to before the include statements. Moving this line to after the include statements produced the error included below when compiling with CUDA 7. It compiled just fine with CUDA 6 though. When I moved this line back to before the include statements, it compiled just fine with CUDA 6 and 7.

/usr/local/cuda/include/thrust/system/cuda/detail/bulk/detail/terminate.hpp(45): error: namespace "std" has no member "PySys_WriteStdout"

1 error detected in the compilation of "/tmp/tmpxft_00001715_00000000-10_GPUSolver.compute_30.cpp1.ii".
error: command 'ccache' failed with exit status 2

wbinventor: You mean put it after the Python.h include? But before the other includes? I get the following error with CUDA 5.5 and 6.5 if I put it before all of the includes:

```bash
nvcc -DFP_PRECISION=double -DDOUBLE -DCUDA -DCCACHE_CC=nvcc -I/usr/local/cuda/include -I/usr/lib/python2.7/dist-packages/numpy/core/include -I/usr/include/python2.7 -c src/accel/cuda/GPUSolver.cu -o build/temp.linux-x86_64-2.7/src/accel/cuda/GPUSolver.o --relocatable-device-code true -c -O3 --compiler-options -fpic -gencode=arch=compute_20,code=sm_20 -gencode=arch=compute_30,code=sm_30
/usr/include/python2.7/sysmodule.h(17): error: cannot overload functions distinguished by return type alone

1 error detected in the compilation of "/tmp/tmpxft_0000676b_00000000-9_GPUSolver.compute_30.cpp1.ii".
error: command 'nvcc' failed with exit status 2
```

wbinventor: See if my latest commit works for CUDA 6 and 7, it works for me with 5.5 and 6.5.

wbinventor: I hate how they had to revise their ``printf`` calls in CUDA, makes it more difficult to SWIG now.

samuelshaner:
Ya, you're right - after the Python.h include but before the other includes. Who knew it would be so hard just to get something to print.

samuelshaner:
@wbinventor this is an awesome PR and adds some really cool features! Everything looks good to me so I'll merge this beast in and move to the other PRs.

wbinventor:
There is a good explanation for the error - namely, CUDA now uses
"std::printf" rather than simply "printf". We want to use the macro to
replace all printfs with the Python wrapper so the output will be visible
from IPython. However, the C macro standard does not permit the use of
semicolons ":" in a macro definition. When we include Python.h, a macro is
included which tries to replace all printfs with PySys_WriteStdout, which
results in calls to std::PySys_WriteStdout which does not exist. Hence, our
new macro will override the one from Python.h, but this will only work
*after* Python.h is included.

samuelshaner:
ahhh I see. Glad you were able to resolve this tricky issue.

PR #144, opened 2015-05-01
Generate FSR numerical centroids

samuelshaner:
This PR implements the generation of the numerical centroids of the flat source regions. The numerical centroids will be useful in extending OpenMOC to accommodate linear source regions using the methodology implemented in CASMO 5 and published by Rodolfo Ferrer. The formula for computing the centroids was taken from Ferrer's 2012 PHYSOR paper "Linear Source Approximation in CASMO 5". Additionally, the centroids could be useful in implementing strategies to stabilize or accelerate the CMFD acceleration process.

A plotting routine has been added to plot the centroids over the FSRs using the command:

`plotter.plot_centroids(geometry)`

Below is an example of the centroids plot for `simple-lattice.py`:

centroids

This PR builds on PR #134 and should not be merged until (at least) PR #134 has been merged.


sample-input/pin-cell/pin-cell.py
```diff
@@ -106,6 +106,11 @@
 solver = CPUSolver(geometry, track_generator)
 solver.setNumThreads(num_threads)
 solver.setSourceConvergenceThreshold(tolerance)
+
```
wbinventor: Any reason to use this polar quadrature here?

src/CPUSolver.cpp
```diff
@@ -185,6 +185,28 @@ void CPUSolver::setNumThreads(int num_threads) {
 
 
 /**
+ * @brief Initializes the FSR volumes and Materials array.
+ * @details This method assigns each FSR a unique, monotonically increasing
+ *          ID, sets the Material for each FSR, and assigns a volume based on
+ *          the cumulative length of all of the segments inside the FSR.
+ */
+void CPUSolver::initializeFSRs() {
```
wbinventor: It looks like you already synchronized with my PR #134?

src/Geometry.cpp
```diff
@@ -643,6 +647,27 @@ Point* Geometry::getFSRPoint(int fsr_id) {
...
+Point* Geometry::getFSRCentroid(int fsr_id) {
+
+  Point* point;
+
+  try{
+    point = _FSR_keys_map.at(_FSRs_to_keys.at(fsr_id))._centroid;
+  }
+  catch(std::exception &e) {
+    log_printf(ERROR, "Could not find centroid in FSR: %i. "
+               "Backtrace:%s", fsr_id, e.what());
```
wbinventor: I've tried to remove the backtraces from the error messages, since the logger module throws an exception anyway. This would look like:

```C++
catch(std::exception &e) {
  log_printf(ERROR, "Could not find centroid in FSR: %i.", fsr_id);
```

sample-input/pin-cell/pin-cell.py
samuelshaner: Oops, I had that in to test your polar quadrature refactorization. I'll take it out.

wbinventor:
This is a pretty cool PR! I'm going to wait to review the rest of it until after PR #134 has been finalized, since it appears that most of the diffs are from that PR. In reality, I think it might be best to put this in the "back of the line" since some of my other PRs may affect it, in particular #141 and #137. Also, I would recommend merging with (or finalizing) PR #138 and #142 before using the centroids in CMFD or the MOC solve.

src/CPUSolver.cpp
samuelshaner: Correct, I merged with your `fsr-volume-getter` branch so PR #134 should be merged first. If more of your PRs are merged into `develop`, I'll continue to update this PR.

src/Geometry.cpp
samuelshaner: Got it. I'll make that fix.

samuelshaner:
Totally agree that this can wait until PR #137, #138, #141, and #142 are merged. I just haven't had the time to look over those yet and didn't want to go through the process of merging with all the changes of those pending PRs at this time. I'll continue to merge with develop as those PRs are merged in.

wbinventor:
I'll review this PR once it has been updated with respect to the current "develop" branch.

samuelshaner:
@wbinventor I updated this PR to reflect the changes in the current "develop" branch so it's ready for your review. No big rush if you are busy with other stuff.

openmoc/plotter.py
```diff
@@ -515,6 +515,139 @@ def plot_flat_source_regions(geometry, gridsize=250, xlim=None, ylim=None):
...
+  plt.imshow(colors, extent=coords['bounds'],
+             interpolation='nearest', cmap=cmap, vmin=0, vmax=num_fsrs)
+
+  # Plot centroids on top of 2D FSR color map
+  for r in range(geometry.getNumFSRs()):
+    point = geometry.getFSRCentroid(r)
+    x = point.getX()
+    y = point.getY()
+    
+    if xlim is None and ylim is None:
```
wbinventor: Is this conditional actually needed? Will matplotlib actually try to plot a centroid if it is outside the bounds of the original image?

openmoc/plotter.py
```diff
@@ -515,6 +515,139 @@ def plot_flat_source_regions(geometry, gridsize=250, xlim=None, ylim=None):
...
+    elif xlim is not None:
+      if x >= xlim[0] and x <= xlim[1]:
+        centroids_x.append(x)
+        centroids_y.append(y)
+    else:
+      if y >= ylim[0] and y <= ylim[1]:
+        centroids_x.append(x)
+        centroids_y.append(y)
+        
+  plt.plot(centroids_x, centroids_y, 'ko', markersize=2)
```
wbinventor: Can we make the centroid marker type and size be keyword arguments to this routine?

openmoc/plotter.py
```diff
@@ -515,6 +515,139 @@ def plot_flat_source_regions(geometry, gridsize=250, xlim=None, ylim=None):
...
+  
+  plt.title('Centroids')
```
wbinventor: I think the title should be "FSR Centroids" to be more exact perhaps?

openmoc/plotter.py
```diff
@@ -515,6 +515,139 @@ def plot_flat_source_regions(geometry, gridsize=250, xlim=None, ylim=None):
...
+# @endcode
+#
+# @param geometry a geometry object which has been initialized with Materials,
+#        Cells, Universes and Lattices
+# @param gridsize an optional number of grid cells for the plot
+# @param xlim optional list/tuple of the minimim/maximum x-coordinates
+# @param ylim optional list/tuple of the minimim/maximum y-coordinates
+def plot_centroids(geometry, gridsize=250, xlim=None, ylim=None):
+
+  global subdirectory
```
wbinventor: After looking at this routine, is it really necessary? Perahsp an optional ``centroids=False`` keyword param for the ``plot_flat_source_regions(...)`` routine would better for code reuse?

src/Geometry.cpp
```diff
@@ -613,7 +617,28 @@ Point* Geometry::getFSRPoint(int fsr_id) {
     point = _FSR_keys_map.at(_FSRs_to_keys.at(fsr_id))._point;
   }
   catch(std::exception &e) {
-    log_printf(ERROR, "Could not find characteristic point in FSR %d", fsr_id);
+    log_printf(ERROR, "Could not find characteristic point in FSR: %i.",
```
wbinventor: Why change from ``%d`` to ``%i`` here?

src/Geometry.h
```diff
@@ -143,6 +146,7 @@ class Geometry {
   void setFSRsToKeys(std::vector<std::size_t> FSRs_to_keys);
   void setNumFSRs(int num_fsrs);
   void setCmfd(Cmfd* cmfd);
+  void setFSRCentroid(int fsr, Point* centroid);
```
wbinventor: Should this be public?

src/Geometry.h
```diff
@@ -34,9 +34,11 @@ struct fsr_data {
   /** The FSR ID */
   int _fsr_id;
 
-  /** Characteristic point in Universe 0 that lies in FSR */
+  /** Characteristic point in Root Universe that lies in FSR */
   Point* _point;
```
wbinventor: Is it necessary to have both a "characteristic point" and a centroid? It seems now that we have the centroid calculator that we can simply use it as the characteristic point (and just use the ``_centroid`` attribute since it is a more intuitive naming convention).

wbinventor: I just realized that the ``fsr_data`` struct does not have a destructor. This means that the ``_point`` and ``_centroid`` attributes will just be left dangling after an ``fsr_data`` is deleted from the ``Geometry`` (*e.g.*, when running many simulations during a parametric study).

src/TrackGenerator.cpp
```diff
@@ -1554,6 +1554,67 @@ void TrackGenerator::correctFSRVolume(int fsr_id, FP_PRECISION fsr_volume) {
...
+
+        x += cos(phi) * curr_segment->_length;
+        y += sin(phi) * curr_segment->_length;
+      }
+    }
+  }
+
+  /* Set the centroid for the FSR */
+  for (int r=0; r < num_FSRs; r++){
+    _geometry->setFSRCentroid(r, centroids[r]);
```
wbinventor: Maybe it is just me, but when I see a "setter" routine taking in a pointer, I assume that there are some OO trust boundaries are setup such that the underying class/struct will retain that pointer. Instead, it looks like the ``Geometry::setFSRCentroid(...)`` routine initializes a new ``Point``, copies the data over from the point passed in, and then ``TrackGenerator::generateFSRCentroids()`` deletes the ``Points`` here. 

What about passing in pointers to the ``Points`` as you do here, and simply making each ``fsr_data`` struct point to the very same point in ``Geometry::setFSRCentroid(...)``? And then the ``fsr_data`` destructor will be in charge of deleting each ``Point``.

wbinventor:
@samuelshaner I tried this code out on my machine and it works well. I made a number of comments - once we hash through them I think this is ready to go.

openmoc/plotter.py
samuelshaner: That's a good idea - I'll update with that.

openmoc/plotter.py
samuelshaner: This conditional is needed in case someone wants to plot only the centroids for a certain region of the geometry. Say someone wants to plot the centroids for one pin cell in an assembly and inputs `xlim` and `ylim` accordingly. If this conditional was not present, the flat source regions would only be plotted for a few the pin cell, but the centroids would be plotted for the entire geometry.

Another option could be to remove the conditional entirely and have the following:

`if xlim is not None:`
`plt.xlim(xlim)`
`if ylim is not None:`
`plt.ylim(ylim)`

What do you think about changing the long condition to just setting the plot limits? The results are the same.

openmoc/plotter.py
samuelshaner: sure

openmoc/plotter.py
samuelshaner: Sure, since we could technically make centroids for cells and lattices.

src/Geometry.cpp
samuelshaner: Not sure, I'll change it back.

src/Geometry.h
samuelshaner: I thought about that, but there are cases where the centroid can be outside of the FSR (e.g. a ring with no sectors) so it is necessary to have a point that lies within the FSR to construct the linked list of coords that uniquely identifies the FSR. 

I'll add a destructor for the point and centroid.

src/Geometry.h
samuelshaner: Probably not; I'll make it private.

src/TrackGenerator.cpp
samuelshaner: That works - I change it to that. I was just thinking it would be better for the Geometry to instantiate its own centroid in case there are changes to OpenMOC in the future where the `setFSRCentroid` method is called with a point allocated on the heap and thus destroyed with the calling method is finished.

openmoc/plotter.py
wbinventor: I think whichever option results in the least amount of code is likely the best. But I see your point and don't have an issue with keeping it as is.

src/Geometry.h
wbinventor: oh yes that make sense now

samuelshaner:
@wbinventor I updated the PR with based on your comments and made a few other small changes. Here are the notable changes:
- Got rid of `plot_centroids` in favor of boolean in `plot_flat_source_regions` to add centroids on the FSR plot.
- Added destructor for the `fsr_data` struct.
- Modified the handling of `std::vector` and `std::map` objects from the `Geometry` so now they are set/returned by vector instead of by deep copy.
- Added destructor for the surface `_neighbors`.

openmoc/plotter.py
```diff
@@ -442,12 +448,32 @@ def plot_flat_source_regions(geometry, gridsize=250, xlim=None, ylim=None):
...
+  if marker_type not in matplotlib.markers.MarkerStyle().markers.keys():
+    py_printf('ERROR', 'Unable to plot the flat source regions since ' + \
+              'marker_type is not a valid marker (%d)', marker_type)
+
+  if not is_float(marker_size) and not is_integer(marker_size):
+    py_printf('ERROR', 'Unable to plot the flat source regions since ' + \
+              'marker_size is not an int or float', marker_size)
+
+  if marker_size <= 0:
+    py_printf('ERROR', 'Unable to plot the flat source regions since ' + \
```
wbinventor: remove "since" in this message string (copy and paste error)

openmoc/plotter.py
```diff
@@ -442,12 +448,32 @@ def plot_flat_source_regions(geometry, gridsize=250, xlim=None, ylim=None):
...
+  if not isinstance(marker_type, str):
+    py_printf('ERROR', 'Unable to plot the flat source regions since ' + \
+              'marker_type is a string')
+
+  if marker_type not in matplotlib.markers.MarkerStyle().markers.keys():
+    py_printf('ERROR', 'Unable to plot the flat source regions since ' + \
+              'marker_type is not a valid marker (%d)', marker_type)
+
+  if not is_float(marker_size) and not is_integer(marker_size):
+    py_printf('ERROR', 'Unable to plot the flat source regions since ' + \
```
wbinventor: Can the markersize be a float?

src/Geometry.cpp
```diff
@@ -613,7 +617,27 @@ Point* Geometry::getFSRPoint(int fsr_id) {
...
+ */
+Point* Geometry::getFSRCentroid(int fsr_id) {
+
+  Point* point;
+
+  try{
+    point = _FSR_keys_map.at(_FSRs_to_keys.at(fsr_id))._centroid;
+  }
+  catch(std::exception &e) {
+    log_printf(ERROR, "Could not find centroid in FSR: %i.", fsr_id);
```
wbinventor: This is still using "%i" instead of "%d" - I *don't* think this matters, but for consistency throughout the code I think "%d" is better

src/Geometry.cpp
```diff
@@ -1081,6 +1105,10 @@ bool Geometry::withinBounds(LocalCoords* coords){
 }
 
 
+void Geometry::setFSRCentroid(int fsr, Point* centroid){
```
wbinventor: This method needs as Doxygen comment block. And since this is a public method (which I think is dangerous, but I understand why), it should have a ``@details`` block which describes it specifically as a helper routine for the ``TrackGenerator`` and not as something for the user to call from Python.

Speaking of public vs. private methods, we could use the SWIG ``%ignore`` directive to enforce that routines like this one are not exposed within Python. What do you think about this @samuelshaner? This might also apply to the similar discussion we've had with @geogunow about a few questionably public routines needed for the ``TrackGenerator`` and ``Geometry`` to communicate with one another during ray tracing.

src/Surface.cpp
```diff
@@ -65,6 +65,14 @@ Surface::Surface(const int id, const char* name){
 Surface::~Surface() {
   if (_name != NULL)
     delete [] _name;
+
+  if (!_neighbors.empty()){
+    std::map<int, std::vector<Cell*>* >::iterator iter;
```
wbinventor: This would be simpler without the ``for`` loop if we simply use the fact that there are only two integer keys for each halfspace:

```C++
if (!_neighbors.empty()) {
  _neighbors[-1].clear();
  _neighbors[+1].clear();
  _neighbors.clear();
}
```

src/TrackGenerator.cpp
```diff
@@ -1554,6 +1557,64 @@ void TrackGenerator::correctFSRVolume(int fsr_id, FP_PRECISION fsr_volume) {
...
+        centroids[fsr]->setX(centroids[fsr]->getX() + _azim_weights[i] *
+                             (x + cos(phi) * curr_segment->_length / 2.0) *
+                             curr_segment->_length / FSR_volumes[fsr]);
+        centroids[fsr]->setY(centroids[fsr]->getY() + _azim_weights[i] *
+                             (y + sin(phi) * curr_segment->_length / 2.0) *
+                             curr_segment->_length / FSR_volumes[fsr]);
+
+        x += cos(phi) * curr_segment->_length;
+        y += sin(phi) * curr_segment->_length;
+      }
```
wbinventor: Hmm, I think the way the operations are ordered within this second loop, the final ``segment`` in each ``Track`` will have no contribution to a centroid. Shouldn't the final two lines to increment ``x`` and ``y`` in come before ``centroids[fsr]->setX(...)`` and ``centroids[fsr]->setY(...)```?

wbinventor:
@samuelshaner - just looked over this a second time - everything looks good save for a few minor comments. I'd like to get your thoughts about using SWIG ``%ignore`` for the public (but not so public) routines created for this PR (and elsewhere in the code), but otherwise am ready to merge this PR.

openmoc/plotter.py
samuelshaner: The markersize in Matplotlib is a float, but I think the user should have the option to input it as an int since this is how I have most commonly seen it.

src/TrackGenerator.cpp
samuelshaner: The `x` and `y` variables are initialized to the beginning of the track or equivalently the beginning of the first segment. For each segment, we need to know the segment midpoint, so when looping over the segments the midpoint is found by taking the segment starting point (`x` and `y`) and adding half the segment length times the appropriate sin/cos of the azimuthal angle. `x` and `y` are then incremented at the end of the loop so they become the starting point of the next segment. After the centroid is incremented for the last segment, `x` and `y` are set to the endpoint of the last segment. 

In short, I believe it is correct as is.

src/Geometry.cpp
samuelshaner: I like that compromise, but it might be wise to consider a cleaner solution in the future.

samuelshaner:
Updated the PR with the requested edits.

openmoc/plotter.py
wbinventor: I agree, just wondered since I have always used ints and didn't realize it would accept a float

src/Geometry.cpp
```diff
@@ -1040,9 +1064,9 @@ std::vector<int> Geometry::getFSRsToMaterialIDs() {
  *          are read from file to avoid unnecessary segmentation.  
  * @param FSR_keys_map map of FSR keys to FSR data
  */
-void Geometry::setFSRKeysMap(std::unordered_map<std::size_t, fsr_data> 
+void Geometry::setFSRKeysMap(std::unordered_map<std::size_t, fsr_data>*
```
wbinventor: Can we add this to the list of SWIG ``%ignore`` routines? I know this isn't a new method, but since it came up in our review here I think it would be good to add.

src/Geometry.cpp
```diff
@@ -1050,8 +1074,8 @@ void Geometry::setFSRKeysMap(std::unordered_map<std::size_t, fsr_data>
  * @brief Sets the _FSRs_to_keys vector
  * @param FSRs_to_keys vector of FSR key hashes indexed by FSR IDs
  */
-void Geometry::setFSRsToKeys(std::vector<std::size_t> FSRs_to_keys){
-  _FSRs_to_keys = FSRs_to_keys;
+void Geometry::setFSRsToKeys(std::vector<std::size_t>* FSRs_to_keys){
```
wbinventor: ditto SWIG ``%ignore``

src/Geometry.cpp
```diff
@@ -1059,8 +1083,8 @@ void Geometry::setFSRsToKeys(std::vector<std::size_t> FSRs_to_keys){
  * @brief Sets the _FSRs_to_material_IDs vector
  * @param FSRs_to_material_IDs vector mapping FSR IDs to cells
  */
-void Geometry::setFSRsToMaterialIDs(std::vector<int> FSRs_to_material_IDs){
-  _FSRs_to_material_IDs = FSRs_to_material_IDs;
+void Geometry::setFSRsToMaterialIDs(std::vector<int>* FSRs_to_material_IDs){
```
wbinventor: ditto SWIG ``%ignore``

src/Geometry.cpp
```diff
@@ -1080,6 +1104,22 @@ bool Geometry::withinBounds(LocalCoords* coords){
...
+ *          and the associated FSR data. _centroid is a point that represents
+ *          the numerical centroid of an FSR computed using all segments
+ *          contained in the FSR. This method is used by the TrackGenerator
+ *          to set the centroid after segments have been created. It is
+ *          important to note that this method is a helper function for the
+ *          TrackGenerator and should not be explicitly called by the user.
+ * @param fsr a FSR id
+ * @param centroid a Point representing the FSR centroid
+ */
+void Geometry::setFSRCentroid(int fsr, Point* centroid){
```
wbinventor: ditto SWIG ``%ignore`` unless you think this one is worthwhile to expose in Python

src/TrackGenerator.cpp
wbinventor: Ok, this makes sense now

wbinventor:
@samuelshaner - I'm ready to approve this once those last three routines are added to the list of SWIG ``%ignore`` routines.

wbinventor:
My tests for this PR all pass, but it needs to be synchronized with latest "develop" branch, specifically PR #151, so that I can build and test with Intel.

samuelshaner:
Thanks for the comments will. I'm in the Tetons this week with my friend
Andrew, but I'll make the edits as soon as I get a chance.

On Sunday, June 21, 2015, Will Boyd <notifications@github.com> wrote:

> My tests for this PR all pass, but it needs to be synchronized with latest
> "develop" branch, specifically PR #151
> <https://github.com/mit-crpg/OpenMOC/pull/151>, so that I can build and
> test with Intel.
>
> —
> <https://github.com/mit-crpg/OpenMOC/pull/144#issuecomment-113941460>.
>


-- 
Regards,
Sam Shaner
Graduate Student

samuelshaner:
@wbinventor I synchronized with the latest "develop" and added the three requested routines to the list of SWIG `%ignore` routines so this PR should be good to merge.

PR #145, opened 2015-05-02
K-nearest neighbor CMFD flux updating scheme

samuelshaner:
This PR implements a k-nearest neighbor algorithm to update the MOC flux based on the ratio of new-to-old flux in the k-nearest neighbor CMFD cells. In this algorithm, the k-nearest neighbor CMFD cells to a FSR are computed based on the distance between the numerical centroid of the CMFD cell and FSR. A weighted sum of the k-nearest CMFD cell flux ratios are used to update the FSR flux. The picture below shows the 3-nearest neighbor cells to a particular FSR along with the corresponding weights applied to the ratio of CMFD flux for each cell:

centroid-updating
 
The k-nearest neighbor CMFD flux updating scheme has been tested on C5G7 and shown increased stability and performance over the conventional flux updating scheme. Additionally, testing on C5G7 showed a window of stability for the k-nearest algorithm to be used without the nonlinear diffusion coefficient damping factor when 3-7 nearest neighbor cells were used. The plot below shows the performance of the k-nearest neighbor algorithm (with damping factor set to 1.0) compared with the conventional updating with a damping factor of 0.6 and unaccelerated MOC for 2D C5G7. Importantly, k-nearest updating reduced the iterations required to converge the source to 1e-5 from 21 iterations (conventional CMFD with 0.6 DF) to 12 iterations (3-nearest).

c5g7-conv-k-nearest

This PR should be placed at the end of the queue as it builds on PRs #134 and #144. I'll continue to merge with develop as other PRs are merged. I will also update the CMFD documentation when this PR is merged.


wbinventor:
@samuelshaner do you want me to start reviewing this soon? Or are you still tweaking it around a bit?

samuelshaner:
Just merged with the latest develop so it should be good to begin review.

src/Cmfd.cpp
```diff
@@ -959,12 +962,12 @@ void Cmfd::updateMOCFlux() {
           iter != _cell_fsrs.at(i).end(); ++iter) {
 
           /* Set new flux in FSR */
-            _FSR_fluxes[*iter*_num_moc_groups+h] = getFluxRatio(i,h)
+          _FSR_fluxes[*iter*_num_moc_groups+h] = getUpdateRatio(i,h, *iter)
```
wbinventor: inconsistent spacing between parameters ``i,h, *iter`` here and on the next line

src/Solver.cpp
```diff
@@ -649,6 +649,8 @@ void Solver::initializeCmfd() {
   _cmfd->setFSRMaterials(_FSR_materials);
   _cmfd->setFSRFluxes(_scalar_flux);
   _cmfd->setPolarQuadrature(_polar_quad);
+  _cmfd->setGeometry(_geometry);
+  _cmfd->generateKNearestStencils();
```
wbinventor: perhaps ``Cmfd`` needs a method like ``initialize()`` which calls both ``generateKNearestStencil()`` and ``initializeSurfaceCurrents()`` under-the-hood? This would also allow both of those method to be made private.

src/Cmfd.h
```diff
@@ -185,6 +206,7 @@ class Cmfd {
   void addFSRToCell(int cmfd_cell, int fsr_id);
   void updateBoundaryFlux(Track** tracks, FP_PRECISION* boundary_flux, 
                           int num_tracks);
+  void generateKNearestStencils();
```
wbinventor: see my comment below about making this routine private

src/Cmfd.h
```diff
@@ -203,10 +225,14 @@ class Cmfd {
   int convertFSRIdToCmfdCell(int fsr_id);
   std::vector< std::vector<int> > getCellFSRs();
   bool isFluxUpdateOn();
+  bool isCentroidUpdateOn();
   FP_PRECISION getFluxRatio(int cmfd_cell, int moc_group);
+  FP_PRECISION getUpdateRatio(int cmfd_cell, int moc_group, int fsr);
```
wbinventor: ditto, make both of these routines private is possible

src/Cmfd.h
```diff
@@ -157,6 +169,15 @@ class Cmfd {
   /** Flag indicating whether to update the MOC flux */
   bool _flux_update_on;
 
+  /** Flag indicating whether to us centroid updating */
```
wbinventor: typo: "us" -> "use"

src/Cmfd.h
```diff
@@ -157,6 +169,15 @@ class Cmfd {
...
+  bool _centroid_update_on;
+
+  /** Number of cells to used in updating MOC flux */
```
wbinventor: typo: "to used" -> "used"

src/Cmfd.h
wbinventor: also add "(default true)"

src/Cmfd.h
wbinventor: also add "(default 3)"

src/Cmfd.cpp
wbinventor: Also, is this setting the update ratio to the FSR fluxes array?

src/Cmfd.cpp
```diff
@@ -1750,22 +1774,278 @@ void Cmfd::setPolarQuadrature(PolarQuad* polar_quad) {
...
+
+      /* Create new stencil */
+      std::vector< std::pair<int, FP_PRECISION> > *stencil =
+        new std::vector< std::pair<int, FP_PRECISION> >;
+      _k_nearest_stencils[*iter] = (*stencil);
+
+      /* Get distance to all cells that touch current cell */
+      for (int j=0; j < 9; j++)
+        _k_nearest_stencils[*iter]
+          .push_back(std::make_pair<int, FP_PRECISION>
```
wbinventor: I would move the ``.push_back(`` to the top line, something like this to improve readability:

```c
_k_nearest_stencils[*iter].push_back(
    std::make_pair(int, FP_PRECISION>(
      int(j), getDistanceToCentroid(centroid, i, j)));
```

src/Cmfd.cpp
```diff
@@ -1750,22 +1774,278 @@ void Cmfd::setPolarQuadrature(PolarQuad* polar_quad) {
...
+                     (int(j), getDistanceToCentroid(centroid, i, j)));
+
+      /* Sort the distances */
+      std::sort(_k_nearest_stencils[*iter].begin(),
+                _k_nearest_stencils[*iter].end(), stencilCompare);
+
+      /* Remove non-existent cells */
```
wbinventor: By "non-existent cells", do you mean ghost cells for cells on the boundaries?

src/Cmfd.cpp
```diff
@@ -1750,22 +1774,278 @@ void Cmfd::setPolarQuadrature(PolarQuad* polar_quad) {
...
+
+  /* Loop over mesh cells */
+  for (int i = 0; i < _num_x*_num_y; i++){
+
+    /* Loop over FRSs in mesh cell */
+    for (iter = _cell_fsrs.at(i).begin();
+         iter != _cell_fsrs.at(i).end(); ++iter) {
+
+      /* Get centroid */
+      centroid = _geometry->getFSRCentroid(*iter);
```
wbinventor: It seems like ``*iter`` is used in quite a few places in this method. Could you add a local variable ``fsr_id = *iter`` and then use ``fsr_id`` instead of ``*iter`` to make this easier to comprehend at first sight?

src/Cmfd.cpp
```diff
@@ -1750,22 +1774,278 @@ void Cmfd::setPolarQuadrature(PolarQuad* polar_quad) {
...
+ *                             3 4 5
+ *                             0 1 2
+ *
+ *         where 4 is the given CMFD cell. If the cell is on the edge or corner
+ *         of the geometry and there are less than k nearest neighbor cells,
+ *         k is reduced to the number of neighbor cells for that instance.
+ */
+void Cmfd::generateKNearestStencils(){
+
+  std::vector< std::pair<int, FP_PRECISION> >::iterator iter2;
```
wbinventor: It's up to you, but if the code doesn't look too ugly, might help to rename these ``iterator`` local variables as ``fsr_iter`` and ``stencil_iter``, or ``f_iter`` and ``s_iter`` for short. I think we should make it a goal to do this throughout the code in places where we use ``iter``, ``iter2``, ..., but am open to alternative opinions.

src/Cmfd.cpp
```diff
@@ -1750,22 +1774,278 @@ void Cmfd::setPolarQuadrature(PolarQuad* polar_quad) {
...
+ *         k is reduced to the number of neighbor cells for that instance.
+ * @param cmfd_cell The cmfd cell containing the FSR.
+ * @param moc_group The MOC energy group being updated.
+ * @param fsr The fsr being updated.
+ * @return the ratio used to update the FSR flux.
+ */
+FP_PRECISION Cmfd::getUpdateRatio(int cmfd_cell, int moc_group, int fsr){
+
+  FP_PRECISION ratio = 0.0;
+  FP_PRECISION total_distance = 1.e-10;
```
wbinventor: Why initialize to this rather than simply 0.0?

src/Cmfd.cpp
```diff
@@ -1750,22 +1774,278 @@ void Cmfd::setPolarQuadrature(PolarQuad* polar_quad) {
...
+ *
+ *         where 4 is the given CMFD cell. If the cell is on the edge or corner
+ *         of the geometry and there are less than k nearest neighbor cells,
+ *         k is reduced to the number of neighbor cells for that instance.
+ * @param cmfd_cell The cmfd cell containing the FSR.
+ * @param moc_group The MOC energy group being updated.
+ * @param fsr The fsr being updated.
+ * @return the ratio used to update the FSR flux.
+ */
+FP_PRECISION Cmfd::getUpdateRatio(int cmfd_cell, int moc_group, int fsr){
```
wbinventor: This is pretty cool, I like the way this routine abstracts an otherwise complicated calculation.

src/Cmfd.cpp
```diff
@@ -1750,22 +1774,278 @@ void Cmfd::setPolarQuadrature(PolarQuad* polar_quad) {
...
+ *        get the distance from.
+ * @return the distance from the CMFD cell centroid to the FSR centroid.
+ */
+FP_PRECISION Cmfd::getDistanceToCentroid(Point* centroid, int cell,
+                                         int stencil_index){
+
+  int x = cell % _num_x;
+  int y = cell / _num_x;
+  FP_PRECISION dist_x, dist_y;
+  bool found = false;
```
wbinventor: This routine seems fine to me, just curious if it would be impossible (or just not worth it) to use one of the ``Point::distance(double x, double y)`` or ``Point::distianceToPoint(Point* point)`` routines here?

src/Cmfd.cpp
```diff
@@ -1824,6 +2104,28 @@ void Cmfd::updateBoundaryFlux(Track** tracks, FP_PRECISION* boundary_flux,
...
+
+
+/** @brief Set a number of k-nearest neighbor cells to use in updating
+ *         the FSR flux.
+ * @param k_nearest The number of nearest neighbor CMFD cells.
+ */
+void Cmfd::setKNearest(int k_nearest){
+
+  if (_k_nearest < 1 || k_nearest > 9)
+    log_printf(ERROR, "Unable to set CMFD k-nearest to %i. k-nearest "
```
wbinventor: Change to ``%d`` to be consistent with the rest of the code, unless you think there is good reason for us to start using ``%i`` throughout.

wbinventor:
@samuelshaner I've thoroughly reviewed and tested this out with both GNU and Intel compilers. Let me know what you think about my comments.

A few concerns - first, when I append the following line to "c5g7-cmfd.py",

```python
cmfd.setCentroidUpdateOn(False)
```

I get the following output:

```
[  NORMAL ]  Computing the eigenvalue...
[  NORMAL ]  Iteration 0:	k_eff = 1.282185	res = 9.994E-01
[  NORMAL ]  Iteration 1:	k_eff = 1.259458	res = 1.012E+00
[  NORMAL ]  Iteration 2:	k_eff = 1.219931	res = 1.011E-01
[  NORMAL ]  Iteration 3:	k_eff = 1.204382	res = 6.825E-02
[  NORMAL ]  Iteration 4:	k_eff = 1.194248	res = 3.664E-02
[  NORMAL ]  Iteration 5:	k_eff = 1.189630	res = 2.533E-02
[  NORMAL ]  Iteration 6:	k_eff = 1.187335	res = 1.173E-02
[  NORMAL ]  Iteration 7:	k_eff = 1.186387	res = 1.182E-02
[  NORMAL ]  Iteration 8:	k_eff = 1.185919	res = 4.676E-03
[  NORMAL ]  Iteration 9:	k_eff = 1.185772	res = 5.392E-03
[  NORMAL ]  Iteration 10:	k_eff = 1.185678	res = 2.014E-03
[  NORMAL ]  Iteration 11:	k_eff = 1.185666	res = 2.545E-03
[  NORMAL ]  Iteration 12:	k_eff = 1.185632	res = 8.652E-04
[  NORMAL ]  Iteration 13:	k_eff = 1.185633	res = 1.290E-03
[  NORMAL ]  Iteration 14:	k_eff = 1.185605	res = 4.840E-04
[  NORMAL ]  Iteration 15:	k_eff = 1.185607	res = 7.786E-04
[  NORMAL ]  Iteration 16:	k_eff = 1.185583	res = 4.430E-04
[  NORMAL ]  Iteration 17:	k_eff = 1.185585	res = 5.598E-04
[  NORMAL ]  Iteration 18:	k_eff = 1.185568	res = 4.011E-04
[  NORMAL ]  Iteration 19:	k_eff = 1.185570	res = 4.025E-04
[  NORMAL ]  Iteration 20:	k_eff = 1.185561	res = 2.712E-04
[  NORMAL ]  Iteration 21:	k_eff = 1.185558	res = 2.113E-04
[  NORMAL ]  Iteration 22:	k_eff = 1.185557	res = 7.958E-05
[  NORMAL ]  Iteration 23:	k_eff = 1.185547	res = 1.322E-04
[  NORMAL ]  Iteration 24:	k_eff = 1.185557	res = 2.940E-04
[  NORMAL ]  Iteration 25:	k_eff = 1.185538	res = 4.298E-04
[  NORMAL ]  Iteration 26:	k_eff = 1.185558	res = 6.251E-04
[  NORMAL ]  Iteration 27:	k_eff = 1.185533	res = 7.584E-04
[  NORMAL ]  Iteration 28:	k_eff = 1.185557	res = 9.123E-04
[  NORMAL ]  Iteration 29:	k_eff = 1.185531	res = 9.794E-04
[  NORMAL ]  Iteration 30:	k_eff = 1.185552	res = 1.048E-03
[  NORMAL ]  Iteration 31:	k_eff = 1.185532	res = 1.031E-03
[  NORMAL ]  Iteration 32:	k_eff = 1.185501	res = 9.416E-04
[  NORMAL ]  Iteration 33:	k_eff = 1.185518	res = 9.731E-04
[  NORMAL ]  Iteration 34:	k_eff = 1.185493	res = 4.828E-04
[  NORMAL ]  Iteration 35:	k_eff = 1.185496	res = 7.366E-04
[  NORMAL ]  Iteration 36:	k_eff = 1.185465	res = 6.322E-04
[  NORMAL ]  Iteration 37:	k_eff = 1.185544	res = 9.729E-04
[  NORMAL ]  Iteration 38:	k_eff = 1.185423	res = 2.698E-03
[  NORMAL ]  Iteration 39:	k_eff = -nan	res = 4.223E-03
[  NORMAL ]  Iteration 40:	k_eff = -nan	res = -NAN
```

It appears that Cmfd w/o neighbors 1) doesn't converge as fast as it did before and 2) eventually crashes with NaNs. I'm guessing this is probably just a very small subtle bug.

Secondly, from just a some rough testing, it appears that CMFD is quite a bit slower now than it was before. Of course with a fine mesh (azimuthal, track spacing) the MOC solve easily dominates the CMFD solve, but just wanted to confirm that this indeed what you have observed and expect.

src/Cmfd.cpp
samuelshaner: The `updateMOCFlux` method updates the FSR flux using the old and new CMFD flux. In short, it sets the new FSR flux to the old fsr flux times some ratio of the new-to-old CMFD flux in the cell containing and near the FSR. I'll change the comment to:

`/* Update FSR flux using ratio of new to old CMFD flux */`

If you think of something better, just let me know.

wbinventor: Sounds good to me.

src/Solver.cpp
samuelshaner: Great idea.

src/Cmfd.cpp
samuelshaner: I moved `.push_back` to the top line, but kept the `(` on the lower line to keep it from auto-aligning with the start of the parentheses.

src/Cmfd.cpp
samuelshaner: Correct, I'll make that more clear in the comment.

src/Cmfd.cpp
samuelshaner: sure

src/Cmfd.cpp
samuelshaner: I like that. I changed the iters to `fsr_iter` and `stencil_iter`.

src/Cmfd.cpp
samuelshaner: Good question. Let's say we are using only one nearest and the FSR centroid is exactly in the middle of the CMFD cell. The `total_distance` will then be 0.0 and we'll get a divide by zero error when computing the ratio.

src/Cmfd.cpp
samuelshaner: Running grep returns no calls to `Point::distance(double x, double y)` so I'll remove it.

src/Cmfd.cpp
samuelshaner: I'll change to `%d` throughout - I just have a habit of using `%i` in print statements.

samuelshaner:
@wbinventor thanks for reviewing the PR! I just update the PR with the recommend changes. CMFD did get quite a bit slower due to computing the update ratio for each FSR. I just pushed an update that precomputes some of the values for the update and removes the redundant computation of the update ratio for each moc energy group within a cmfd group. This update improves the performance so that the `c5g7-cmfd.py` with 3-nearest updating is within ~15% of the case when centroid update is turned off.

Also, as you noted, Cmfd will diverge when the centroid updating is turned off and the MOC relaxation factor is kept at 1.0. From my testing, the k-nearest updating has a window of convergence based on the number of neighbors used in the updating as shown in the picture below for C5G7 and the B&W cores. One option is to put in a divergence check in the Solver methods that use CMFD. If the residual begins increasing, we could stop the run and output an error noting that divergence has been encountered. Since Cmfd is unstable, we give the user the option to tweak the MOC relaxation factor to make it stable; however, with great power comes great responsibility as the user could easily tweak it to make the problem unstable. The hope with the k-nearest updating is that we could just get rid of the MOC relaxation factor all together or at least keep it hidden from users and only advise using it for testing purposes.

If you think we should add a divergence check or just remove the option to set the MOC relaxation factor, I'm happy to make the change.

bw-conv-comparison-2

wbinventor:
Ok this explanation makes a lot of sense now. Is it correct to say that the relaxation factor will not be applied if centroid updating is on? I do think that if a relaxation factor is set in the "c5g7-cmfd.py" file, it should be one that converges the event that one were to turn off centroid updating in that sample script, do you agree? 

Also, based on your plot it appears that C5G7 should converge in <15 iterations, but this was not the case for me when I ran it yesterday (23 iterations I think). Is your plot for a looser tolerance on convergence than the default?

Since as you say "great power comes with great responsibility," this stuff (eventually) needs to be documented well so we all know how to use it properly.

wbinventor:
Also, let me know once everything has been addressed, I'll independently test things out once more and merge it in.

samuelshaner:
The relaxation factor is still applied even if the centroid updating is on. They are set independent of each other. Yes, I agree it's probably best to have robust options in the input file in case users want to play around with the features a bit. I'll just push a commit with the relaxation factor set back to 0.6.

For the data on the plot, I was using the full 7 group CMFD, instead of the two group modification we have in the `c5g7-cmfd.py` input file. I was also using the non-energy-integrated total source in each FSR for convergence; however, at the time, the non-energy-integrated source was being divided by the total number of FSRs, not the total number of FSRs times the number of groups. With the current default settings on convergence and using 7 group CMFD, c5g7 converges in ~15 iterations.

I'm heading out to see my parents tomorrow, but I'll work on the documentation tonight and during the week.

wbinventor:
Ok, so it sounds like you are happy with this PR as it is?

samuelshaner:
I'm happy with it and give the go ahead to have it merged in.

samuelshaner:
I just updated the PR with one small modification. I moved the `computeResidual` in `Solver::computeEigenvalue` to after the CMFD update. It doesn't have a big effect on the convergence, but I feel its better to compute the residual with the flux after the CMFD update.

PR #146, opened 2015-05-07
Removed num_groups in calculation of fission source residual

samuelshaner:
The `_source_residuals` is currently divided by `_num_fissionable_FSRs * _num_groups`; however, the fission source is integrated over energy groups, so dividing by `_num_groups` is not necessary. This PR removes the `_num_groups` from the calculation of the `source_residual` in the CPUSolver, VectorizedSolver, and GPUSolver.

This PR builds on PR #137 and therefore should be reviewed after PR #137 is merged.


samuelshaner:
Issue was fixed in PR #142.

PR #147, opened 2015-05-12
Refactored exponentials and CMFD surface current tallying

wbinventor:
This PR is a merger of PR #137, #139, #140 and #140. I had hoped that the different PRs would make the reviews easier, but some of the functionality between them became mixed which led to some overlap. Hopefully this large PR will be "bug-free" now.


src/TrackGenerator.cpp
```diff
@@ -1608,3 +1551,82 @@ void TrackGenerator::correctFSRVolume(int fsr_id, FP_PRECISION fsr_volume) {
...
+
+        /* Split the segment into sub-segments */
+        for (int k=0; k < min_num_cuts; k++) {
+
+          /* Create a new Track segment */
+          segment* new_segment = new segment;
+          new_segment->_material = material;
+          new_segment->_length = length / FP_PRECISION(min_num_cuts);
+          new_segment->_region_id = fsr_id;
+          new_segment->_cmfd_surface_bwd = -1;
```
samuelshaner: With the segment constructor setting the the cmfd surface indices to -1 now, lines 1613 and 1614 are no longer necessary.

wbinventor: Good call, I just removed them. Yet another case of where the previous overlap between PRs led to some confusion and then this redundancy.

samuelshaner:
Need to add `boundary_type.h` to `.i` files.

wbinventor:
Sorry for that mistake, just added "boundary_type.h" to the swig interface files.

wbinventor:
I just sychronized all of the SWIG interface files (*e.g.*, "openmoc.i", "openmoc_intel_single.i", "openmoc_bgq_double.i", etc.) for the different compile time configurations. Seems that they differed significantly from different PRs that were merged in the past. I just ran a simulation with ``VectorizedSolver`` from ``openmoc.intel.single`` and it seems to be working fine.

You comments illustrate how badly we need a new build system for OpenMOC. I hear that @geogunow will have some time to learn and apply CMake for us this summer?

samuelshaner:
A build system and a test suite sure would be nice! Hopefully @geogunow can just tack on a new build system when he adds the irregular lattice input file and concurrent hash map.

I ran `VectorizedSolver` and it works great, but I get an error when I try to plot anything (when I have used the `VectorizedSolver` only):

  File "simple-lattice.py", line 161, in <module>
    plotter.plot_tracks(track_generator)
  File "/home/shaner/.local/lib/python2.7/site-packages/openmoc/plotter.py", line 81, in plot_tracks
    directory = get_output_directory() + subdirectory
NameError: global name 'get_output_directory' is not defined 

Any idea why this would be a problem only when `VectorizedSolver` is being used?

wbinventor:
I just fixed this issue - it was actually a more long-standing issue with the way that the ``process`` and ``plotter`` submodules imported ``openmoc`` when an Intel or BGQ compiled version is in use. In any case, I've confirmed that it works with ``openmoc.intel.single`` at least.

Yes, I like the idea of @geogunow adding in the concurrent hash map, and especially the test suite, those are great ideas.

openmoc/plotter.py
```diff
@@ -14,6 +14,8 @@
 #         openmoc.plotter module.
 openmoc = ''
 
+print('whoa!!!')
```
samuelshaner: While the plotter is pretty awesome, I'm not sure we need to print this when it is included.

wbinventor: Whoops just removed that.

Issue #63, opened 2014-03-14
Parallelize build system


wbinventor:
At the moment, the build system compiles each C++/CUDA source file sequentially. As a result, the build process can be quite slow. Although the use of ccache (i.e, with the "-with-ccache" flag to setup.py) speeds up compilation tremendously by only building those source files which have changed, it would be nice to speed it up even further with an option to compile source files in parallel.

To the user, this feature should simply come in the form of an additional compile time flag for the number of threads to use for parallel compilation.

A nice stack overflow exchange on this topic and the subtleties of implementing this capability for Linux/Mac and Windows is here:

http://stackoverflow.com/questions/11013851/speeding-up-build-process-with-distutils

PR #148, opened 2015-05-14
Parallel build system

wbinventor:
This PR makes the build system compile objects in parallel using a Python ``ThreaPool`` from the built-in ``multiprocessing.pool`` module. I came across this simple change on StackOverflow the other day and decided to add it to our build system since it is a nice and easy addition. It seems to work quite well and in particular, greatly speeds up the compilation of the ``openmoc.cuda`` module's CUDA source files since nvcc cannot leverage [ccache](https://ccache.samba.org/) in the way that C/C++ compilers can.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,8 +2,11 @@
 The setup script for OpenMOC
 '''
 
-from distutils.errors import DistutilsOptionError
 import os, string
+from distutils.errors import DistutilsOptionError
+import distutils.ccompiler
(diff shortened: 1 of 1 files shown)


wbinventor:
This address Issue #63.

samuelshaner:
@wbinventor - this is great! Looks good and works for me so I'll merge it right in.

PR #149, opened 2015-05-21
Simplified build system

wbinventor:
This PR greatly simplifies the distutils-based build system. In particular, the ability to compile and install different OpenMOC installations (*.e.g.*, ``openmoc.intel.single``, ``openmoc.gnu.double``, etc.) is eliminated since it was mainly useful for my M.S. thesis and not much else. In addition several changes were made to the "setup.py" and "config.py" files to make them more intuitive for developers to navigate and extend. The overall effect of these changes is to largely gut the build system to make it simpler and more effective. This PR is independent of all other PR's currently in the queue and can be reviewed on its own, though it will nicely complement #148.


wbinventor:
The new build system also supports Apple's C/C++ [Clang](http://en.wikipedia.org/wiki/Clang) compiler, version 3.5 and up.

samuelshaner:
Have you tried compiling with clang? I've never used clang, but when trying it out on my mac I was getting some linking errors.

wbinventor:
First, I think that I need to resolve some merge conflicts for this PR from
the other PRs that were merged in last week. I made it halfway through this
process before I was preoccupied for the weekend, but will wrap it up later
today.

As for clang, I do think you must use clang version 3.5 or later. I can't
recall the issue, but I know that earlier versions don't support some of
the C++11 features used in the code and t failed at link time for me too.

config.py
```diff
@@ -210,13 +217,15 @@ class configuration:
...
-  if (get_platform()[:6] == 'macosx'):
+  if ('macosx' in get_platform()):
     linker_flags['gcc'] = ['-fopenmp', '-dynamiclib', '-lpython2.7',
                            '-Wl,-install_name,' + get_openmoc_object_name()]
   else:
     linker_flags['gcc'] = ['-fopenmp', '-shared',
                            '-Wl,-soname,' + get_openmoc_object_name()]
 
+  linker_flags['clang'] = ['-fopenmp', '-shared',
+                         '-Wl,-soname,' + get_openmoc_object_name()]
```
samuelshaner: The `-soname` and `-shared` linker flags for clang should be changed to '-install_name' and `-dynamiclib` for clang on mac OS X just as with gcc.

samuelshaner:
Fixed the mac clang linking errors. Just needed to add a path to the OpenMP include files and libraries. Installs just fine with clang on a mac.

wbinventor:
I just pushed a commit which fixes the current build failure for the CUDA module on the "develop" branch. This was caused by the introduction of the helper ``second`` template method in "Universe.h" in PR #150 which caused a namespace collision with a method of the same name in the CUDA Thrust library. For some reason, this was not caught during testing of that PR, even though I am confident that I tried it out with CUDA. Anyway, I renamed ``second`` to ``pair_second`` so that there is no longer a collision.

config.py
wbinventor: Ok thanks for letting me know. I just pushed those changes. Does the Mac version of clang also need ``-lpython2.7``?

wbinventor:
I should add to my previous comment that PR #150 for neighbor cells introduced the use of some C++11 features that are only supported in NVIDIA CUDA 6.5 and later.

config.py
samuelshaner: yes is does.

samuelshaner:
Great to hear you got the build failure! I'll gonna test it first thing in the morning on a GPU and then merge it in if all looks good.

config.py
wbinventor: Alright, just added it in.

wbinventor:
@samuelshaner - I've addressed your comments about building with clang on a Mac. Let me know if it works now.

samuelshaner:
Just tested it and it works great with clang on a mac.

wbinventor:
Great, just let me know if you see anything else that you want me to revise.

PR #150, opened 2015-05-22
Neighbor cell optimizations for ray tracing

wbinventor:
This PR introduces a number of optimizations to the ray tracing process in OpenMOC. The most notable optimization and new feature is the construction of "neighbor cell" lists for each ``Cell`` and ``Surface``. This is a similar ray tracing optimization to what is currently used in [OpenMC](https://mit-crpg.github.io/openmc/). However, this implementation is a bit more forgiving, albeit at the expense of a little speed (in particular, it doesn't have the same [constraints](https://github.com/mit-crpg/openmc/issues/260) on the re-use of ``Surfaces`` in different ``Cells`` in different ``Universes`` as in OpenMC). In addition, the ray tracing procedures were simplified quite a bit in the process.

The optimizations in this PR are particularly useful for problems with a lot of unstructured (non-Cartesian) spatial discretization - *e.g.*, geometries with many``Cells`` per ``Universe``. This code makes the current O(n) lookup for the "next" ``Cell`` while tracing a ray across a ``Universe`` much closer to O(1) lookup amongst the few (hopefully << n) neighbors of the previous ``Cell``. Of course this speedup will vary greatly for different geometries. But it should mean that the time spent ray tracing will vary much less during a spatial discretization study.

From a few quick tests on my laptop, the ray tracing runtime for the pure vanilla "c5gy.py" benchmark decreased from 115 second to 85 seconds, a ~25% performance improvement over the current "develop" branch. The speedup increases with more rings/sector discretizations per pin cell. For example, when using 5 rings / 16 sectors per pin cell (an increase of 4x over the standard "c5g7.py"), the ray tracing runtime decreased from 670 seconds to 87 second (7x speedup). This illustrates how little the runtime changes for the new implementation with pin cell discretization.

These changes make analyses of larger geometries - such as highly discretized assemblies or full-core [BEAVRS](http://crpg.mit.edu/pub/beavrs) - much more practical.

### DIFF
--- a/src/Cell.cpp
+++ b/src/Cell.cpp
@@ -542,7 +542,7 @@ void Cell::findBoundingBox() {
  *          the Cell if it is on the same side of every Surface in the Cell.
  * @param point a pointer to a Point
  */
-bool Cell::cellContainsPoint(Point* point) {
+bool Cell::containsPoint(Point* point) {
 
   /* Loop over all Surfaces inside the Cell */
   std::map<int, surface_halfspace>::iterator iter;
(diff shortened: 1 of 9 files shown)


samuelshaner:
@wbinventor - another great PR! Could not find anything that I didn't agree with and everything looked great in testing. Excited to see the performance gain for some big problems!

Issue #152, opened 2015-06-11
Floor index in GPUExpEvaluator computeExponential function


samuelshaner:
In a recent PR, it was noted that the index calculation in the computeExponential function in the ExpEvaluator should be floored instead of rounded. This change should be reflected in the GPUExpEvaluator as well.

wbinventor:
Good call, I definitely thought I made that change in both places but looks like I did not. See #153 for the fix.

PR #153, opened 2015-06-12
Now using floor instead of integer rounding for GPUExpEvaluator

wbinventor:


wbinventor:
This is to address Issue #152.

PR #154, opened 2015-06-13
Fixed errors caused by improper logger initialization

wbinventor:
This PR fixes an issue I noticed when accidentally trying to ``materialize`` a material data file that was non-existent (*i.e.*, the path to it was incorrect). In this case, instead of reporting an error message, the code simply segfault'ed. After tracing this error down, I found that was *not* an issue with the ``materialize`` module but rather a bug introduced by PR #140 with improper initialization of the logger module's OpenMP ``log_error_lock`` static variable. Fortunately the code changes necessary to fix this are minimal.

Previously, ``log_error_lock`` was initialized in the ``set_output_directory(...)`` routine which was *only* called by the user and not internally by OpenMOC. If the user never called ``set_output_directory(...)``, the logger module would instead create a "log" directory to store all of its log files. The creation of the "log" directory was explicitly done in the ``log_printf(...)`` routine rather than by correctly calling the ``set_output_directory(...)`` routine. 

I modifed ``log_printf(...)`` to call ``set_output_directory(...)`` to initialize the output directory. In addition, I added a new ``initialize_logger(...)`` routine which explicitly initializes the ``log_error_lock`` OMP variable. The ``initialize_logger(...)`` is called when OpenMOC is imported into Python (as can now be see in the "openmoc/\_\_init\_\_.py" file) to ensure that the lock is always ready to go if and when an "ERROR" message is channeled through the logger.

It should be noted that this bug occurs only when compiling with Intel's rather than GNU's C++ compiler. The reason (I think) is that ``g++`` implicitly calls ``omp_init_lock(...)`` for all variables of data type ``omp_lock_t`` while ``icpc`` requires explicit initialization in the source code. But this bug was never an issue when compiled with Intel *if* there was not an "ERROR" message to report. Oh so subtle...

This PR also slightly improves the ``try-catch`` exception handling for the ``openmoc.materialize`` module.

### DIFF
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -102,7 +102,7 @@ void set_output_directory(char* directory) {
 
   /* Check to see if directory exists - if not, create it */
   struct stat st;
-  if ((stat(directory, &st)) == 0)
+  if ((stat(directory, &st)) == 0) {
     log_directory = std::string("") + directory + std::string("/log");
     mkdir(log_directory.c_str(), S_IRWXU);
   }
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -77,6 +77,20 @@ static omp_lock_t log_error_lock;
 
 
 /**
+ * @brief Initializes the logger for use.
+ * @details This should be immediately called when the logger is imported
+ *          into Python and before any of its other routines are called. The
+ *          routine initializes an OpenMP mutual exclusion lock which is used 
+ *          to preclude race conditions from occurring when an ERROR message
+ *          is reported and program execution is terminated.
+ */
+void initialize_logger() {
+  /* Initialize OpenMP mutex lock for ERROR messages with exceptions */
+  omp_init_lock(&log_error_lock);
+}
+
+
+/**
  * @brief Sets the output directory for log files.
  * @details If the directory does not exist, it creates it for the user.
  * @param directory a character array for the log file directory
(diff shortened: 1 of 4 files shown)


wbinventor:
@PrezNattyGibbs, @geogunow care to take a look at this so we can clear out our queue of PRs? This one is pretty short and sweet.

PrezNattyGibbs:
This looks straightforward and your description makes sense.  This is a little outside my wheelhouse of understanding exactly what's going on, but I'm happy if you're happy.

geogunow:
Looks simple. I read through the PR thoroughly (it's short) and agree with everything. I think this is ready to merge.

samuelshaner:
With these changes, the `plots` and `tracks` directories are now subdirectories of the `log` directory. @wbinventor - any reason why you made this change? Or was this unintentional?

Also, when I tested this on my mac and on NSECluster, no log file was generated. Are you seeing this behavior?

wbinventor:
@samuelshaner thanks for trying this out and exposing that issue. I can't believe I missed the fact that the log files were not being created. I think I've fixed this now.

samuelshaner:
@wbinventor can you push your changes?

wbinventor:
Sorry for the delay, here it is.

src/log.cpp
```diff
@@ -84,18 +98,14 @@ static omp_lock_t log_error_lock;
...
 
   output_directory = std::string(directory);
+  std::string log_directory;
 
   /* Check to see if directory exists - if not, create it */
   struct stat st;
-  if ((!stat(directory, &st)) == 0) {
-    mkdir(directory, S_IRWXU);
-    mkdir((output_directory+"/log").c_str(), S_IRWXU);
+  if ((stat(directory, &st)) == 0)
```
samuelshaner: You are missing a `{` at the end of this line.

wbinventor: Done

PR #151, opened 2015-06-11
Quick fix for Intel compiler

wbinventor:
This PR fixes an issue that was introduced in the recent spate of PRs with the Intel compiler. I'm honestly unsure how this slipped through the cracks since I did extensive testing with each PR individually, but this bug still slipped by even after all of them were merged. I noticed this bug since OpenMOC will only compile with icpc on nsecluster (the version of gcc is *very* old and does not support some necessary C++11 features).

PR #157, opened 2015-06-17
Parallel hash map

geogunow:
This pull request fixes segmentation faults due to accessing std::unordered_map objects by adding a parallel_hash_map class which provides thread-safe guarantees and incorporates fine-grained locking.

The current implementation of OpenMOC uses a std::unordered_map to store FSR data during segmentation in the form of key/value pairs where the key is a unique string identifier referring to the FSR's position within the geometrical hierarchy. The map is locked during inserts but simply read lock-free during lookup calls. While this might seem to be thread-safe behavior, issues are encountered when the hash map is queried during a resize. Specifically, when the underlying map is deleted during a resize and another thread attempts to read the map a segmentation fault is encountered.

The new implementation uses a parallel hash map object whose source code is contained within parallel_hash_map.h. During inserts, the new parallel hash map uses fine grained locking across buckets rather than a global lock over the entire hash map. The fine grained locking across buckets ensures the map can be accessed in a thread-safe manner during resize operations.

The performance is very similar to the previous implementation although the scaling does seem slightly better as seen in the following plots:

speedup
segmentation_time

However, the true advantage of the parallel hash map is that it is thread safe and did not encounter any failures in segmentation during the test trials:

failure_rate

Other small changes were made to the code to ensure consistency and appropriate code structure. @samuelshaner, @wbinventor, and @PrezNattyGibbs have provided feedback on the code structure. Further work might be pursued so that OpenMOC is more conducive to multi-threading. Feedback on the revised code structure is encouraged.


src/TrackGenerator.cpp
```diff
@@ -1,5 +1,5 @@
 #include "TrackGenerator.h"
-
+#include<stdlib.h>
```
geogunow: This can be removed

wbinventor: I'm just curious, if it can be removed, why are you commenting on it rather
than removing it?

geogunow: This comment is outdated. The include has been removed in the current PR.

src/Geometry.cpp
```diff
@@ -33,9 +28,12 @@ Geometry::Geometry() {
  * @brief Destructor clears FSR to Cells and Materials maps.
  */
 Geometry::~Geometry() {
-
   /* Free FSR  maps if they were initialized */
-  if (_num_FSRs != 0) {
+  if (_FSR_keys_map.size() != 0) {
+    fsr_data **values = _FSR_keys_map.values();
```
wbinventor: a newline around the ``values`` block would be nice for the reader

src/Geometry.cpp
```diff
@@ -528,48 +516,49 @@ int Geometry::findFSRId(LocalCoords* coords) {
...
-    /* Release lock */
-    omp_unset_lock(_num_FSRs_lock);
-
   }
+  
   /* If FSR has already been encountered, get the fsr id from map */
-  else
-    fsr_id = _FSR_keys_map.at(fsr_key_hash)._fsr_id;
+  else{
+    fsr_data volatile* fsr;
```
wbinventor: @geogunow - what do you think about encapsulating this ``do-while`` loop as a more generic routine in the ``ParallelHashMap``? Could it not be contained within the ``ParallelHashMap::at(...)`` routine? If you don't like this that's fine, just seems like it might be cleaner as well as make your ``ParallelHashMap`` even more powerful for other applications such that the ``do-while`` loop is in one place only.

src/Geometry.cpp
```diff
@@ -835,8 +824,51 @@ void Geometry::segmentize(Track* track) {
...
   start.prune();
   end.prune();
+}
+
+
+/**
+ * @brief Initialize key and material ID vectors for lookup by FSR ID
+ * @detail This function initializes and sets reverse lookup vectors by FSR ID.
+ *      This is called after the FSRs have all been identified and allocated
+ *      during segmentation.
```
wbinventor: can you just add a clarifying comment here in the ``@detail`` doxygen comment block that specifies which routine(s) must be called before this one, *i.e.*, ``TrackGenerator::generateTracks()`` or something even more specific

src/ParallelHashMap.h
```diff
@@ -0,0 +1,838 @@
...
+/**
+ * @class fixed_hash_map ParallelHashMap.h "src/ParallelHashMap.h"
+ * @brief A fixed-size hash map supporting insertion and lookup operations
+ * @details The fixed_hash_map class supports insertion and lookup operations
+ *    but not deletion as deletion is not needed in the OpenMOC application. 
+ *    This hash table uses chaining for collisions and does not incorporate
+ *    concurrency objects except for tracking the number of entries in the
+ *    table for which an atomic increment is used. This hash table is not
+ *    thread safe but is used as a building block for the ParallelHashMap
+ *    class. This table guarantees O(1) insertions and lookups on avarge.
```
wbinventor: typo: "avarge" -> "average"

src/ParallelHashMap.h
```diff
@@ -0,0 +1,838 @@
...
+    K* keys();
+    V* values();
+    void clear();
+    void print_buckets();
+};
+
+/**
+ * @class ParallelHashMap ParallelHashMap.h "src/ParallelHashMap.h"
+ * @brief A thread-safe hash map supporting insertion and lookup operations
+ * @details The ParallelHashMap class is built ontop of the fixed_hash_map
```
wbinventor: typo: "ontop" -> "on top"

src/Geometry.h
```diff
@@ -135,17 +136,16 @@ class Geometry {
...
 
   /* Set parameters */
   void setFSRsToMaterialIDs(std::vector<int> FSRs_to_material_IDs);
   void setFSRsToKeys(std::vector<std::size_t> FSRs_to_keys);
-  void setNumFSRs(int num_fsrs);
   void setCmfd(Cmfd* cmfd);
 
 #ifndef CUDA
-  void setFSRKeysMap(std::unordered_map<std::size_t, fsr_data> FSR_keys_map);
+  void setFSRKeysMap(ParallelHashMap<std::size_t, fsr_data*>* FSR_keys_map);
```
wbinventor: Per our convo on the closed source side about exposing this to the user, how about we add it as a SWIG ``%ignore`` to the SWIG interface file? See an example of the ``%ignore`` in use in @samuelshaner's PR #144 [here](https://github.com/mit-crpg/OpenMOC/pull/144/files#diff-d2dacb1ed6c26bddff18021b37dd9c1eR56)

wbinventor: And if you see any other routines here that are public out of necessity but which should not be exposed to the user in Python, go ahead and ``%ignore`` them as well (including perhaps ``Geometry::getFSRKeysMap()``).

src/ParallelHashMap.h
```diff
@@ -0,0 +1,838 @@
...
+ * @param key key of the key/value pair to be inserted
+ * @param value value of the key/value pair to be inserted
+ */
+template <class K, class V>
+void fixed_hash_map<K,V>::insert(K key, V value)
+{
+  /* get hash into table using fast modulus */
+  size_t key_hash = std::hash<K>()(key) & (_M-1);
+
+  /* check to see if key already exisits in map */
```
wbinventor: typo: "exisits" -> "exists"

src/ParallelHashMap.h
```diff
@@ -0,0 +1,838 @@
...
+  {
+    if(_buckets[i] == NULL)
+      std::cout << i << " -> NULL" << std::endl;
+    else
+      std::cout << i << " -> " << _buckets[i] << std::endl;
+  }
+}
+
+/**
+ * @brief Constructor for generates initial underlying table as a fixed-sized
```
wbinventor: grammar or typo issue - possible remove "for" here

src/ParallelHashMap.h
```diff
@@ -0,0 +1,838 @@
...
+ * @brief Constructor initializes fixed-size table of buckets filled with empty
+ *      linked lists.
+ * @details The constructor initializes a fixed-size hash map with the size
+ *      as an input parameter. If no size is given the default size (64)
+ *      is used. Buckets are filled with empty linked lists presented as
+ *      NULL pointers.
+ * @param M size of fixed hash map
+ */
+template <class K, class V>
+fixed_hash_map<K,V>::fixed_hash_map(size_t M)
```
wbinventor: just curious, why are you using CamelCase for ``ParallelHashMap`` but underscores (Python PEP8 sytax) for the ``fixed_hash_map`` class?

wbinventor: Of course as I mentioned before, this is your data structure which is encapsulated and usable outside of OpenMOC, so if you want to use a coding style more similar to C++ STL than OpenMOC that is fine.

src/ParallelHashMap.h
```diff
@@ -0,0 +1,838 @@
...
+{
+  return _num_locks;
+}
+
+/**
+ * @brief Returns an array of the keys in the underlying table
+ * @details All buckets are scanned in order to form a list of all keys
+ *      present in the table and then the list is returned. Threads
+ *      announce their presence to ensure table memory is not freed
+ *      during access.
```
wbinventor: Add a NOTE or WARNING here in the ``@details`` comment block that it is the caller's responsibility to free the memory allocated for the returned ``keys`` array.

src/ParallelHashMap.h
```diff
@@ -0,0 +1,838 @@
...
+
+  return key_list;
+}
+
+/**
+ * @brief Returns an array of the values in the underlying table
+ * @details All buckets are scanned in order to form a list of all values
+ *      present in the table and then the list is returned. Threads
+ *      announce their presence to ensure table memory is not freed
+ *      during access.
```
wbinventor: Ditto - add a NOTE or WARNING here in the ``@details`` comment block that it is the caller's responsibility to free the memory allocated for the returned ``values`` array.

src/TrackGenerator.cpp
```diff
@@ -1447,6 +1453,7 @@ bool TrackGenerator::readTracksFromFile() {
   _geometry->setFSRKeysMap(FSR_keys_map);
   _geometry->setFSRsToMaterialIDs(FSRs_to_material_IDs);
   _geometry->setFSRsToKeys(FSRs_to_keys);
+
```
wbinventor: delete unnecessary whitespace

src/Geometry.h
```diff
@@ -73,12 +77,9 @@ class Geometry {
    *  the Geometry. False is for vacuum and true is for reflective BCs. */
   boundaryType _right_bc;
 
-  /** The total number of FSRs in the Geometry */
-  int _num_FSRs;
-
   /** An map of FSR key hashes to unique fsr_data structs */
 #ifndef CUDA
-  std::unordered_map<std::size_t, fsr_data> _FSR_keys_map;
+  ParallelHashMap<std::size_t, fsr_data*> _FSR_keys_map;
```
wbinventor: We *should* now be able to remove the ``#ifndef CUDA`` around this code block.

wbinventor:
@geogunow - I've looked over this PR thoroughly for a second time and it looks very, very good! I only made a few minor comments, nothing that should take long to fix. Pending what @samuelshaner and @PrezNattyGibbs think, we should be able to merge this in this week. This is a **major** improvement over the status quo and I am glad that we can now safely ray trace in parallel, should make a lot of our calculations more practical in the near future.

I gave your code some test runs on my laptop as well as nsecluster (C5G7, 128 azim angles, 0.05 spacing). On my laptop with an i5 processor (4 cores) I see a ~40% speedup in ray tracing, and on nsecluster with 12 core Xeons a 6.7x speedup. The speedup on the cluster is dramatically better than the performance improvement in your plots - any idea why? I will say I compiled on the cluster with the Intel compiler, perhaps you should give that a try.

As a side note, I am still a bit curious just what the bottleneck is that precludes better performance. I took a screen capture of the Ubuntu System Monitor when running with 4 threads/cores on my laptop:

ray-tracing-perf-profile

As you can see, the performance profile changes dramatically between ray tracing and MOC transport sweeps. Although this is very hand wavy performance profiling (you guys at CESAR are the experts here, and this is with a heavyweight OS doing lots of other funny things), it goes without question that something is going on during ray tracing which kills performance. I don't think this is an issue for this PR, but since we are talking about speeding up ray tracing, thought I'd post this up here for a source of thought and discussion for future work.

wbinventor:
You should note that you will need to synchronize this with the latest version of the "develop" branch in order to compile with Intel per PR #151

src/Geometry.cpp
geogunow: I see what you're saying but I'm a bit hesitant to put it in ParallelHashMap because the ``do-while`` loop arises due to an application-specific quirk. And one that we should probably get rid of - namely that we try to add in a value which is an ``fsr_data`` object containing an attribute ``fsr_id`` which refers to the order in which the element was added. I would like to find a way around doing that altogether. But for now we need to do this synchronization in order to make that strange operation thread safe. I think once this is merged in, I will look at further restructuring the code so that we don't need insert_and_get_count and we don't need the ``do-while`` block when checking the ``fsr_id`` but for now I think the code should stay in its current form.

src/ParallelHashMap.h
geogunow: Agreed, ``fixed_hash_map`` changed to ``FixedHashMap``

geogunow: I am keeping a generic version of the hash table open source on my github. This hash table I want to conform to OpenMOC style.

geogunow:
It should be synced with develop now

geogunow:
As to the performance, my plots showed **only the speedup during ray tracing**. As to why it's not performing optimally, I'm not sure. But speeding up ray tracing with parallel computing is a far harder problem than during transport sweeps as the synchronization during transport sweeps is trivial and practically non-existent. Most of the calculation during transport sweeps is performed on local variables so not many parallel computing challenges need to be addressed. For ray tracing, I think there are ways to improve the was we currently do things, but let's save that for another PR. But if you are interested, I view these as the theoretical "sequential bottlenecks" of the implementation in the PR.

- In ``ParallelHashMap`` we synchronize on the number of elements in the map during inserts meaning
```
_N++;
```
for insert or
```
count = _N++;
```
in the case of ``insert_and_get_count`` must be done atomically. I don't think this is a huge deal but it is worth noting that ``insert`` should be faster since less synchronization is required. 

- In ``ParallelHashMap`` we lock buckets during inserts. This means that the entire insert operation locks out any threads trying to add a value to the same bucket.

- In ``Geometry::segmentize()`` we need to run a ``do-while`` loop to get the ``fsr_id``. This is because when we insert a pointer to an ``fsr_data`` object into the map, we first insert a null pointer. Then after the insert was successful, we change the null pointer to point to an actual ``fsr_data`` object. 

- In ``Geometry::initializeFSRVectors`` we allocate arrays sequentially. Then we fill the arrays and add material IDs to cells in parallel. Then we **sequentially** add CMFD cells to the CMFD mesh due to the current CMFD structure. I didn't want to touch the CMFD structure at all in this PR but the structure should probably be updated to be more flexible to multicore programming.

Other multicore issues can effect performance besides theoretical sequential bottlenecks. For instance, on my first crack at implementing the ``ParallelHashMap`` the performance was absolutely awful due to false sharing. In addition, thread contention and saturating the inter-core bus (lots of threads sending messages) can have a large impact on performance. It would be nice to profile the ray tracing part of the code to pinpoint the bottleneck but that is hard to do with the python wrapper. That is why I am currently working on a PR to have an alternative C++ build system.

wbinventor:
First, the speedups I reported (40% and 6.7x) were for ray tracing only. I'm still curious why I saw 6.7x on 12 cores vs. your 3.5x on 32 cores (I have a hard time believing it was simple the secret sauce in the Intel compiler, but perhaps so).

Second, I completely agree with you that a C++ build system is what is needed to systematically profile the ray tracing code (and everything else) to isolate the hot spots.

That being said, I'm skeptical that significantly better multi-core scaling might be achieved simply with an even better ``ParallelHashMap``. The reason for my skepticism - as was discussed for this PR on ClosedMOC - is that there was no need for a ``ParallelHashMap`` during ray tracing prior to #111 due to our use of a deterministic method to identify FSRs prior to ray tracing. This "distribcells" algorithm was removed and replaced with the present method by @samuelshaner in order to accommodate geometries with overlapping cells (*e.g.*, a core barrel cell which overlaps the fuel assembly lattice in a full-core model). Using the "distribcells" method, ray tracing was trivially parallelized over ``Tracks`` but without a very good speedup - perhaps 5-7x on 12 cores, but I don't remember exactly since we never focused on this before. 

In any case, the scaling sucked even with a trivially threadsafe ray tracing algorithm. Therefore, my guess would is the data structures used to represent and store geometric primitives may need to be restructured to unlock better performance, but of course that hypothesis can only be confirmed or rejected with some performance profiling.

wbinventor:
@PrezNattyGibbs, @samuelshaner, have you two looked this over, tested it out, and feel comfortable with it?

openmoc/openmoc.i
```diff
@@ -53,6 +53,10 @@
 %warnfilter(506) log_printf(logLevel level, const char *format, ...);
 %warnfilter(511) swig::SwigPyIterator;
 %warnfilter(511) Cell::setFill;
+%ignore setFSRsToMaterialIDs(std::vector<int> FSRs_to_material_IDs);
```
samuelshaner: Can you update lines 56-58 to reflect the recent changes in PR #144? I'm not sure why those are not showing up as a merge conflict in this PR.

src/Geometry.cpp
```diff
@@ -33,9 +28,15 @@ Geometry::Geometry() {
...
  */
 Geometry::~Geometry() {
-
+  
   /* Free FSR  maps if they were initialized */
-  if (_num_FSRs != 0) {
+  if (_FSR_keys_map.size() != 0) {
+    fsr_data **values = _FSR_keys_map.values();
+
+    for(int i=0; i<_FSR_keys_map.size(); i++)
```
samuelshaner: Per the Google C++ Style Guide, there should be a space between `for` and `(`.

src/Geometry.cpp
```diff
@@ -33,9 +28,15 @@ Geometry::Geometry() {
  * @brief Destructor clears FSR to Cells and Materials maps.
  */
 Geometry::~Geometry() {
-
+  
   /* Free FSR  maps if they were initialized */
```
samuelshaner: Typo - two spaces between `FSR` and `maps`.

src/Geometry.cpp
```diff
@@ -528,48 +519,49 @@ int Geometry::findFSRId(LocalCoords* coords) {
...
-
-    /* Recheck to see if FSR has been added to maps after getting the lock */
-    if (_FSR_keys_map.find(fsr_key_hash) != _FSR_keys_map.end())
-      fsr_id = _FSR_keys_map.at(fsr_key_hash)._fsr_id;
+  if (!_FSR_keys_map.contains(fsr_key_hash)){
+
+    /* Try to get a clean copy of the fsr_id, adding the FSR data 
+       if necessary where -1 indicates the key was already added */
+    fsr_id = _FSR_keys_map.insert_and_get_count(fsr_key_hash, NULL);
+    if( fsr_id == -1)
```
samuelshaner: ditto - space between `if` and `(`.

src/Geometry.cpp
```diff
@@ -528,48 +519,49 @@ int Geometry::findFSRId(LocalCoords* coords) {
...
+    {
+      fsr_data volatile* fsr;
+      do{
```
samuelshaner: space missing between `do` and `{`.

src/Geometry.cpp
```diff
@@ -528,48 +519,49 @@ int Geometry::findFSRId(LocalCoords* coords) {
...
+        fsr = _FSR_keys_map.at(fsr_key_hash);
+      } while(fsr == NULL);
```
samuelshaner: space missing.

geogunow: Looking at it now. Is there a reason you changed from vectors to pointers
to vectors for the FSRsToKeys, etc ?

samuelshaner: For the same reason we change the methods that pass maps to now pass pointers to maps (i.e. to avoid deep copies).

src/Geometry.cpp
```diff
@@ -528,48 +519,49 @@ int Geometry::findFSRId(LocalCoords* coords) {
...
   }
+  
   /* If FSR has already been encountered, get the fsr id from map */
-  else
-    fsr_id = _FSR_keys_map.at(fsr_key_hash)._fsr_id;
+  else{
+    fsr_data volatile* fsr;
+    do{
+      fsr = _FSR_keys_map.at(fsr_key_hash);
+    } while(fsr == NULL);
```
samuelshaner: ditto - space between `while` and `(`. I'll stop pointing out these style guide issues now.

src/Geometry.cpp
```diff
@@ -835,8 +827,53 @@ void Geometry::segmentize(Track* track) {
...
+ *      are loaded from a file.
+ */
+void Geometry::initializeFSRVectors(){
+  
+  /* get keys and values from map */
+  std::size_t *key_list = _FSR_keys_map.keys();
+  fsr_data **value_list = _FSR_keys_map.values();
+
+  /* allocate vectors */
+  int N = _FSR_keys_map.size();
```
samuelshaner: Could `N` be changed to `num_FSRs` to be more descriptive?

openmoc/openmoc.i
```diff
@@ -59,6 +59,12 @@
 %ignore setFSRKeysMap(std::unordered_map<std::size_t, fsr_data>* FSR_keys_map);
 %ignore setFSRsToKeys(std::vector<std::size_t>* FSRs_to_keys);
 %ignore setFSRsToMaterialIDs(std::vector<int>* FSRs_to_material_IDs);
+%ignore setFSRsToMaterialIDs(std::vector<int> FSRs_to_material_IDs);
```
samuelshaner: Lines 62 and 63 are no longer needed.

src/Geometry.cpp
geogunow: True, but we don't really follow that rule in OpenMOC

PrezNattyGibbs: I think we're putting spaces before `{` in the vast majority of cases.  Where are you seeing us not doing it?

wbinventor: Agreed, I *think* this overwhelmingly the case. And if not, it should be in
the future since it makes for cleaner code.

geogunow: Yes, there are inconsistencies all over OpenMOC with regard to the space before ``{``.

If I type
```
grep -r \){ src
```
from the ``OpenMOC`` directory, it returns:
```
src/accel/cuda/GPUSolver.cu:  for (int e=0; e < _num_groups; e++){
src/accel/cuda/GPUSolver.cu:  for (int g=0; g < _num_groups; g++){
src/accel/cuda/GPUSolver.cu:void GPUSolver::initializeExpEvaluator(){
src/accel/cuda/GPUSolver.cu:    for (iter=host_materials.begin(); iter != host_materials.end(); ++iter){
src/log.cpp:int get_log_level(){
src/linalg.h:inline void vector_copy(T* vector_from, T* vector_to, int length){
src/linalg.h:inline void matrix_zero(T** matrix, int width, int length){
src/linalg.h:  for (int i = 0; i < length; i++){
src/linalg.h:inline void vector_zero(T* vector, int length){
src/linalg.h:                                  int block_width){
src/linalg.h:  for (int i = 0; i < num_blocks; i++){
src/linalg.h:    for (int g = 0; g < block_width; g++){
src/linalg.h:      for (int e = 0; e < block_width; e++){
src/linalg.h:inline void vector_scale(T* vector, T scale_value, int length){
src/ExpEvaluator.cpp:  for (int i=0; i < num_array_values; i++){
src/ExpEvaluator.cpp:    for (int p=0; p < num_polar; p++){
src/Geometry.cpp.orig:Cmfd* Geometry::getCmfd(){
src/Geometry.cpp.orig:void Geometry::setCmfd(Cmfd* cmfd){
src/Geometry.cpp.orig:  if (univ->getId() == _root_universe->getId()){
src/Geometry.cpp.orig:    if (_cmfd != NULL){
src/Geometry.cpp.orig:  if (!_FSR_keys_map.contains(fsr_key_hash)){
src/Geometry.cpp.orig:      if (_cmfd != NULL){
src/Geometry.cpp.orig:  if (_cmfd != NULL){
src/Geometry.cpp.orig:  while(curr != NULL){
src/Geometry.cpp.orig:    if (_cmfd != NULL){
src/Geometry.cpp.orig:void Geometry::initializeFSRVectors(){
src/Geometry.cpp.orig:void Geometry::initializeCmfd(){
src/Geometry.cpp.orig:ParallelHashMap<std::size_t, fsr_data*>* Geometry::getFSRKeysMap(){
src/Geometry.cpp.orig:std::unordered_map<std::size_t, fsr_data>* Geometry::getFSRKeysMap(){
src/Geometry.cpp.orig:std::vector<std::size_t>* Geometry::getFSRsToKeys(){
src/Geometry.cpp.orig:                             FSR_keys_map){
src/Geometry.cpp.orig:void Geometry::setFSRsToKeys(std::vector<std::size_t>* FSRs_to_keys){
src/Geometry.cpp.orig:void Geometry::setFSRsToMaterialIDs(std::vector<int>* FSRs_to_material_IDs){
src/Geometry.cpp.orig:bool Geometry::withinBounds(LocalCoords* coords){
src/Geometry.cpp.orig:void Geometry::setFSRCentroid(int fsr, Point* centroid){
src/Geometry.cpp.orig:Cell* Geometry::findCellContainingFSR(int fsr_id){
src/Surface.cpp:Surface::Surface(const int id, const char* name){
src/Surface.cpp:boundaryType Surface::getBoundaryType(){
src/Surface.cpp:double Plane::getMinX(int halfspace){
src/Surface.cpp:double Plane::getMaxX(int halfspace){
src/Surface.cpp:double Plane::getMinY(int halfspace){
src/Surface.cpp:double Plane::getMaxY(int halfspace){
src/Surface.cpp:double Plane::getMinZ(int halfspace){
src/Surface.cpp:double Plane::getMaxZ(int halfspace){
src/Surface.cpp:double XPlane::getMinX(int halfspace){
src/Surface.cpp:double XPlane::getMaxX(int halfspace){
src/Surface.cpp:double YPlane::getMinY(int halfspace){
src/Surface.cpp:double YPlane::getMaxY(int halfspace){
src/Surface.cpp:double ZPlane::getMinZ(int halfspace){
src/Surface.cpp:double ZPlane::getMaxZ(int halfspace){
src/Surface.cpp:double Circle::getMinX(int halfspace){
src/Surface.cpp:double Circle::getMaxX(int halfspace){
src/Surface.cpp:double Circle::getMinY(int halfspace){
src/Surface.cpp:double Circle::getMaxY(int halfspace){
src/Surface.cpp:double Circle::getMinZ(int halfspace){
src/Surface.cpp:double Circle::getMaxZ(int halfspace){
src/ParallelHashMap.h:    node(K k_in, V v_in) : next(NULL), key(k_in), value(v_in){}
src/CPUSolver.cpp:                                FP_PRECISION* fsr_flux){
src/CPUSolver.cpp:    for (int p=0; p < _num_polar; p++){
src/CPUSolver.cpp:                                    FP_PRECISION* track_flux, bool fwd){
src/LocalCoords.cpp:    else if (curr->getType() == LAT){
src/Solver.cpp:void Solver::initializeCmfd(){
src/Solver.cpp:    if (_cmfd != NULL && _cmfd->isFluxUpdateOn()){
src/Cmfd.cpp:  if (_M != NULL){
src/Cmfd.cpp:  if (_A != NULL){
src/Cmfd.cpp:void Cmfd::setNumX(int num_x){
src/Cmfd.cpp:void Cmfd::setNumY(int num_y){
src/Cmfd.cpp:int Cmfd::getNumX(){
src/Cmfd.cpp:int Cmfd::getNumY(){
src/Cmfd.cpp:void Cmfd::setWidth(double width){
src/Cmfd.cpp:void Cmfd::setHeight(double height){
src/Cmfd.cpp:void Cmfd::computeXS(){
src/Cmfd.cpp:  for (int i = 0; i < _num_x * _num_y; i++){
src/Cmfd.cpp:      for (int g = 0; g < _num_cmfd_groups; g++){
src/Cmfd.cpp:           iter != _cell_fsrs.at(i).end(); ++iter){
src/Cmfd.cpp:        for (int b = 0; b < _num_cmfd_groups; b++){
src/Cmfd.cpp:          for (int h = 0; h < _num_moc_groups; h++){
src/Cmfd.cpp:      for (int h = _group_indices[e]; h < _group_indices[e+1]; h++){
src/Cmfd.cpp:             iter != _cell_fsrs.at(i).end(); ++iter){
src/Cmfd.cpp:          for (int g = 0; g < _num_moc_groups; g++){
src/Cmfd.cpp:      for (int g = 0; g < _num_cmfd_groups; g++){
src/Cmfd.cpp:void Cmfd::computeDs(int moc_iteration){
src/Cmfd.cpp:  for (int y = 0; y < _num_y; y++){
src/Cmfd.cpp:    for (int x = 0; x < _num_x; x++){
src/Cmfd.cpp:      for (int surface = 0; surface < 4; surface++){
src/Cmfd.cpp:        for (int e = 0; e < _num_cmfd_groups; e++){
src/Cmfd.cpp:          if (surface == 0 || surface== 2){
src/Cmfd.cpp:          else if (surface == 1 || surface == 3){
src/Cmfd.cpp:          if (cell_next == -1){
src/Cmfd.cpp:            if (_boundaries[surface] == REFLECTIVE){
src/Cmfd.cpp:            else if (_boundaries[surface] == VACUUM){
src/Cmfd.cpp:            if (surface == 0){
src/Cmfd.cpp:            else if (surface == 1){
src/Cmfd.cpp:            else if (surface == 2){
src/Cmfd.cpp:            else if (surface == 3){
src/Cmfd.cpp:            if (fabs(d_tilde) > fabs(d_hat) && moc_iteration != 0){
src/Cmfd.cpp:              if (sense == -1){
src/Cmfd.cpp:                if (1 - fabs(d_tilde)/d_tilde < 1e-8){
src/Cmfd.cpp:                if (1 - fabs(d_tilde)/d_tilde < 1e-8){
src/Cmfd.cpp:FP_PRECISION Cmfd::computeKeff(int moc_iteration){
src/Cmfd.cpp:  if (_A == NULL){
src/Cmfd.cpp:      for (int i = 0; i < _num_x*_num_y; i++){
src/Cmfd.cpp:    catch(std::exception &e){
src/Cmfd.cpp:  for (int iter = 0; iter < 25000; iter++){
src/Cmfd.cpp:    for (int i = 0; i < _num_x*_num_y*_num_cmfd_groups; i++){
src/Cmfd.cpp:                       FP_PRECISION* vec_b, FP_PRECISION conv, int max_iter){
src/Cmfd.cpp:  while (iter < max_iter){
src/Cmfd.cpp:    for (int y = 0; y < _num_y; y++){
src/Cmfd.cpp:      for (int x = y % 2; x < _num_x; x += 2){
src/Cmfd.cpp:        for (int g = 0; g < _num_cmfd_groups; g++){
src/Cmfd.cpp:          for (int e = 0; e < _num_cmfd_groups; e++){
src/Cmfd.cpp:    for (int y = 0; y < _num_y; y++){
src/Cmfd.cpp:      for (int x = 1 - y % 2; x < _num_x; x += 2){
src/Cmfd.cpp:        for (int g = 0; g < _num_cmfd_groups; g++){
src/Cmfd.cpp:          for (int e = 0; e < _num_cmfd_groups; e++){
src/Cmfd.cpp:    for (int i = 0; i < _num_x*_num_y*_num_cmfd_groups; i++){
src/Cmfd.cpp:void Cmfd::rescaleFlux(){
src/Cmfd.cpp:void Cmfd::constructMatrices(){
src/Cmfd.cpp:  for (int y = 0; y < _num_y; y++){
src/Cmfd.cpp:    for (int x = 0; x < _num_x; x++){
src/Cmfd.cpp:      for (int e = 0; e < _num_cmfd_groups; e++){
src/Cmfd.cpp:        for (int g = 0; g < _num_cmfd_groups; g++){
src/Cmfd.cpp:          if (e != g){
src/Cmfd.cpp:        if (x != _num_x - 1){
src/Cmfd.cpp:        if (x != 0){
src/Cmfd.cpp:        if (y != 0){
src/Cmfd.cpp:        if (y != _num_y - 1){
src/Cmfd.cpp:        for (int g = 0; g < _num_cmfd_groups; g++){
src/Cmfd.cpp:void Cmfd::updateMOCFlux(){
src/Cmfd.cpp:  for (int i = 0; i < _num_x*_num_y; i++){
src/Cmfd.cpp:    for (int e = 0; e < _num_cmfd_groups; e++){
src/Cmfd.cpp:      for (int h = _group_indices[e]; h < _group_indices[e+1]; h++){
src/Cmfd.cpp:FP_PRECISION Cmfd::computeDiffCorrect(FP_PRECISION d, FP_PRECISION h){
src/Cmfd.cpp:  if (_optically_thick){
src/Cmfd.cpp:    for (int p = 0; p < _num_polar; p++){
src/Cmfd.cpp:void Cmfd::setFSRMaterials(Material** FSR_materials){
src/Cmfd.cpp:void Cmfd::setFSRVolumes(FP_PRECISION* FSR_volumes){
src/Cmfd.cpp:void Cmfd::setFSRFluxes(FP_PRECISION* scalar_flux){
src/Cmfd.cpp:void Cmfd::setSORRelaxationFactor(FP_PRECISION SOR_factor){
src/Cmfd.cpp:int Cmfd::getNumCmfdGroups(){
src/Cmfd.cpp:int Cmfd::getCmfdGroup(int group){
src/Cmfd.cpp:void Cmfd::setGroupStructure(int* group_indices, int length_group_indices){
src/Cmfd.cpp:  if (_group_indices == NULL){
src/Cmfd.cpp:  if (group_indices == NULL){
src/Cmfd.cpp:    for (int i = 0; i < length_group_indices; i++){
src/Cmfd.cpp:    for (int i = 1; i < length_group_indices; i++){
src/Cmfd.cpp:void Cmfd::initializeFlux(){
src/Cmfd.cpp:  catch(std::exception &e){
src/Cmfd.cpp:  for (int y = 0; y < _num_y; y++){
src/Cmfd.cpp:    for (int x = 0; x < _num_x; x++){
src/Cmfd.cpp:      for (int g = 0; g < _num_cmfd_groups; g++){
src/Cmfd.cpp:void Cmfd::initializeMaterials(){
src/Cmfd.cpp:    for (int y = 0; y < _num_y; y++){
src/Cmfd.cpp:      for (int x = 0; x < _num_x; x++){
src/Cmfd.cpp:  catch(std::exception &e){
src/Cmfd.cpp:void Cmfd::initializeCellMap(){
src/Cmfd.cpp:  for (int y = 0; y < _num_y; y++){
src/Cmfd.cpp:    for (int x = 0; x < _num_x; x++){
src/Cmfd.cpp:void Cmfd::initializeGroupMap(){
src/Cmfd.cpp:  if (_group_indices_map == NULL){
src/Cmfd.cpp:  for (int e = 0; e < _num_cmfd_groups; e++){
src/Cmfd.cpp:    for (int h = _group_indices[e]; h < _group_indices[e+1]; h++){
src/Cmfd.cpp:int Cmfd::findCmfdSurface(int cell, LocalCoords* coords){
src/Cmfd.cpp:int Cmfd::findCmfdCell(LocalCoords* coords){
src/Cmfd.cpp:void Cmfd::setLattice(Lattice* lattice){
src/Cmfd.cpp:void Cmfd::setLatticeStructure(int num_x, int num_y){
src/Cmfd.cpp:Lattice* Cmfd::getLattice(){
src/Cmfd.cpp:void Cmfd::addFSRToCell(int cmfd_cell, int fsr_id){
src/Cmfd.cpp:void Cmfd::setNumMOCGroups(int num_groups){
src/Cmfd.cpp:int Cmfd::getNumMOCGroups(){
src/Cmfd.cpp:int Cmfd::getNumCells(){
src/Cmfd.cpp:void Cmfd::setNumFSRs(int num_fsrs){
src/Cmfd.cpp:void Cmfd::splitCorners(){
src/Cmfd.cpp:  for (int x = 0; x < _num_x; x++){
src/Cmfd.cpp:    for (int y = 0; y < _num_y; y++){
src/Cmfd.cpp:      if (x > 0 && y > 0){
src/Cmfd.cpp:        for (int e = 0; e < ncg; e++){
src/Cmfd.cpp:      else if (x == 0 && y != 0){
src/Cmfd.cpp:        for (int e = 0; e < ncg; e++){
src/Cmfd.cpp:      else if (x != 0 && y == 0){
src/Cmfd.cpp:        for (int e = 0; e < ncg; e++){
src/Cmfd.cpp:      if (x < _num_x - 1 && y > 0){
src/Cmfd.cpp:        for (int e = 0; e < ncg; e++){
src/Cmfd.cpp:      else if (x == _num_x - 1 && y > 0){
src/Cmfd.cpp:        for (int e = 0; e < ncg; e++){
src/Cmfd.cpp:      else if (x < _num_x - 1 && y == 0){
src/Cmfd.cpp:        for (int e = 0; e < ncg; e++){
src/Cmfd.cpp:      if (x < _num_x - 1 && y < _num_y - 1){
src/Cmfd.cpp:        for (int e = 0; e < ncg; e++){
src/Cmfd.cpp:      else if (x == _num_x - 1 && y != _num_y - 1){
src/Cmfd.cpp:        for (int e = 0; e < ncg; e++){
src/Cmfd.cpp:      else if (x != _num_x - 1 && y == _num_y - 1){
src/Cmfd.cpp:        for (int e = 0; e < ncg; e++){
src/Cmfd.cpp:      if (x > 0 && y < _num_y - 1){
src/Cmfd.cpp:        for (int e = 0; e < ncg; e++){
src/Cmfd.cpp:      else if (x == 0 && y != _num_y - 1){
src/Cmfd.cpp:        for (int e = 0; e < ncg; e++){
src/Cmfd.cpp:      else if (x != 0 && y == _num_y - 1){
src/Cmfd.cpp:        for (int e = 0; e < ncg; e++){
src/Cmfd.cpp:      for (int e = 0; e < ncg; e++){
src/Cmfd.cpp:int Cmfd::getCellNext(int cell_num, int surface_id){
src/Cmfd.cpp:  if (surface_id == 0){
src/Cmfd.cpp:  else if (surface_id == 1){
src/Cmfd.cpp:  else if (surface_id == 2){
src/Cmfd.cpp:  else if (surface_id == 3){
src/Cmfd.cpp:bool Cmfd::isOpticallyThick(){
src/Cmfd.cpp:void Cmfd::setOpticallyThick(bool optically_thick){
src/Cmfd.cpp:FP_PRECISION Cmfd::getMOCRelaxationFactor(){
src/Cmfd.cpp:void Cmfd::setMOCRelaxationFactor(FP_PRECISION relax_factor){
src/Cmfd.cpp:void Cmfd::setBoundary(int side, boundaryType boundary){
src/Cmfd.cpp:int Cmfd::getBoundary(int side){
src/Cmfd.cpp:int Cmfd::convertFSRIdToCmfdCell(int fsr_id){
src/Cmfd.cpp:  for (int cell=0; cell < _num_x*_num_y; cell++){
src/Cmfd.cpp:           iter != _cell_fsrs.at(cell).end(); ++iter){
src/Cmfd.cpp:std::vector< std::vector<int> > Cmfd::getCellFSRs(){
src/Cmfd.cpp:void Cmfd::setCellFSRs(std::vector< std::vector<int> > cell_fsrs){
src/Cmfd.cpp:void Cmfd::setFluxUpdateOn(bool flux_update_on){
src/Cmfd.cpp:bool Cmfd::isFluxUpdateOn(){
src/Cmfd.cpp:FP_PRECISION Cmfd::getFluxRatio(int cmfd_cell, int moc_group){
src/Cmfd.cpp:			      int num_tracks){
src/Cmfd.cpp:    if (bc){
src/Cmfd.cpp:    if (bc){
src/Cmfd.cpp:  if (curr_segment->_cmfd_surface_fwd != -1 && fwd){
src/Cmfd.cpp:  else if (curr_segment->_cmfd_surface_bwd != -1 && !fwd){
src/VectorizedSolver.cpp:                                       FP_PRECISION* fsr_flux){
src/VectorizedSolver.cpp:  for (int p=0; p < _num_polar; p++){
src/TrackGenerator.cpp:  if (_geometry->getCmfd() != NULL){
src/TrackGenerator.cpp:      for (int j=0; j < _num_tracks[i]; j++){
src/TrackGenerator.cpp:        if (cmfd != NULL){
src/TrackGenerator.cpp:  for (int i=0; i < num_FSRs; i++){
src/TrackGenerator.cpp:  if (cmfd != NULL){
src/TrackGenerator.cpp:    for (int cell=0; cell < num_cells; cell++){
src/TrackGenerator.cpp:        if (cmfd != NULL){
src/TrackGenerator.cpp:  for (int fsr_id=0; fsr_id < num_FSRs; fsr_id++){
src/TrackGenerator.cpp:  if (cmfd != NULL){
src/TrackGenerator.cpp:    for (int cell=0; cell < num_cells; cell++){
src/TrackGenerator.cpp:      for (int fsr = 0; fsr < num_FSRs; fsr++){
src/TrackGenerator.cpp:void TrackGenerator::generateFSRCentroids(){
src/TrackGenerator.cpp:  for (int r=0; r < num_FSRs; r++){
src/Universe.cpp:double Universe::getMaxY(){
src/Universe.cpp:double Lattice::getMaxY(){
src/Universe.cpp:  for (iter = unique_universes.begin(); iter != unique_universes.end(); ++iter){
src/Universe.cpp:  for (iter = unique_universes.begin(); iter != unique_universes.end(); ++iter){
src/Universe.cpp:    for (int i = 0; i < _num_x; i++){
src/Universe.cpp:int Lattice::getLatticeCell(Point* point){
src/Universe.cpp:  if (fabs(x - left) <= ON_SURFACE_THRESH){
src/Universe.cpp:  else if (fabs(x - right) <= ON_SURFACE_THRESH){
src/TrackGenerator.cpp.orig:  if (_geometry->getCmfd() != NULL){
src/TrackGenerator.cpp.orig:      for (int j=0; j < _num_tracks[i]; j++){
src/TrackGenerator.cpp.orig:        if (cmfd != NULL){
src/TrackGenerator.cpp.orig:  for(int i=0; i < num_FSRs; i++){
src/TrackGenerator.cpp.orig:  for (iter = FSR_keys_map->begin(); iter != FSR_keys_map->end(); ++iter){
src/TrackGenerator.cpp.orig:  if (cmfd != NULL){
src/TrackGenerator.cpp.orig:    for (int cell=0; cell < num_cells; cell++){
src/TrackGenerator.cpp.orig:        if (cmfd != NULL){
src/TrackGenerator.cpp.orig:  for (int fsr_id=0; fsr_id < num_FSRs; fsr_id++){
src/TrackGenerator.cpp.orig:  if (cmfd != NULL){
src/TrackGenerator.cpp.orig:    for (int cell=0; cell < num_cells; cell++){
src/TrackGenerator.cpp.orig:      for (int fsr = 0; fsr < num_FSRs; fsr++){
src/TrackGenerator.cpp.orig:void TrackGenerator::generateFSRCentroids(){
src/TrackGenerator.cpp.orig:  for (int r=0; r < num_FSRs; r++){
src/Material.cpp:  if (_dif_coef == NULL){
src/Material.cpp:  if (_dif_hat == NULL){
src/Material.cpp:  if (_dif_tilde == NULL){
src/Material.cpp:  if (_buckling == NULL){
src/Material.cpp:  for (int i=0; i < _num_groups; i++){
src/Material.cpp:  for (int i=0; i < _num_groups; i++){
src/Material.cpp:  if (_dif_coef == NULL){
src/Material.cpp:  if (_buckling == NULL){
src/Material.cpp:  if (_dif_hat == NULL){
src/Material.cpp:  if (_dif_tilde == NULL){
src/Material.cpp:  if (_dif_tilde == NULL){
src/Material.cpp:Material* Material::clone(){
```

geogunow:
I think I have fixed all the style issues in the PR. Let me know if you spot anything else.

src/Geometry.cpp
PrezNattyGibbs: It's interesting to see where all these are.  A lot of them are in bits of code that haven't been touched in years, and Will would probably blame on his early struggles learning C++.  A rather impressive number of them are `if (something != NULL){`, which probably means they were copy and pasted.  And all the CMFD probably means Sam and Will didn't agree on the style guide until after those got written!

In any case, let's pretend that the space before ` {` is standard practice.  And at some point, someone should kick your grep up a notch and go all the way to a  sed or awk command with a replace!

samuelshaner: Almost all of the style guide issues seem to be mine. I'll go back and fix
those style issues. It's not an issue with disagreement on the style guide;
rather ignorance on my part at the time of writing that code. Going forward
we should definitely try to comply with the Google C++ Style Guide (or
whatever style guide we collectively decide on). Thanks for pointing this
out Geoff.

On Tue, Jun 30, 2015 at 3:30 PM, Nathan A. Gibson <notifications@github.com>
wrote:

> In src/Geometry.cpp
> <https://github.com/mit-crpg/OpenMOC/pull/157#discussion_r33613713>:
>
> > -
> > -    /* Get the lock */
> > -    omp_set_lock(_num_FSRs_lock);
> > -
> > -    /* Recheck to see if FSR has been added to maps after getting the lock */
> > -    if (_FSR_keys_map.find(fsr_key_hash) != _FSR_keys_map.end())
> > -      fsr_id = _FSR_keys_map.at(fsr_key_hash)._fsr_id;
> > +  if (!_FSR_keys_map.contains(fsr_key_hash)){
> > +
> > +    /* Try to get a clean copy of the fsr_id, adding the FSR data
> > +       if necessary where -1 indicates the key was already added */
> > +    fsr_id = _FSR_keys_map.insert_and_get_count(fsr_key_hash, NULL);
> > +    if( fsr_id == -1)
> > +    {
> > +      fsr_data volatile* fsr;
> > +      do{
>
> It's interesting to see where all these are. A lot of them are in bits of
> code that haven't been touched in years, and Will would probably blame on
> his early struggles learning C++. A rather impressive number of them are if
> (something != NULL){, which probably means they were copy and pasted. And
> all the CMFD probably means Sam and Will didn't agree on the style guide
> until after those got written!
>
> In any case, let's pretend that the space before { is standard practice.
> And at some point, someone should kick your grep up a notch and go all the
> way to a sed or awk command with a replace!
>
> —
> <https://github.com/mit-crpg/OpenMOC/pull/157/files#r33613713>.
>



-- 
Regards,
Sam Shaner
Graduate Student

wbinventor: @samuelshaner and @PrezNattyGibbs make some good points. This is really a relic of the old days when OpenMOC was in its very infancy, and we didn't have a consistent style guide. Now that we do - largely based on Google's - let's try to stick to it. Of course this means that there is still some purging left to do to get everything consistent, but we owe it to ourselves to make these sorts of comments during code review to ensure the consistency of new code.

Also, for the record, as @geogunow noted above, there are 253 instances of unspaced parentheses and brackets ") " throughout the code. But there are 942 instances of spaces between parentheses and brackets ") {". So we have about 26% of our inconsistent codebase needs some housecleaning :-)

wbinventor:
So is this ready to go?

samuelshaner:
I think it's ready to go. I'll merge it now.

Issue #155, opened 2015-06-15
Problem with Setting Boundary Conditions


geogunow:
I have been writing a test case today to showcase an example where the user might want cells of unequal x and y lengths. That part seems to be working fine - OpenMOC seems to handle uneven mesh properly now. However, it does seem to have a problem with applying the correct boundary conditions. When I run my case with one vacuum boundary (the left to be specific), I would expect to see a pure one-directional gradient as for the flux. However, I instead see the flux has the profile shown below:

![fsr-flux-group-1](https://cloud.githubusercontent.com/assets/2420416/8172028/36bd934c-1382-11e5-839a-adcd8bdafab1.png)

I suspect this should be an easy to fix bug in the setting of boundary conditions. Here is the source code for this case:

```
import numpy
from openmoc import *
import openmoc.log as log
import openmoc.plotter as plotter
from openmoc.options import Options


###############################################################################
#######################   Main Simulation Parameters   ########################
###############################################################################

options = Options()

num_threads = options.getNumThreads()
track_spacing = options.getTrackSpacing()
num_azim = options.getNumAzimAngles()
tolerance = options.getTolerance()
max_iters = options.getMaxIterations()

log.set_log_level('NORMAL')

log.py_printf('TITLE', 'Simulating a one group homogeneous infinite medium...')
log.py_printf('HEADER', 'The reference keff = 1.43...')


###############################################################################
#######################   Main Simulation Parameters   ########################
###############################################################################

log.py_printf('NORMAL', 'Creating materials...')

basic_material = Material(name='1-group infinite medium')
basic_material.setNumEnergyGroups(1)
basic_material.setSigmaA(numpy.array([0.069389522]))
basic_material.setSigmaF(numpy.array([0.0414198575]))
basic_material.setNuSigmaF(numpy.array([0.0994076580]))
basic_material.setSigmaS(numpy.array([0.383259177]))
basic_material.setChi(numpy.array([1.0]))
basic_material.setSigmaT(numpy.array([0.452648699]))


###############################################################################
###########################   Creating Surfaces   #############################
###############################################################################

log.py_printf('NORMAL', 'Creating surfaces...')

L = 100.0
left = XPlane(x=-L/2, name='left')
right = XPlane(x=L/2, name='right')
top = YPlane(y=L/2, name='top')
bottom = YPlane(y=-L/2, name='bottom')

left.setBoundaryType(VACUUM)
right.setBoundaryType(REFLECTIVE)
[...]

PR #158, opened 2015-06-18
Fixed track boundary condition assignment in TrackGenerator

samuelshaner:
The TrackGenerator currently sets the boundary conditions for tracks reflecting off the XMax surface to the boundary condition of the XMin surface. This PR fixes issue #155 by fixing the track BC assignment.


PrezNattyGibbs:
Easy enough fix!  Looks like it works to me.

PR #161, opened 2015-06-25
Python/C++ memory ownership rules

wbinventor:
This PR improves the memory ownership and cleanup of objects owned (created in) Python and pointed to by one or more C/C++ objects. The current version of the code allows for ambiguities as to whether Python-wrapped C++ objects pointed to from both Python and C++ are owned by their Python or C++ variables. These ambiguities may lead to segmentation faults in certain situations - such as during parametric sweeps over many simulation parameters or cross-section libraries from within Python.

For example, in the present version of the code, the following will lead to a segmentation fault:

```python
# Initialize circular fuel pin surface
circle = openmoc.Circle(x=0.0, y=0.0, radius=0.45)

# Initialize the cells for the fuel pin and moderator
fuel = openmoc.Cell(name='fuel cell')
moderator = openmoc.Cell(name='moderator cell')

# Add the circle surface to each cell
fuel.addSurface(halfspace=-1, surface=circle)
moderator.addSurface(halfspace=+1, surface=circle)

# Initialize a NEW circular fuel pin surface with the SAME var name
circle = openmoc.Circle(x=0.0, y=0.0, radius=0.5)

# Finish building the geometry
...

# Generate tracks using ray tracing across the geometry
track_generator = openmoc.TrackGenerator(geometry)
track_generator.generateTracks()
```

The reason for this is as follows. when the first ``circle`` is instantiated, it is a C++ object with a SWIG-generated Python wrapper which it **owns** the underlying memory used by the object. Even after ``circle`` is assigned as a surface for the ``moderator`` and ``fuel`` cells, the Python ``circle`` variable still *thinks* that it owns the object. Hence, when the ``circle`` variable is redefined to a brand new ``Circle`` object, the Python garbage collectors goes ahead and frees the memory for the old ``circle`` reference, leaving ``fuel`` and ``moderator`` cells with dangling pointers to the original (deallocated) surface. Finally, during ray tracing, the cells attempt to reference the deallocated surface, leading to a segfault.

To resolve situations like this, this PR uses the the [``thisown``](http://www.swig.org/Doc1.3/Python.html) class attributes generated by SWIG to specify ownership of objects by Python or C++. The ``thisown`` is an attribute of the Python wrapper for each C++ object which specifies whether Python is in charge of memory deallocation for the object (``thisown == 1``) or if C++ is in charge (``thisown == 0``). In the old version of the code, ``thisown == 1`` for any and all OpenMOC objects throughout the program lifetime, meaning that Python would *always* take charge of memory deallocation.

Now, when an object is set/assigned/registered to another object (*e.g.*, a ``Surface`` is added to a ``Cell``), the ``thisown`` attribute on the Python wrapper is set to 0. This indicates that the underlying C++ object is now in charge of deleting the object when its destructor is called (either from within Python or C++). If an object is removed from another object (*e.g.*, a ``Surface`` is removed from a ``Cell``), then ``thisown`` is reset to 1 to indicate that Python is in charge of deallocating memory to the object. In this case, if there is a variable in Python which points to the object removed, then the object will persist in memory until Python no longer has a reference to the object; otherwise, if no reference points to the object, the Python garbage collector will delete it.

I believe that the new memory ownership model introduced here will make it significantly easier to initialize objects in Python in any order without running into issues (specifically segfaults). 

In addition, this PR modularizes SWIG interface files into multiple smaller files to make them easier to read and extend.


PrezNattyGibbs:
On a first pass, this looks like a good change.  I had no idea of the subtleties of memory deallocation!  I'm going to spend a bit more time on this one later.

geogunow:
Makes sense to me. I have actually run into the seg-faults myself while during parameter convergence studies. Although I don't think my inputs had anything like the example you mentioned so I don't think this PR would fix my issues.

But nevertheless, I did I quick review of the PR and everything seems in order. I'll take a more thorough look at it tonight.

samuelshaner:
This seems to work fine on NSECluster, but I get the following error when running on my mac. Any ideas?

[sam simple-lattice]$ python simple-lattice.py
[  NORMAL ]  Importing materials data from HDF5...
[  NORMAL ]  Creating surfaces...
[  NORMAL ]  Creating cells...
[  NORMAL ]  Creating universes...
  File "simple-lattice.py", line 108, in <module>
    pin1.addCell(large_fuel)
  File "/Users/sam/Library/Python/2.7/lib/python/site-packages/openmoc/openmoc.py", line 2149, in addCell
    args[0].thisown = 0
NameError: global name 'args' is not defined
Segmentation fault: 11

wbinventor:
That is very odd indeed. Can you send me the "openmoc.py" file that is generated when you build this on your Mac? Also, what version of SWIG are you using on your Mac?

The "args" is what you see as a parameter to the different routines in "openmoc.py" when building on nsecluster or my Linux machines. This can be seen from the ``%append`` block for ``Universe::addCell(...)`` in the SWIG interface file [here](https://github.com/mit-crpg/OpenMOC/pull/161/files#diff-444a22fd3e5f261c7936292f41a3c638R21). Perhaps your version of SWIG is naming the parameters something other than "args" which is the cause of this error.

samuelshaner:
Here's the openmoc.py file from my mac. I'm using swig 3.0.5, which I think
is the latest. Not sure if SWIG is name the parameters something other than
"args", but hopefully the openmoc.py file will help explain.

On Tue, Jun 30, 2015 at 9:37 PM, Will Boyd <notifications@github.com> wrote:

> That is very odd indeed. Can you send me the "openmoc.py" file that is
> generated when you build this on your Mac? Also, what version of SWIG are
> you using on your Mac?
>
> The "args" is what you see as a parameter to the different routines in
> "openmoc.py" when building on nsecluster or my Linux machines. This can be
> seen from the %append block for Universe::addCell(...) in the SWIG
> interface file here
> <https://github.com/mit-crpg/OpenMOC/pull/161/files#diff-444a22fd3e5f261c7936292f41a3c638R21>.
> Perhaps your version of SWIG is naming the parameters something other than
> "args" which is the cause of this error.
>
> —
> <https://github.com/mit-crpg/OpenMOC/pull/161#issuecomment-117387811>.
>



-- 
Regards,
Sam Shaner
Graduate Student

wbinventor:
@samuelshaner it doesn't look like your openmoc.py file came through. Perhaps send it to me by email?

geogunow:
I was going to submit another issue regarding a bug I found but I am pretty sure this PR would fix it. The bug occurs when calling
```
string = lattices[-1].toString()
```
multiple times. The following message is printed to the screen:
```
swig/python detected a memory leak of type 'std::string *', no destructor found.
```

wbinventor:
No, I'm pretty sure it is a separate issue. See my reply to your PR. A
quick sanity check would be to enlarge the logger buffer to something
insanely huge like 1E6 and see if you still get a default (you shouldn't).

geogunow:
I am pretty sure this occurs because a pointer to a ``std::string`` object is returned to Python. A single call of this has no problems/ Although it doesn't return anything useful as Python doesn't seem to know how to handle the string object.

If I write
```
string = lattices[-1].toString()
print string
```
Python outputs
```
<Swig Object of type 'std::string *' at 0x7fbf56699690>
```
which isn't great but is not technically a bug.  The memory leak occurs when, for a second time, 
```
string = lattices[-1].toString()
```
is called since Python will try to destruct the current version of ``string``. This would all be solved if we make the ``Lattice::toString`` function only accessible from C++ as it doesn't seem to play well with Python.

wbinventor:
Ok well that is interesting. Have you tried printing
"lattices[-1].toString()" ? If this works, it would confirm that the issue
is in the logger rather than the lattice's string builder.

geogunow:
@wbinventor yes, see the above comment

wbinventor:
@samuelshaner - it appears that SWIG 2 (default on Ubuntu 14.10) uses varargs by default to wrap a lot of class methods, while SWIG 3 does not. This is the reason that this PR breaks on your machine and not on mine or nsecluster. Although I like the treatment of class method arguments in SWIG 3 much better than 2 (it is much cleaner and transparent), I'm not sure how to make this PR cross-compatible for both versions 2 and 3.

I've posted this issue to StackOverflow here](http://stackoverflow.com/questions/31258380/swig-default-treatment-of-varargs-compatibility-between-version-2-and-3) as well as to the SWIG mailing list. Based on prior experience, I expect a reply within a few hours and will update you once I get this sorted out.

wbinventor:
@samuelshaner - can you give this another try? I'm now using ``locals()`` to determine the name of the parameters given by SWIG to these routines. This is anathema to me, but makes sense in this situation to provide SWIG 2 and 3 cross-compatibility. It seems to work for 2, let me know if it also works for 3.

samuelshaner:
This fix works for the `addCell` method, but I'm getting the same error for `setUniverses`:

[sam simple-lattice]$ python simple-lattice.py 
[  NORMAL ]  Importing materials data from HDF5...
[  NORMAL ]  Creating surfaces...
[  NORMAL ]  Creating cells...
[  NORMAL ]  Creating universes...
[  NORMAL ]  Creating simple 4 x 4 lattice...
  File "simple-lattice.py", line 128, in <module>
    [pin2, pin3, pin2, pin3]])
  File "/Users/sam/Library/Python/2.7/lib/python/site-packages/openmoc/openmoc.py", line 2349, in setUniverses
    universes = locals()['args'][0]
KeyError: 'args'

wbinventor:
Alright so I installed SWIG 3 alongside 2 on my machine and have verified that the latest commit works for both versions. Can you give it yet another try @samuelshaner? Sorry for the trial and error here.

PrezNattyGibbs:
Props to @wbinventor for working the word "anathema" into a Github comment.

wbinventor:
Haha, was it too forced?

samuelshaner:
I just tried a fresh install of OpenMOC and merged with your `thisown` branch, but got the same error. I'm going to install swig 3.0 on AWS and give it a try on there to see if I can reproduce the error on Linux or whether it might just be something with mac, my install of swig, or something else.

samuelshaner:
I just ran OpenMOC on AWS with Swig 2 and Swig 3 (3.0.6). It works just fine with Swig 2, but I get the same error that I got on mac with version Swig 3. 

@wbinventor - are you sure Swig 3 is actually being used after you installed swig 3? Not really sure what else the issue could be. I'll look at the .i files a bit to see if I can get anything working. If it wouldn't be too much of a hassle could you try installing swig 3 on AWS and try that out? Swig 3 doesn't seem to be on apt-get, so I just installed it by scp'ing the swig tar file up and installing it with:

`>>> tar -xf swig-3.0.6.tar`
`>>> cd swig`
`>>> ./configure --without-pcre`
`>>> make`
`>>> make install`

wbinventor:
@samuelshaner - I think I figured out what was wrong. I completely removed all old SWIG installations and installed the latest SWIG 3.0.6 from source as you recommended and found the same issue. The latest commit works on my machine - let me know if you find any problems with it.

PR #163, opened 2015-07-01
Update to src files to comply with Google C++ style guide conditionals and function calls formatting

samuelshaner:
As recognized in PR #157 there are many cases where OpenMOC does not comply with the Google C++ style guide, in particular in the formatting of conditionals and function calls. This PR seeks to update the src files to comply with the Google C++ style guide for those cases. 

In reviewing this PR, if you see any further violations of the Google C++ style guide, please do share!


wbinventor:
This looks good to me.

PR #164, opened 2015-07-01
Adding simple gradient sample cases

geogunow:
In this PR, I add a few sample inputs to showcase examples where the user might desire to create a lattice with unequal x and y widths.

To showcase this, I have created two sample cases. The first is a one-directional gradient with the solution shown below

fsr-flux-group-1

The second is a two dimensional gradient with a symmetrical solution. 

fsr-flux-group-1

The keen observer might notice that these are almost the same case, just the second case is half the size and solved across the whole domain. Both of these cases have analogous files that include CMFD acceleration, much like the current C5G7 benchmark inputs.


wbinventor:
Minor point - can we change the new subdirectories and filenames to be shorter? In particular:

* "one-directional-gradient" -> "one-directional"
* "one-directional-gradient.py" -> "one-directional.py"
* ...

This will reduce the redundancy of including "gradient" in everything and make it easier to read the contents in each directory.

wbinventor:
Thanks for this PR @geogunow! It's nice to include these (finally working) input files now. 

My only comment is really more of a question for you and the other developers. In a number of our sample inputs (C5G7, LRA, and now these), we redundantly build the same materials/geometry model in multiple input scripts. What do you think of migrating the redundant pieces into shared (importable) "geometry.py" files such that the only differences in each script would be the setup of the Solver / CMFD objects?

geogunow:
@wbinventor that sounds like a great idea. Should that be included in this PR?

wbinventor:
Perhaps for your new inputs. It's up to you for you want to do it for C5G7
and LRA in this PR (it's certainly not what this PR was for I know).

sample-input/gradients/one-directional/one-directional-cmfd.py
```diff
@@ -0,0 +1,155 @@
...
+basic_material.setSigmaT(numpy.array([0.452648699]))
+
+
+###############################################################################
+###########################   Creating Surfaces   #############################
+###############################################################################
+
+log.py_printf('NORMAL', 'Creating surfaces...')
+
+L = 100.0
```
wbinventor: Can we move ``L`` as a runtime parameter to the top block "Main Simulation Paramters" in the input file? And perhaps rename it to something more intuitive such as ``length`` or at least comment it in each input file.

sample-input/gradients/one-directional/one-directional-cmfd.py
```diff
@@ -0,0 +1,155 @@
...
+root_universe = Universe(name='root universe')
+root_universe.addCell(root_cell)
+
+###############################################################################
+############################    Creating Lattices   ###########################
+###############################################################################
+
+log.py_printf('NORMAL', 'Creating 100 x 1 lattice...')
+
+num_cells_x = 100
```
wbinventor: ditto, put these at the top of the script

wbinventor:
These new example scripts run nicely on my machine and this looks just about ready to go. After addressing my comments about the runtime parameters, as well as revising the scripts to reuse a ``geometry.py`` script with the geometric specs, this should be good to go.

geogunow:
@wbinventor How would you do that? I tried creating a ``build_cube`` routine in ``geometry.py`` that takes parameters ``length``, ``num_cells_x``, ``num_cells_y``, ``material``, and ``boundary_conditions`` as input but when the function returns the ``Geometry`` object it seems that Python destructs some of the necessary structure after the function returns and I get segmentation faults.

wbinventor:
You can just copy-paste all of the lines of code that constructs the ``Geometry`` object into a new "geometry.py" file. Then in each "xxx-directional.py" file you can import the ``Geometry`` as so:

```python
from geometry import Geometry
```

Does this make sense? We should do the same for both C5G7 and LRA.

geogunow:
@wbinventor The problem with that is the "geometry.py" file would require input parameters such as ``length`` and ``num_cells_x``. It cannot access those parameters in the main python files because they are not in the same scope. And I don't see a way I can pass input parameters while importing a python module.

wbinventor:
Why not just put those params in "geometry.py"?

geogunow:
The boundary conditions change between the cases

wbinventor:
Well, one solution is to have two "geometry.py" files for both "one-directional" and "two-directional". This is what I was thinking of actually, but if you want to generalize further I can show you how to extract and iterate over the bounding surfaces in Python.

geogunow:
@wbinventor I would be interested, but it seems ``Surface::setBoundaryType`` must be called before ``Cell::addSurface`` is called for the boundary conditions to be implemented correctly. In the current structure, ``Cell::addSurface`` would be called in "geometry.py".

wbinventor:
Yes, but you can call ``Surface::removeSurface()`` and then ``Surface::addSurface()`` (or even just call ``Surface::addSurface()`` twice). You can call ``Cell::getSurfaces()`` to get a Python dictionary of ``Surfaces`` combined with ``Universe::getCells()`` and/or ``Universe::getAllCells()`` (which return a ``dict`` of ``Cells``) to do what you need.

geogunow:
@wbinventor I just restructured the cases. Let me know if this fits your standards.

wbinventor:
This looks really good to me! Thanks for taking the time to refactor this to set a good example for how to reuse geometries in multiple scripts. The only thing I'd like to see before merging is a move of all Python imports to the top of each script per the PEP8 standard [here](https://www.python.org/dev/peps/pep-0008/#imports).

geogunow:
Strictly speaking, that's impossible without some magic because I need to call ```import sys``` and then ```sys.path.append('..')``` before I can call ``import boundaries``. I suppose I could throw all that up top if you'd prefer. I put it in the "Load Cubic Geometry" section to make it more understandable to the user.

wbinventor:
Alright, I'm fine with that if @samuelshaner is as well. I'm certainly not
the best developer to be preaching the PEP8 gospel :-)

On Wed, Jul 15, 2015 at 9:58 AM, Geoffrey Gunow <notifications@github.com>
wrote:

> Strictly speaking, that's impossible without some magic because I need to
> call import sys and then sys.path.append('..') before I can call import
> boundaries. I suppose I could throw all that up top if you'd prefer. I
> put it in the "Load Cubic Geometry" section to make it more understandable
> to the user.
>
> —
> <https://github.com/mit-crpg/OpenMOC/pull/164#issuecomment-121679546>.
>



-- 
Will Boyd
Nuclear Science & Engineering
Cell: 423.413.8469
wbinventor@gmail.com

sample-input/gradients/one-directional/one-directional-cmfd.py
```diff
@@ -0,0 +1,83 @@
...
+###############################################################################
+###########################   Creating CMFD Mesh    ###########################
+###############################################################################
+log.py_printf('NORMAL', 'Creating Cmfd mesh...')
+
+cmfd = Cmfd()
+cmfd.setMOCRelaxationFactor(0.6)
+cmfd.setSORRelaxationFactor(1.5)
+cmfd.setLatticeStructure(51,1)
+cmfd.setGroupStructure([1])
```
samuelshaner: I would recommend getting rid of this line as CMFD is using the same group structure as MOC. I'm actually surprised this even works since the group structure should start at 1 and end with the number of MOC energy groups + 1.

sample-input/gradients/cube.py
```diff
@@ -0,0 +1,98 @@
...
+#########################   Creating the Geometry   #########################
+#############################################################################
+
+log.py_printf('NORMAL', 'Creating geometry...')
+
+geometry = Geometry()
+geometry.setRootUniverse(root_universe)
+geometry.initializeFlatSourceRegions()
+
+#return geometry
```
samuelshaner: Can you remove this?

samuelshaner:
How about setting the BCs of the cube to be all reflective and then resetting the BCs as needed in the one and two directional input files? This way the `import sys`, `sys.path.append('..')`, and `from cube import *` could all be at the top of the input files. For the one directional files, you would just need to add `left.setBoundaryType(VACUUM)` in the input file before you call `geometry.initializeFlatSourceRegions()`. You could then delete and `boundaries.py` and possibly rename `cube.py` to `reflective_cube.py` if you felt it was necessary.

wbinventor:
I like @samuelshaner's proposal. Also, not to get back on the PEP8 campaign trail, but I would also prefer if you do something like the following:

```python
from cube import left, geometry
```

rather than:

```python
from cube import *
```

sample-input/gradients/one-directional/one-directional-cmfd.py
```diff
@@ -0,0 +1,83 @@
...
+sys.path.append('..')
+import boundaries
+boundaries.left_bc = VACUUM
+boundaries.right_bc = REFLECTIVE
+boundaries.top_bc = REFLECTIVE
+boundaries.bottom_bc = REFLECTIVE
+from cube import *
+
+geometry.setCmfd(cmfd)
+geometry.initializeFlatSourceRegions()
```
wbinventor: This isn't needed since you initialize FSRs in "cube.py"

samuelshaner: Actually it is needed in the `one-directional-cmfd.py` since `geometry.initializeFlatSourceRegions()` initializes CMFD, which sets the boundary conditions for CMFD. Thus if the BCs change, cmfd needs to be reinitialized by the geometry or the appropriate BC in CMFD needs to be set by the user. I would recommend keeping as is in this file, but removing in the other files without CMFD.

With this in mind, we might consider having `initializeCmfd()` method be called from the `TrackGenerator::generateTracks()` method instead of in `Geometry::initializeFlatSourceRegions` just prior to track generation to avoid any issues of differing BCs. If others agree, this can be added as an issue for a later PR.

geogunow:
@samuelshaner You need to set the boundaries before calling ``Cell::addSurface`` or it doesn't work

samuelshaner:
Why?

geogunow:
They don't work otherwise. See my comment above.

samuelshaner:
Have you tried it? I implemented what I described in my comment and it seemed to work just fine.

geogunow:
@samuelshaner It should run just fine but did you check the eigenvalue? When I ran it the eigenvalue was off, implying the boundary conditions were not truly being updated.

samuelshaner:
I see why now. It's because the `Cell:findBoundingBox()` method only updates the Cell's BCs when a surface is added. After a surface is added, the BC cannot be changed.

wbinventor:
Make an issue for this and the CMFD initialization. These are the sorts of
things we need to fix for the Python interface to be truly configurable and
useful

geogunow:
I think this should be good to merge now

samuelshaner:
I would recommend we wait until issues 170 and 171 are resolved so we can get rid of the double `Geometry::initializeFlatSourceRegions()` and not have the imports in the middle of the input files. The issues require short PRs and I'll submit them tonight.

geogunow:
@samuelshaner I reformed the code in the most recent commit to not call ``Geometry::initializeFlatSourceRegions()`` and not have imports in the middle of the files.

sample-input/gradients/one-directional/one-directional.py
```diff
@@ -0,0 +1,71 @@
+import numpy
+from openmoc import *
+import openmoc.log as log
+import openmoc.plotter as plotter
+from openmoc.options import Options
+import sys
+sys.path.append('..')
+from cube import geometry, root_cell, left, right, top, bottom
```
wbinventor: don't you only need to import the one/two surfaces with BCs that change?

sample-input/gradients/one-directional/one-directional-cmfd.py
```diff
@@ -0,0 +1,100 @@
...
+
+###############################################################################
+#########################   Load the Cubic Geometry   #########################
+###############################################################################
+
+log.py_printf('NORMAL', 'Importing cubic geometry...')
+
+import sys
```
wbinventor: this file hasn't been updated

samuelshaner:
I would still recommend waiting so the `Geometry::initializeFlatSourceRegions()` can be moved back to `cube.py` and `root_cell.findBoundingBox()` can be removed.

wbinventor:
I agree @samuelshaner, there's no rush and these types of PRs do a good job of exposing those sorts of weaknesses in the existing codebase. Let's merge in your fixes first and then this PR.

geogunow:
@wbinventor I would prefer the user set all boundaries in the file to avoid confusion.

wbinventor:
sure, that makes sense

geogunow:
I think this is ready to merge now

sample-input/gradients/cube.py
```diff
@@ -0,0 +1,80 @@
+import numpy
+from openmoc import *
+import openmoc.log as log
```
wbinventor: The ``openmoc.log``, ``openmoc.plotter`` and ``openmoc.options`` submodules are not used in this file

sample-input/gradients/one-directional/one-directional.py
```diff
@@ -0,0 +1,68 @@
+import numpy
```
wbinventor: NumPy is not used except in "cube.py"

sample-input/gradients/two-directional/two-directional-cmfd.py
```diff
@@ -0,0 +1,80 @@
+import numpy
```
wbinventor: ditto NumPy

sample-input/gradients/one-directional/one-directional-cmfd.py
```diff
@@ -0,0 +1,80 @@
+import numpy
```
wbinventor: ditto NumPy

sample-input/gradients/two-directional/two-directional.py
```diff
@@ -0,0 +1,68 @@
+import numpy
```
wbinventor: ditto NumPy

wbinventor:
This looks good and works fine on my system @geogunow. Sorry to make two more minor comments, trying to up the QA since the code is really moving forward these days. It's on you to hold my PRs to the same or higher standards ;-)

geogunow:
@wbinventor I removed numpy altogether. We really don't need it.

Issue #159, opened 2015-06-19
C++ Alternative Build System


geogunow:
I'm not sure if this belongs in the issues section, but it might really help to have an alternative build system that just uses C++ (instead of SWIG) with compiled inputs for common cases (such as C5G7). It looks like this could be simple as the C++ source files wouldn't need to be changed at all.

Adding this option would allow form much easier performance profiling and debugging as well. Everyone in CESAR would rejoice. Let me know what you guys think and if there are any issues you see that might arise. I would envision this as a separate "profiling" directory containing a makefile and then maybe two or three hard-baked C++ input files like C5G7.cpp or LRA.cpp.

PrezNattyGibbs:
Certainly seems like a reasonable use of the issue tracker!

It'd also help with running the code through valgrind and various other tools that get cranky with Python.  I support this (but don't volunteer to do it!).

geogunow:
I can do this, but I would first like to hear from everyone if they foresee any issues. Of course I won't put in plotting, just the solver since C++ plotting can be messy. But there might be other python interfaces that might be difficult. In particular, how would I pull the material data? Right now it seems that is done entirely in python with an included python file in the openmoc directory (materialize.py). How could this be changed for C++?

PrezNattyGibbs:
The `materialize` module is in Python, not C++, yes.  But all the material setting routines are also in C++, so it's only an issue of pulling data out of HDF5.  The c5g7 data really isn't that much, so you could even just hard code a data block for your purposes if you didn't want to deal with HDF5.

Lattices could get a little funny without the Python interface (and numpy arrays).  But I don't think it'd be too painful to do it with C++ arrays.

I don't think there are any other places where you'd have to do more than declaring all your variable names and then directly translating the input files line by line.

wbinventor:
It probably comes with no surprise that I think a pure C++/CUDA build with compiled inputs is a return to the dark ages from a usability standpoint. Nonetheless, I fully understand that Python/C++ integration is a very ineffective way to build the code for performance profiling and analysis. Hence, I think that a pure C++/CUDA build option is a fine goal to have.

[...]

PR #165, opened 2015-07-02
Adding Alternative C++ Build System for Profiling

geogunow:
In this PR I introduce an alternative build system for OpenMOC that relies solely on C++. This is helpful primarily for High Performance Computing (HPC) and profiling reasons. While the SWIG interface allows users great flexibility and ease of use when building geometries, the SWIG interface also poses challenges for profiling the code and installing it on newer HPC architectures which might not have all the python requirements needed by OpenMOC.

I have created a new directory entitled "profile" which includes a standard Makefile which can be used to compile hard-coded C++ inputs. I have provided example inputs for C5G7 (with and without CMFD acceleration), the one-directional and two-directional gradient cases, and for the homogeneous one group benchmark. Users of this added functionality should be proficient enough in C++ to be able to edit the Makefile in order to compile the case of interest. It is suggested that users of this directory should concentrate on profiling the test cases before attempting to write their own C++ input file.

The source code of OpenMOC would ideally not be modified. However, slight alterations were made in order to compile without SWIG. Still, adjustments to the OpenMOC source will be directly adopted by the profiling directory so that it profiles the most recent version. 

Any suggestions on alternative strategies or styles are welcomed. 

This PR is a response to issue #159.


config.py
```diff
@@ -373,10 +373,12 @@ class configuration:
                               ('DOUBLE', None),
                               ('NVCC', None),
                               ('CCACHE_CC', 'nvcc')]
-  # define OPENMP
+
+  # define OPENMP and SWIG (for log output)
   for compiler in macros:
     for precision in macros[compiler]:
       macros[compiler][precision].append(('OPENMP', None))
+      macros[compiler][precision].append(('__SWIG__', None))
```
wbinventor: Any reason you use ``__SWIG__`` here rather than ``SWIG``? If so, perhaps all of our macros in "config.py" should be renamed with the double underscores.

profile/Makefile
```diff
@@ -0,0 +1,170 @@
...
+#===============================================================================
+
+# Regular gcc Compiler
+ifeq ($(COMPILER),gnu)
+  CC = g++
+endif
+
+# intel Compiler
+ifeq ($(COMPILER),intel)
+  CC = icc
```
wbinventor: If the Intel compiler is used, you might also append "VectorizedSolver.cpp" to the list of sources. I'd actually be very interested to see that guy profiled to determine wth is wrong with it.

profile/Makefile
```diff
@@ -0,0 +1,170 @@
...
+endif
+
+# IBM XL Compiler
+ifeq ($(COMPILER),ibm)
+  CC = mpicc
```
wbinventor: Last I knew, this wasn't the IBM C++ compiler. At least on BG/Q back in 2013 we used "bgxlc++", in particular, a threadsafe version called "bgxlc++_r". But perhaps things have changed.

profile/Makefile
```diff
@@ -0,0 +1,170 @@
...
+CFLAGS += -I/usr/lib/python2.7/dist-packages/numpy/core/include 
+
+# Vector Flags
+CFLAGS += -DVEC_LENGTH=8
+CFLAGS += -DVEC_ALIGNMENT=16
+
+# Optimization Flags
+ifeq ($(OPTIMIZE),yes)
+ifeq ($(COMPILER),gnu)
+  CFLAGS += -Ofast -ffast-math -ftree-vectorize -msse2
```
wbinventor: append the "-DGNU" macro here - I don't think it is referenced anywhere right now, but would be good to have for the future

profile/Makefile
wbinventor: Yeah, I think this assumes that the user has an MPI compiler wrapper for "bgxlc++_r"?

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
+#include "../src/CPUSolver.h"
+#include "../src/Solver.h"
```
wbinventor: Is this second include needed? It should be included with "CPUSolver.h"

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+      0.0, 0.0, 0.0, 1.2525E-4, 0.271401, 0.010255, 1.0021E-8,
+      0.0, 0.0, 0.0, 0.0, 0.0012968, 0.265802, 0.016809,
+      0.0, 0.0, 0.0, 0.0, 0.0, 0.0085458, 0.27308};
+  chi["UO2"] = std::array<double, 7> {0.58791, 0.41176, 3.3906E-4, 1.1761E-7,
+    0.0, 0.0, 0.0};
+  sigmaT["UO2"] = std::array<double, 7> {0.177949, 0.329805, 0.480388, 
+    0.554367, 0.311801, 0.395168, 0.564406};
+
+  /* Define MOX-4.3% cross-sections */
+  sigmaA["MOX-4.3%"] = std::array<double, 7> {0.0084339, 0.0037577, 0.02797,
```
wbinventor: I think you need an "\" before the percent symbols "%" here

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+  int max_iters = 1000;
+
+  /* Set logging information */
+  set_log_level("NORMAL");
+  log_printf(TITLE, "Simulating the OECD's C5G7 Benchmark Problem...");
+ 
+  /* Define material properties */
+  log_printf(NORMAL, "Defining material properties..."); 
+ 
+  std::map<std::string, std::array<double, 7> > sigmaA;
```
wbinventor: Per the Google C++ Style Guide ([link](https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Variable_Names)), as well as OpenMOC's own Style Guide ([link](https://mit-crpg.github.io/OpenMOC/devguide/style_guide.html#variable-names)), variable names should be all lowercase with words separated by underscores.

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+      0.0, 0.0, 0.0, 0.0, 9.1726E-4, 0.316765, 0.23877,
+      0.0, 0.0, 0.0, 0.0, 0.0, 0.049792, 1.09912};
+  chi["Guide Tube"] = std::array<double, 7> {0, 0, 0, 0, 0, 0, 0};
+  sigmaT["Guide Tube"] = std::array<double, 7> {0.126032, 0.29316, 0.28424, 
+    0.28096, 0.33444, 0.56564, 1.17215};
+
+  /* Create materials */
+  log_printf(NORMAL, "Creating materials...");
+  std::map<std::string, Material*> materials;
+  int numEnergyGroups = 7;
```
wbinventor: Define this at the top with the other simulation parameters. I think it would also be best to use this variable to define the size of your arrays in a more programmatic (and transparent) fashion, but that isn't necessary here.

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+  std::map<std::string, std::array<double, 7> >::iterator it;
+  int id_num = 0;
+  for (it = sigmaA.begin(); it != sigmaA.end(); it++) {
+    
+    std::string name = it->first;
+    materials[name] = new Material(id_num, name.c_str());
+    materials[name]->setNumEnergyGroups(numEnergyGroups);
+    id_num++;
+    
+    materials[name]->setSigmaA(sigmaA[name].data(), sigmaA[name].size());
```
wbinventor: Why not use ``numEnergyGroups`` (or ``num_groups`` after addressing the Style Guide issues) here rather than calling ``size()``?

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+  mox7.addCell(&mox7_cell);
+  mox7.addCell(&moderator_ring1);
+  mox7.addCell(&moderator_ring2);
+  mox7.addCell(&moderator_ring3);
+
+  /* 8.7% MOX pin cell */
+  Cell mox87_cell;
+  mox87_cell.setNumRings(3);
+  mox87_cell.setNumSectors(8);
+  mox87_cell.setFill(materials["MOX-8.7%"]);
```
wbinventor: Escape character for "%" necessary here?

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+                        1, 1, 2, 1, 1, 2, 1, 1, 2, 1, 1, 2, 1, 1, 2, 1, 1,
+                        1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+                        1, 1, 1, 2, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 1, 1, 1,
+                        1, 1, 1, 1, 1, 2, 1, 1, 2, 1, 1, 2, 1, 1, 1, 1, 1,
+                        1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+                        1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1};
+
+    std::map<int, Universe*> names = {{1, &uo2}, {2, &guide_tube}, 
+                                      {3, &fission_chamber}};
+    for(int n=0; n<17*17; n++)
```
wbinventor: This certainly works, or you could consider getting C++ fancy with [``std::transform``](http://en.cppreference.com/w/cpp/algorithm/transform). I honestly think we should consider making a ``Lattice`` utility method to assist with this templating concept, but that can be kept for a future PR.

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+  refined_reflector_cell.setFill(&refined_ref_lattice);
+
+  /* Sliced up water cells - right side of geometry */
+  Lattice right_ref_lattice;
+  right_ref_lattice.setWidth(1.26, 1.26);
+  Universe* right_ref_matrix[17*17];
+  for(int i=0; i<17; i++){
+    for(int j=0; j<17; j++){
+      int index =  17*j + i;
+      if(i<11)
```
wbinventor: whitespace between ``if`` and ``(``

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+  for(int n=0; n<10*10; n++)
+    refined_ref_matrix[n] = &reflector;
+  refined_ref_lattice.setUniverses(10, 10, refined_ref_matrix);
+  refined_reflector_cell.setFill(&refined_ref_lattice);
+
+  /* Sliced up water cells - right side of geometry */
+  Lattice right_ref_lattice;
+  right_ref_lattice.setWidth(1.26, 1.26);
+  Universe* right_ref_matrix[17*17];
+  for(int i=0; i<17; i++){
```
wbinventor: whitespace between ``for`` and ``(``

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+    for(int j=0; j<17; j++){
```
wbinventor: whitespace between ``for`` and ``(``

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+                        1, 2, 4, 3, 3, 4, 3, 3, 4, 3, 3, 4, 3, 3, 4, 2, 1,
+                        1, 2, 2, 2, 3, 3, 3, 3, 3, 3, 3, 3, 3, 2, 2, 2, 1,
+                        1, 2, 2, 4, 2, 3, 3, 3, 3, 3, 3, 3, 2, 4, 2, 2, 1,
+                        1, 2, 2, 2, 2, 4, 2, 2, 4, 2, 2, 4, 2, 2, 2, 2, 1,
+                        1, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 1,
+                        1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1};
+
+    std::map<int, Universe*> names = {{1, &mox43}, {2, &mox7}, {3, &mox87},
+                                      {4, &guide_tube}, {5, &fission_chamber}};
+    for(int n=0; n<17*17; n++)
```
wbinventor: whitespace between ``for`` and ``(``

profile/c5g7-cmfd.cpp
wbinventor: whitespace between ``for`` and ``(``

profile/Makefile
```diff
@@ -0,0 +1,170 @@
...
+endif
+ifeq ($(COMPILER), intel)
+  CFLAGS += -openmp
+endif
+ifeq ($(COMPILER), ibm)
+  CFLAGS += -qsmp
+endif
+  CFLAGS += -DOPENMP
+endif
+
```
wbinventor: What about CUDA?

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+    }
+  }
+  corner_ref_lattice.setUniverses(17, 17, corner_ref_matrix);
+  corner_reflector_cell.setFill(&corner_ref_lattice);
+
+  /* Sliced up water cells for bottom of geometry */
+  Lattice bottom_ref_lattice;
+  bottom_ref_lattice.setWidth(1.26, 1.26);
+  Universe* bottom_ref_matrix[17*17];
+  for(int i=0; i<17; i++){
```
wbinventor: whitespace between ``for`` and ``(``

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+    for(int j=0; j<17; j++){
```
wbinventor: whitespace between ``for`` and ``(``

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+      int index = 17*j + i;
+      if(j<11)
```
wbinventor: whitespace between ``if`` and ``(``

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+    }
+  }
+  right_ref_lattice.setUniverses(17, 17, right_ref_matrix);
+  right_reflector_cell.setFill(&right_ref_lattice);
+
+  /* Sliced up water cells for bottom corner of geometry */
+  Lattice corner_ref_lattice;
+  corner_ref_lattice.setWidth(1.26, 1.26);
+  Universe* corner_ref_matrix[17*17];
+  for(int i=0; i<17; i++){
```
wbinventor: whitespace between ``for`` and ``(``

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+    for(int j=0; j<17; j++){
```
wbinventor: whitespace between ``for`` and ``(``

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+      int index = 17*j + i;
+      if(i<11 && j<11)
```
wbinventor: whitespace between ``if`` and ``(``

profile/Makefile
```diff
@@ -0,0 +1,170 @@
...
+ifeq ($(COMPILER),intel)
+  CC = icc
+endif
+
+# IBM XL Compiler
+ifeq ($(COMPILER),ibm)
+  CC = mpicc
+endif
+
+# BG/Q gcc Cross-Compiler
```
wbinventor: Might also consider the [clang](http://clang.llvm.org/) compiler which all the cool kids seem to be flocking to these days.

profile/c5g7-cmfd.cpp
```diff
@@ -0,0 +1,509 @@
...
+    &assembly2,         &assembly1,         &right_reflector,
+    &bottom_reflector,  &bottom_reflector,  &corner_reflector}; 
+  full_geometry.setUniverses(3, 3, universes);
+  root_cell.setFill(&full_geometry);
+  
+  /* Create CMFD mesh */
+  log_printf(NORMAL, "Creating CMFD mesh...");
+
+  Cmfd cmfd;
+  cmfd.setMOCRelaxationFactor(0.6);
```
wbinventor: This breaks when compiling "c5g7-cmfd.cpp" with double precision. I think by simply casting the parameters to ``Cmfd`` as floats you should get around this.

wbinventor:
@geogunow this is a very cool PR! I'm glad you were able to get this new profile-friendly build system up and working. I reviewed this PR through "c5g7-cmfd.cpp". Most of my comments will need to be addressed for each of the new C++ "input files" as well, but I think they are pretty straightforward.

I tested all of the inputs on my machine with all of the compiler options. Everything seems to work for GNU's compiler (save one of my comments above), but I'm unable to compile with Intel. Here is what happens when I run ``make COMPILER=intel``:

```bash
icc -std=c++11 -DFP_PRECISION=float -DSINGLE -I/usr/local/include/python2.7 -I/usr/lib/python2.7/dist-packages/numpy/core/include  -DVEC_LENGTH=8 -DVEC_ALIGNMENT=16 -O3 -xhost -ansi-alias -no-prec-div -DINTEL -openmp -DOPENMP -c c5g7-cmfd.cpp -o c5g7-cmfd.o
c5g7-cmfd.cpp(35): error: a list initializer is not allowed for type "std::array<double, 7UL>"
    sigmaA["Water"] = std::array<double, 7> {6.0105E-4, 1.5793E-5, 3.3716E-4, 
                                            ^

c5g7-cmfd.cpp(40): error: a list initializer is not allowed for type "std::array<double, 49UL>"
        {0.0444777, 0.1134, 7.2347E-4, 3.7499E-6, 5.3184E-8, 0.0, 0.0, 
        ^

c5g7-cmfd.cpp(369): error: expected an expression
      std::map<int, Universe*> names = {{1, &uo2}, {2, &guide_tube}, 
                                        ^

c5g7-cmfd.cpp(369): error: expected an expression
      std::map<int, Universe*> names = {{1, &uo2}, {2, &guide_tube}, 
                                                   ^

c5g7-cmfd.cpp(370): error: expected an expression
                                        {3, &fission_chamber}};
                                        ^

c5g7-cmfd.cpp(401): error: expected an expression
      std::map<int, Universe*> names = {{1, &mox43}, {2, &mox7}, {3, &mox87},
                                        ^

c5g7-cmfd.cpp(401): error: expected an expression
      std::map<int, Universe*> names = {{1, &mox43}, {2, &mox7}, {3, &mox87},
                                                     ^

c5g7-cmfd.cpp(401): error: expected an expression
      std::map<int, Universe*> names = {{1, &mox43}, {2, &mox7}, {3, &mox87},
                                                                 ^

c5g7-cmfd.cpp(402): error: expected an expression
                                        {4, &guide_tube}, {5, &fission_chamber}};
                                        ^

c5g7-cmfd.cpp(402): error: expected an expression
                                        {4, &guide_tube}, {5, &fission_chamber}};
                                                          ^

c5g7-cmfd.cpp(401): error: no instance of constructor "std::map<_Key, _Tp, _Compare, _Alloc>::map [with _Key=int, _Tp=Universe *, _Compare=std::less<int>, _Alloc=std::allocator<std::pair<const int, Universe *>>]" matches the argument list
            argument types are: (<error-type>, <error-type>, <error-type>, <error-type>, <error-type>)
      std::map<int, Universe*> names = {{1, &mox43}, {2, &mox7}, {3, &mox87},
                                       ^

compilation aborted for c5g7-cmfd.cpp (code 2)
make: *** [c5g7-cmfd.o] Error 2
```

This was with icc v13.1.0. Have you tried compiling with Intel? If so, which version did you use? This needs to be dealt with, but otherwise, I think this PR will be ready to go quite soon.

wbinventor:
A few other things come to mind which might improve this PR. First, you might consider making a single "materials.h" file to define all of the materials and to include in each "input" file. Second, you might put the input files in a new "profile/models" subdirectory. Finally, it might be worth adding a "make all" option to build all executables at once.

config.py
geogunow: I was taught to use __WHATEVER__ so that it doesn't conflict with any names you might actually use. But I can change this to SWIG to fit the OpenMOC style.

profile/Makefile
geogunow: I'll just add it to the list of source files to compile. It should probably be re-compiled every time.

geogunow: Actually GNU can't compile VectorizedSolver.cpp

profile/Makefile
geogunow: Yes, I need to take another look at the compilers

profile/Makefile
wbinventor: right, so only add it to the sources when using the intel compiler

profile/c5g7-cmfd.cpp
geogunow: What compiler are you using? I compile and run just fine with GNU on my machine.

wbinventor:
@geogunow all of your changes look really good. The only thing that still seems awry for me is that I'm still unable to compile with Intel due to the same errors I reported earlier. I've tried using ``icpc`` v0.13.1 on nsecluster (which requires changing the "Python.h" include directory to "/usr/local/include/Python2.7" btw). Once that issue is fixed, I think it would be good for @samuelshaner to look everything over and check this out on his Mac. Otherwise, this is looking good to me!

geogunow:
@wbinventor this PR works on my computer and the computers at Argonne. Today I tried compiling on nsecluster and was able to reproduce your errors. The error actually originates from Intel not keeping up with C++11 in some of their compilers. Only ``icpc`` v15.0 has full C++11 capability.  The format I use to create the lattices in C5G7 requires the use of ``std::array`` which is listed as a C++11 data structure so its likely that ``icpc`` v0.13.1 on nsecluster doesn't handle ``std::array`` properly. I am open to other options for creating lattices if you have any. But using ``std::array`` seemed like the best option for readability. 

Also, have you tried using the GNU compiler? I received some strange errors from that one as well because the default ``g++`` compiler on nsecluster is **very** old. A newer compiler is installed that makes it work, but you need to change the makefile to point to the correct directory and also change the LD_LIBRARY_PATH variable. In short, we should have Sterling update nseculster so that the default is the newer compiler which is gcc 4.8 (right now it uses 4.4).

wbinventor:
OK, that's good to know that it was just a C++11 compliance issue. I think it's best to stick to your convention and make full use of those nice features and data structures. Perhaps you can get Sterling to update the default GNU compilers and install the latest Intel compilers on nsecluster.

wbinventor:
One more request for this PR - put all of the executables in a new "profile/bin" directory, and add them to ".gitignore" so they will not accidentally be checked into the repo.

geogunow:
@wbinventor I slightly disagree with your recommendation on the profile/bin directory for the executables due to the OpenMOC structure. When OpenMOC is run, it creates a ``log`` and ``tracks`` folder in which it stores the logs and tracking information. However, the files in those folders do not include any information on which executable they are from. So what could happen is the user could run one of the executables which would create a tracking file for that case. Then the user could try to run a different case that could try to load the previous tracking information and a segmentation fault could occur. In general, I think we might want to address this behavior sometime (the possibility of loading an incorrect tracking file) but in its current form I think it's best to keep the executables separate from one another so that this doesn't happen. Having the executables next to their source "input" file seems reasonable to me.

wbinventor:
I do see your point. The main reason I proposed a "bin" directory was that I realized we need these executables to be in the ".gitignore" file and thought this would be an easy way to do it. Regardless of where the executables are located, you need to figure out a way to make sure Git ignores them.

Considering what you mentioned, it would be easy for the log files to reference the executable that they came from. Simply add a logger message to top of the corresponding model file, something like the following:

```c
log_printf(NORMAL, "Running c5g7-cmfd");
```

Second, I would be surprised if you ran into a segmentation fault when running one executable with a "tracks" directory from a different executable. The reason is that the first piece of each track binary file is a "geometry string" - basically one big string which represents the geometry in use [see [here](https://github.com/mit-crpg/OpenMOC/blob/develop/src/TrackGenerator.cpp#L1123)]. If the "geometry string" in a track binary file does not match the "geometry string" built from the geometry used by the particular executable, the ``TrackGenerator`` will perform ray tracing and override the tracks file with a new one. IF you do indeed see a segmentation fault, then this is an issue since I specifically architected the ``TrackGenerator`` and its binary files to bypass this issue.

All that being said, I do see the value in separating each model executable into its own directory so that track files do not have to be re-generated each time.

geogunow:
I believe I have come across a segmentation fault from loading an incorrect tracking file. I think it was from running a homogeneous cube and varying the lattice discretization (20x20 cells vs 10x10 cells, etc).

wbinventor:
Alright, if you can dig it up post it as an issue so we can fix it. Must be some geometry parameter isn't properly scraped and included in the geometry string.

geogunow:
Current status: Intel and GNU work, but I can't figure out how to get Clang and NVCC to work.

wbinventor:
Since you'll be the main person using this build system (I assume), it's up
to you if you want to make them work. If so, I would make nvcc a priority
over clang.

wbinventor:
I'll take responsibility for this one, but would like for at least one of @samuelshaner or @PrezNattyGibbs to also try this out on their system.

samuelshaner:
@geogunow - The code compiles just fine on a mac with your Makefile, but I get the following malloc error when each input is done executing (after the results are printed):

homogeneous-one-group(72280,0x7fff77b59300) malloc: *** error for object 0x7fff5f10ba50: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6

wbinventor:
I believe all of these input files will now fail due to double deallocation of some of the objects instantiated in each C++ input file. This is related to PR #161. Although I think that our discussion yesterday about destructors remains pertinent @geogunow, after looking at the inputs I think there is another issue at hand. All of the objects (Surfaces, Cells, etc) are statically allocated meaning that the C++ garbage collector will call their destructors when they call out of scope. However, the Geometry will try to delete these object in its drstructor, so one way or another, they will be deallocated twice. 

One way to get around this would be dynamic memory allocation of these objects in your input files. An alternative might be to craft up a way to use smart pointers. Or perhaps you have a different preferred approach.

geogunow:
Yes, maybe there was some confusion but this was the exact issue I was
trying to discuss yesterday. I'm just curious why the structure is like
this. From what I know it's "good style" in C++ to have every call to new
associated with a call to delete. If the Geometry object deletes its
members this isn't really possible.

wbinventor:
Yes, but the memory allocation process was designed to facilitate
parametric studies in Python. Feel free to propose a better approach.

wbinventor:
And I think the issue we discussed yesterday was not this, but the
Universe, Cell and Geometry destructors.

geogunow:
Yes, but it's only an issue because of the Universe and Geometry destructors. Can you send me the python code you are trying to run? I'm not sold that we really need to call delete on external data structures from the Universe and Geometry destructors.

geogunow:
I took a much closer look at this stuff today and have decided the only reason the Python version works is complete voodoo magic. It really should not work. Double deletes should be all over the python version but for some reason they aren't. It's like Python is predicting they are coming and the GC deletes things in the perfect order so that errors aren't encountered.

geogunow:
If anyone is really interested as to why I'm talking like a nut job, re-install OpenMOC with the most recent commit (it contains debugging statements). Then run my C++ branch (just run make and then make run in the profile directory) and then compare with running the regular c5g7.py input with OpenMOC in Python. What happens in the Python version is really strange.

wbinventor:
I do think that there are a few places where items may be deleted twice, and this should be corrected. I'm not sure what you mean about the Python "voodoo" so if you could give an example, perhaps I can help. Some of this "voodoo" may be what you see in [openmoc/thisown.i](https://github.com/mit-crpg/OpenMOC/blob/develop/openmoc/thisown.i) which I introduced in PR #161. This is used to ensure that Python does not doubly delete items after they have been "assigned" (or added) to other objects in a script.

geogunow:
Ok the voodoo is the following: the change that makes the C++ alternative build system crash is that the ``Universe`` destructor attempts to delete all the ``Cell`` objects that it "owns."  The problem in the C++ build system is that when the Universe destructors are called, they make share common ``Cell`` objects. Inherently this would lead to double deletes. And that is indeed the case in the C++ build. A segmentation fault occurs in the destructor. In the most recent commit in this PR I added print statements in the ``Universe`` destructor. And quite clearly we can see that the segmentation fault occurs when the second ``Unverse`` destructor attempts to delete a ``Cell`` object that was deleted by the first ``Universe`` destructor. This makes perfect sense.

However, in the python build, this segmentation fault does not occur. After re-installing OpenMOC with the added print statements, we see that whenever the ``Universe`` destructor is called, **all universes think they own no cells!** 

I don't think that is the behavior that you were expecting, sense this essentially means the ``for`` loops in the ``Universe`` destructor that delete members are completely useless. I have no idea why this occurs.

wbinventor:
Actually, I think this behavior makes perfect sense. The Python voodoo is all a result of the "thisown.i" file. The ``Universe`` destructors are not called from the Python script since the ``Universes`` are added to a ``Lattice``. Take a look at lines [91-107](https://github.com/mit-crpg/OpenMOC/blob/develop/openmoc/thisown.i#L91) of "thisown.i" to see how SWIG/Python is instructed *not* to call a ``Universe`` object instructor if that ``Universe`` is added to a ``Lattice.

However, I do think that you raise a good point that the ``Universe`` destructor is faulty in that it will lead to segmentation faults if a single ``Cell`` is shared by multiple ``Universes`` for the C++-only build. I think the fix for this is to move the ``Cell`` deallocation to the ``Geometry`` destructor. This would add a code block which looks very similar to the one that is already there to destruct ``Materials``. By deleting all objects in the CSG tree at the head node (the ``Geometry`` object) we can avoid these sorts of issues (which I thought I had done in PR #161 but looks like not entirely).

Does this make sense?

wbinventor:
I just opened PR #176 which I believe resolves the outstanding issues with the ``Universe`` destructor noted here. I'll let you review that PR @geogunow, and we can merge it in soon to eliminate the segfaults in this PR.

jtramm:
Using the new C++ only build system, we were able to get SimpleMOC running on BG/Q. There were several changes from the main develop branch that needed to be made:

1) Merging in of this pull request ( #165 )

2) Merging in the fix from PR #174 

3) Fixing the OpenMP atomic directives in the parallel hash map that are too new for the BG/Q gcc (4.4.6) compiler. Namely, the more exotic atomic operations present in src/ParallelHashMap.h:

Line 260:
  #pragma omp atomic update

should be changed to:
#pragma omp atomic

and Lines 298 & 299:
  #pragma omp atomic capture
  N = _N++;

should be change to:
#pragma omp critical (local_hash1)
{
  N = _N++;
}

4) The C++11 compiler flag needed to be changed from:

-std=c++11

to:

-std=gnu++0x

This should probably done only for BG/Q, perhaps with a machine specific variable.

5) This is not a code issue, but for other users looking to compile on BG/Q, the compiler must be gcc (not XL or bgclang). This is set via the .soft file in your BG/Q by making sure that:

+mpiwrapper-gcc

is the only mpiwrapper selected. Then, compilation must be done via mpic++ (NOT g++) to ensure that the proper cross compiler is used (even if you're not actually using MPI, this works fine).


I think that's about it, unless there's anything else I'm forgetting @theshuo ?

-John

theshuo:
Yeah, the Makefile needs to be modified so the proper gcc flags are set for BG/Q. Namely, 

1) -Ofast changed to -O3
2) remove -msse2 flag

geogunow:
The C++ alternative build is now stable for GNU, Intel, Clang, and BG/Q. Currently there is no support for NVIDIA/CUDA but I think that should be added in a later PR once someone is interested in testing the GPU solver.

wbinventor:
Awesome, I think this PR has come a log way and is looking good. I did run
into one thing last night when writing some C++ models for @jtramm and
@theshuo. Re-running "make" for a model which I had changed since the
previous build did not seem to reflect itself in the new executable.
Instead, I had to run "make clean" and then "make" which was a bit
inconvenient. Any idea on how to fix this?

geogunow:
@wbinventor This most recent commit should fix that.

wbinventor:
@geogunow can you add a new section to the Installation page of the User's Guide on this new build system option?

geogunow:
Sure, I'll work on it this weekend.

geogunow:
Will, how do I edit the instillation instructions? I can't seem to find
where to edit it.

wbinventor:
Have you followed the instructions on how to update the documentation in the Developer's Guide [here](https://mit-crpg.github.io/OpenMOC/devguide/documentation.html)? If so, then the installation instructions will be in "sphinx/docs/source/usersguide/install.rst" file (on GitHub [here](https://mit-crpg.github.io/OpenMOC/devguide/documentation.html)).

wbinventor:
Note that we have not been using code reviews for the documentation up to this point. This means that it is your prerogative to fully inspect the documentation that you generate using Sphinx/Doxygen to make sure that is as you expect. If you'd like to enforce code review for the docs in the future, then you'll have to restructure the "build system" for the documentation a bit.

geogunow:
Will, should this be in the developers guide instead? Using a makefile
seems much more up the developers alley.

wbinventor:
Sure, sounds reasonable, perhaps "Advanced Installation for Developers" or something like it. Just make sure to include a link to it from the User's Guide Installation Instructions.

Issue #168, opened 2015-07-11
Error in Loading Tracking File Incorrectly


geogunow:
In the current tracking file dump/loading system, there is a bug where a tracking file is loaded that does not correspond to the case being run. The result is an incorrect error message.

It seems the root of this is not accounting for different lattice discretizations. To showcase this error, I will refer to the one-directional gradient case presented in PR #164. 
⋅⋅ 1. Run one-directional-cmfd.py
⋅⋅ 2. Change line 93 from ``num_cells_x = 100`` to ``num_cells_x = 200``
`` 3. Run one-directional-cmfd.py again

OpenMOC will incorrectly attempt to load the previous tracking file. The result is an error message reading
```
Could not find FSR ID with key: CMFD = (0, 0) : UNIV = 10003 : LAT= 10004 (2, 0) : UNIV = 10002 : CELL = 10000. Try creating geometry with finer track spacing
```
after attempting to plot the flat source regions.

PR #169, opened 2015-07-12
Fix Geometry::toString() method for improved importation of tracking files

wbinventor:
This PR fixes a bug that is the culprit behind Issue #168. The bug related to the way the ``Geometry::toString()`` routine assumed all ``Universes`` in a ``Geometry`` were indeed ``Universes`` and not the ``Lattice`` subclass. As a result, when building a string representation of the ``Geometry``, the ``Universe::toString()`` routine was called in place of the ``Lattice::toString()`` method, leaving out important information such as the ``Lattice`` dimensions and ``Universe`` IDs. Now, the routine calls ``Lattice::toString()`` explicitly when necessary, embedding the relevant geometric information needed to avoid perturbed geometries such as the examples discussed in #168.

### DIFF
--- a/src/Geometry.cpp
+++ b/src/Geometry.cpp
@@ -1005,7 +1005,7 @@ std::string Geometry::toString() {
   string << "\n\tUniverses:\n\t\t";
   for (univ_iter = all_universes.begin();
        univ_iter != all_universes.end(); ++univ_iter)
-      string << univ_iter->second->toString() << "\n\t\t";
+    string << univ_iter->second->toString() << "\n\t\t";
 
   std::string formatted_string = string.str();
   formatted_string.erase(formatted_string.end()-3);
--- a/src/Geometry.cpp
+++ b/src/Geometry.cpp
@@ -1004,16 +1004,8 @@ std::string Geometry::toString() {
 
   string << "\n\tUniverses:\n\t\t";
   for (univ_iter = all_universes.begin();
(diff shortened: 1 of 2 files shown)


PrezNattyGibbs:
This is a straightforward change, and if it works right (I'll let @geogunow confirm it works for what he was seeing!), I'm happy with it.

But isn't it a bit goofy that C++ isn't calling the right routine here?  This seems like a really clunky way to have to do it.  I guess it just sticks with whatever the type of the pointer is, but I'd think you'd want the underlying class type to shine through in general.  Whatever, I'm not a C++ compiler or standards author...

geogunow:
Everything checks out and seems to work. But I agree with @PrezNattyGibbs that the fix seems a bit goofy. Maybe a cleaner fix would be to have separate for loops for Universes and Lattices?

wbinventor:
You two are absolutely right, I didn't like this fix myself. I didn't check
until now but the reason the subclass method wasn't being called is that
the "Universe::toString()" method is not "virtual". I'll change this when
I'm near a computer later and everything will be made OO again.

wbinventor:
Alright, I've simplified this PR to a single line adjustment to make ``Universe::toString()`` a ``virtual`` method. Double check me to make sure it does the trick for #168.

geogunow:
I just tried it on a variety of cases and it looks good. I think this is ready to merge.

wbinventor:
Pull the trigger!

PR #172, opened 2015-07-15
moved call for CMFD initialization from Geometry to TrackGenerator

samuelshaner:
This PR address issue 170 by moving the call to initialize the CMFD object from the `Geometry::initializeFlatSourceRegions()` method to `TrackGenerator::generateTracks()` method so the CMFD object can be generated just prior to generation of the tracks.

### DIFF
--- a/src/Geometry.cpp
+++ b/src/Geometry.cpp
@@ -760,10 +760,6 @@ void Geometry::initializeFlatSourceRegions() {
 
   /* Create map of Material IDs to Material pointers */
   _all_materials = getAllMaterials();
-
-  /* Initialize CMFD */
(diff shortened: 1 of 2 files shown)


geogunow:
Looks good to me. Do we need to change our input examples at all to account for it? Right now I believe initializeFlatSourceRegions is called directly in the sample inputs.

wbinventor:
Looks good to me. I think we still need to call "initializeFSRs()" in the
inputs since it is used to subdivide cells into rings/sectors. But perhaps
that can be buried into the generateTracks() routine as well. I'll let
@samuelshaner comment on that.

PR #173, opened 2015-07-15
Recompute cell min/max values and boundary conditions when queried

samuelshaner:
This PR addresses issue 171 by recomputing the Cell's min/max values and boundary conditions when these values are queried. To avoid code reuse and because these evaluations are often quick, the `Cell::findBoundingBox()` method is called during each call to `Cell::getMin_()`, `Cell::getMax_()`, and `Cell::get____BoundaryType()` and all bounding box properties are updated instead of just the value being queried. This PR builds on PR #172 and should merged after that one.

### DIFF
--- a/src/Cell.cpp
+++ b/src/Cell.cpp
@@ -161,6 +161,7 @@ int Cell::getNumSectors() {
  * @return the minimum x-coordinate
  */
 double Cell::getMinX() {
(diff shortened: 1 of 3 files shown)


geogunow:
Looks good

wbinventor:
Also looks ready to me, pending #172.

PR #174, opened 2015-07-16
Hotfix for Intel compiler

wbinventor:
This PR fixes a bug which was preventing me from building the code using the Intel compiler (v0.13.1 on nsecluster). This was introduced by me in d35e0ca16b66c6f89a7d83550f4b6b4d01df4b33 in PR #161 in the ``Universe::removeUniverse(...)`` routine. Although this routine compiled just fine with GNU, the Intel compiler can be more strict sometimes and it caused it to fail.

### DIFF
--- a/src/Universe.cpp
+++ b/src/Universe.cpp
@@ -1080,12 +1080,14 @@ void Lattice::setUniverses(int num_y, int num_x, Universe** universes) {
  */
 void Lattice::removeUniverse(Universe* universe) {
 
+  Universe* null = NULL;
+
   /* Set all locations in the array of universes array to NULL */
   for (int j=0; j < _num_y; j++) {
     for (int i = 0; i < _num_x; i++) {
       if (universe->getId() == getUniverse(i,j)->getId())
         _universes.at(j).push_back(std::pair<int, Universe*>
-                                 (-1, NULL));
+                                 (-1, null));
     }
   }
 }
--- a/src/Universe.cpp
+++ b/src/Universe.cpp
@@ -1080,8 +1080,6 @@ void Lattice::setUniverses(int num_y, int num_x, Universe** universes) {
  */
 void Lattice::removeUniverse(Universe* universe) {
 
-  Universe* null = NULL;
-
   /* Set all locations in the array of universes array to NULL */
   for (int j=0; j < _num_y; j++) {
     for (int i = 0; i < _num_x; i++) {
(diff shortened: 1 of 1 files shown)


PrezNattyGibbs:
So `NULL` pointers have to be typed?  Goofy compilers!

Looks fine to me.

src/Universe.cpp
```diff
@@ -1080,12 +1080,14 @@ void Lattice::setUniverses(int num_y, int num_x, Universe** universes) {
  */
 void Lattice::removeUniverse(Universe* universe) {
 
+  Universe* null = NULL;
+
   /* Set all locations in the array of universes array to NULL */
   for (int j=0; j < _num_y; j++) {
     for (int i = 0; i < _num_x; i++) {
       if (universe->getId() == getUniverse(i,j)->getId())
         _universes.at(j).push_back(std::pair<int, Universe*>
```
PrezNattyGibbs: While the change looks fine and won't change the existing routine, I don't understand what this routine does.  If some `Universe` in a `Lattice` matches the given one, we add a `NULL` pointer to a vector of X `Universes` at the corresponding Y coordinate?  How does that remove a `Universe`?

PrezNattyGibbs: Shouldn't we be replacing the `Universe` with the `NULL` pointer?

So something like `_universe.at(j)[i] = std::pair<int,Universe*>(-1,null)`?  Or am I missing something?

wbinventor: Good call, I agree. Will fix soon.

wbinventor: I've updated to use your recommendation @PrezNattyGibbs.

PrezNattyGibbs:
I bet you thought this PR was supposed to be simple!

I tried to run the latest version, removing a universe from the lattice, and then calling `log.py_printf('ERROR', 'Stop here.')` to prevent the rest of the input file from running.  And when I do that, `Lattice::~Lattice()` throws a seg fault because the call `std::map<int, Universe*> unique_universes = getUniqueUniverses();` calls `Universe::getId()` for the `NULL` universes.

wbinventor:
Just update this to deal with the issue you pointed out last week @PrezNattyGibbs. Let me know if you see anything else that could be improved.

geogunow:
Looks reasonable to me. I'll merge tonight unless someone speaks up before then.

PR #176, opened 2015-07-21
Improved Geometry and Universe destructors

wbinventor:
This PR is intended to address concerns raised by @geogunow in PR #165 over double deletions of ``Cells`` in the ``Universe`` destructors. Now the ``Geometry`` deletes all ``Materials`` followed by all ``Cells`` followed by ``Universes``. This should work for both the Python and C++-only build systems (as verified when merging into @geogunow's "profile" branch).


geogunow:
I think this is good to go, merging now

PR #179, opened 2015-07-22
Fixed valgrind issues

wbinventor:
This PR addresses a few inconsistencies in the way memory is allocated/deallocated with free/delete per an email conversation with @cjosey and his valgrind runs with OpenMOC.

### DIFF
--- a/src/CPUSolver.cpp
+++ b/src/CPUSolver.cpp
@@ -190,15 +190,18 @@ void CPUSolver::zeroTrackFluxes() {
 
 
 /**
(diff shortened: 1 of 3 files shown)


cjosey:
I tested this and found no issues.  I spoke with @geogunow about the remaining error in the parallel hashmap, and he indicated that the warning that appears during the announce check is fine, as those uninitialized values are used as false sharing padding for performance reasons.

wbinventor:
Feel free to merge this in @cjosey

PR #162, opened 2015-06-30
Secondary eigenvalue capabilities using Krylov subspace methods

cjosey:
As per Will's request.

This modification allows the expression of the M and A operators of the  transport equation Ax = 1/k Mx as independent transport sweeps.  The method currently in OpenMOC represents a single operator on which a non-linear power iteration method is performed.

As separated operators, each can be cast as a SciPy LinearOperator and sparse matrix math can be performed.  In the example c5g7-krylov.py, I use a combination of GMRES and eigs (an IRAM implementation) to then solve for the first 6 eigenvalues and eigenvectors of the problem.

This method has a few caveats:
* The method is not yet derived for any boundary condition other than vacuum.
* It requires 2k+1 eigenvector-sized arrays to be stored, where k is the number of eigenvectors required.
* Eigs does not output convergence information.
* CMFD not only does not accelerate the method, but actively prevents it from working.
* It is slower than unaccelerated non-linear power iteration for a single eigenvalue
* GMRES must converge more tightly than IRAM, but I do not know by precisely how much.
* I implemented it assuming double precision.  It can probably be rewritten to read the FP_PRECISION type and adjust accordingly.
* Implementation was only done for CPUSolver

Issue #117, opened 2015-01-20
Doxygen errors for CMFD class


wbinventor:
There are quite a few errors/warnings when running Doxygen on the CMFD class.

samuelshaner:
I'll get on this.

wbinventor:
Closing this with #247.

PR #182, opened 2015-07-24
CMFD clean up by adding surface constants and Matrix/Vector linear algebra classes

samuelshaner:
This PR cleans up the CMFD class through the following modifications:
- Adding named constants to represent the surfaces of a rectangle helps to clarify the IDs used to represent the surfaces and corners of a rectangle.
- `Matrix` and `Vector` classes are now used for the CMFD production/loss matrices and flux/source vectors. The matrix object is organized as a list-of-lists and converted to compressed sparse row format for efficient matrix-vector operations. Along with these additions, the linear algebra functions have been cleaned up and made generic, which will help for things down the road like extending to 3D or computing the adjoint via transposing the matrix.

This PR builds on the CMFD k-nearest update PR and should thus be merged after that one.


wbinventor:
I can take charge of reviewing this over the course of the week, though a separate set of eyes from @geogunow and/or @PrezNattyGibbs wouldn't be a bad idea either. Fortunately, the diff for this has decreased substantially with the inclusion of #145.

geogunow:
Some of my comments on this I mistakenly put on the boundary condition PR. But in general, I would like to see a bit more of cool C++ features being used, such as operator overloading for matrix multiplication.

samuelshaner:
Thanks for the comments @geogunow! I have thought about trying to abstract away some of the long if/else if sections, but have found it difficult. I'll think about the abstraction a bit more and look into overloading the linear algebra operators and let you guys know when I have an update.

geogunow:
@samuelshaner if you'd like to see some examples of operator overloading, I
made my own matrix class for 22.213. You can find the source code here:
https://github.com/geogunow/22.213/blob/master/rksolver/src/Sparse.cpp. It
might be a good reference to see how to start things off.

samuelshaner:
I merged `cmfd-surfaces` with the latest version of `develop`. While I completely agree that the abstractions to reduce code reuse and support operator overloading for linear algebra functions should be implemented, I would personally vote to put that up as an issue and save that for a future PR so we can get the periodic boundary conditions and modifications to support 3D geometries PRs merged. Looking over the comments on my other PRs, I couldn't find any issues to address in this PR so this PR should be up to date with your comments and is ready for further review or to be merged.

geogunow:
Update this PR with develop and then it should be ready to be merged

samuelshaner:
@geogunow - I merged with develop so this PR should be ready to merge.

openmoc/openmoc.i
```diff
@@ -64,8 +67,11 @@
 %ignore setFSRsToMaterialIDs(std::vector<int>* FSRs_to_material_IDs);
 %ignore setFSRKeysMap(ParallelHashMap<std::size_t, fsr_data*>* FSR_keys_map);
 %ignore initializeFSRVectors();
-
-
+%ignore getArray();
```
wbinventor: Perhaps add a comment here as to why SWIG is being instructed to ignore these methods?

src/Cmfd.cpp
```diff
@@ -139,7 +107,7 @@ void Cmfd::setNumY(int num_y) {
 
 /**
  * @brief Get the number of Mesh cells in a row.
- * @return number of Mesh cells in a row
+ * @return The number of Mesh cells in a row
```
wbinventor: Two questions - first, are a lot of these changes to the Doxygen comments to address #117 or just general cleanup? Second, it looks like you capitalize the first word in the parameter comments in many but not all of the Doxygen comment blocks. At this point, I tend to think that capitalization is best for the documentation, though this is not consistent throughout the code. I don't see this as an issue at all for this PR, but it might be something that we should come to an agreement upon and add to the Style Guide to be consistent in the future.

src/Cmfd.cpp
```diff
@@ -181,6 +149,13 @@ void Cmfd::setHeight(double height) {
  * @brief Create cross-sections and fluxes for each Cmfd cell by
  *        energy condensing and volume averaging cross sections from
  *        the MOC sweep.
+ * @details This method performs a cell-wise energy condensation and volume
+ *         average of the cross sections of the fine, unstructured FSR mesh.
+ *         The cross sections are condense such that all reaction rates and
```
wbinventor: typo: "condense" -> "condensed"

src/Cmfd.cpp
```diff
@@ -181,6 +149,13 @@ void Cmfd::setHeight(double height) {
...
+ *         the neutron produciton rate from fission are conserved. It is
```
wbinventor: typo: "produciton" -> "production"

src/Cmfd.cpp
```diff
@@ -368,21 +345,21 @@ void Cmfd::computeDs(int moc_iteration) {
...
           /* Set halfspace sense of the Surface */
-          if (surface == 0 || surface == 1)
+          if (surface == SURFACE_X_MIN || surface == SURFACE_Y_MIN)
             sense = -1.0;
           else
             sense = 1.0;
 
           /* Set the length of this Surface and the perpendicular Surface */
-          if (surface == 0 || surface== 2) {
+          if (surface == SURFACE_X_MIN || surface== SURFACE_X_MAX) {
```
wbinventor: I don't see the point of this ``else-if`` block. It looks like irregardless of the value of ``surface``, the ``length`` and ``length_perpen`` variables will be set to the same value. Am I missing something here @samuelshaner?

wbinventor: And if I am missing something, an alternative approach which might make this and some other conditional blocks shorter and simpler might be to split up the ``SURFACE_X/Y_MIN/MAX`` enumerated types into two - ``SURFACE_X``/``SURFACE_Y`` and ``SURFACE_MIN`` and ``SURFACE_MAX``,  or something similar.

src/Cmfd.cpp
```diff
@@ -418,26 +395,26 @@ void Cmfd::computeDs(int moc_iteration) {
           else{
 
             /* Set properties for cell next to Surface */
-            if (surface == 0) {
+            if (surface == SURFACE_X_MIN) {
```
wbinventor: Again, it appears that irregardless of the value of ``surface``, the ``next_length_perpen`` variable is set to ``_cell_height``.

src/Cmfd.cpp
```diff
@@ -447,10 +424,10 @@ void Cmfd::computeDs(int moc_iteration) {
                     * d * f + next_length_perpen * d_next*f_next);
 
             /* Compute net current */
-            current = sense * _surface_currents[cell*_num_cmfd_groups*8 +
-                      surface*_num_cmfd_groups + e] - sense
-                      * _surface_currents[cell_next*_num_cmfd_groups*8 +
-                      next_surface*_num_cmfd_groups + e];
+            current = sense * _surface_currents->getValue
```
wbinventor: Personally, I find lines like these easier to comprehend when they are broken up into multiple lines, even at the expense of a few new ("temporary") variables.

src/Cmfd.cpp
```diff
@@ -517,9 +494,15 @@ void Cmfd::computeDs(int moc_iteration) {
...
- *  @param MOC iteration number
+/**
+ * @brief Solve the nonlinear diffusion acceleration problem to accelerate the
+ *        convergence of the MOC problem.
+ * @details This method uses the information from the last MOC transport sweep
+ *         and solves a simplified nonlinear diffusion problem. The diffusion
+ *         problem is tightly converged and the solution is used to update the
+ *         the solution of the MOC problem.
+ *  @param moc_iteration MOC iteration number
+ *  @return The dominant eigenvalue of the nonlinear diffusion problem
```
wbinventor: Nice, this routine is definitely looking a lot cleaner now!

src/Cmfd.cpp
```diff
@@ -606,311 +539,140 @@ FP_PRECISION Cmfd::computeKeff(int moc_iteration) {
...
+ *         approipriate positions in the loss + streaming matrix and
+ *         fission gain matrix.
  */
 void Cmfd::constructMatrices() {
 
   log_printf(INFO,"Constructing matrices...");
     
   FP_PRECISION value, volume;
-  int cell, row;
+  int cell;
```
wbinventor: should ``cell`` really be ``cell_id``?

src/Cmfd.cpp
```diff
@@ -606,311 +539,140 @@ FP_PRECISION Cmfd::computeKeff(int moc_iteration) {
...
         /* Set transport term on diagonal */
-        value = (material->getDifHat()[e]
-                + material->getDifTilde()[e])
-            * _cell_height;
-        
-
-        _A[cell][e*(_num_cmfd_groups+4)+e+2] += value;
+        value = (material->getDifHat()[SURFACE_X_MAX*_num_cmfd_groups + e]
+                - material->getDifTilde()[SURFACE_X_MAX*_num_cmfd_groups + e])
+          * _cell_height;
```
wbinventor: Some of these continuation lines have inconsistent indentation which makes them tough to parse. I'm not sure if we've adopted @geogunow's proposal to use 2 tabs per line yet (doesn't look like it in the Style Guide). My personal opinion is that whatever we adopt can be applied to the code in future PRs, but that current PRs should at least be consistent with themselves.

src/Cmfd.cpp
```diff
@@ -606,311 +539,140 @@ FP_PRECISION Cmfd::computeKeff(int moc_iteration) {
...
-
-        for (int i = 0; i < _num_cmfd_groups+4; i++)
-          log_printf(DEBUG, "i: %d, A value: %f",
-                     i, _A[cell][e*(_num_cmfd_groups+4)+i]);
-
-        for (int i = 0; i < _num_cmfd_groups; i++)
-          log_printf(DEBUG, "i: %d, M value: %f",
-                     i, _M[cell][e*(_num_cmfd_groups)+i]);
       }
     }
```
wbinventor: These parentheses are not indented a multiple of 2 tabs each.

src/Cmfd.cpp
```diff
@@ -1026,8 +791,9 @@ void Cmfd::setFSRFluxes(FP_PRECISION* scalar_flux) {
 
 
 /**
- * @brief Set successive over-relaxation relaxation factor.
- * @param over-relaxation factor
+ * @brief Set the successive over-relaxation relaxation factor for the
```
wbinventor: is "over-relaxation relaxation" the proper term?

src/Cmfd.cpp
```diff
@@ -1167,21 +900,11 @@ void Cmfd::initializeSurfaceCurrents() {
...
-  _surface_currents = new FP_PRECISION[size];
-
-  /* Allocate memory for OpenMP locks for each Cmfd Mesh surface */ 
-  _surface_locks = new omp_lock_t[num_mesh_cells * 8];
-
-  /* Loop over all mesh cell surfaces to initialize OpenMP locks */
-  #pragma omp parallel for schedule(guided)
-  for (int r=0; r < num_mesh_cells * 8; r++)
-    omp_init_lock(&_surface_locks[r]);
+  int size = num_mesh_cells * _num_cmfd_groups * NUM_SURFACES;
```
wbinventor: What is ``size`` used here for?

src/Cmfd.cpp
```diff
@@ -1333,217 +1056,104 @@ void Cmfd::setNumFSRs(int num_fsrs) {
...
 
 /** @brief Split the currents of the Mesh cell corners to the nearby surfaces.
- * @details left bottom corner -> bottom surface and left surface
- *          of mesh cell below; right bottom corner -> bottom surface
- *          and right surface of mesh cell below; right top corner ->
- *          right surface and top surface of mesh cell to the right;
- *          left top corner -> left surface and top surface of mesh
- *          cell to the left. The currents tallied on a corner is split
- *          equally to the adjoining surfaces.
+ * @details This metohd splits the current from each corner evenly to the two
```
wbinventor: typo: "metohd" -> "method"

src/Cmfd.cpp
```diff
@@ -1333,217 +1056,104 @@ void Cmfd::setNumFSRs(int num_fsrs) {
...
-              _surface_currents[(y*_num_x+x)*ncg*8 + 6*ncg + e];
-          _surface_currents[(y*_num_x+x)*ncg*8 + 3*ncg + e] += 
-              0.5 * _surface_currents[(y*_num_x+x)*ncg*8 + 6*ncg + e];
-          _surface_currents[(y*_num_x+x+1)*ncg*8 + 3*ncg + e] += 
-              0.5 * _surface_currents[(y*_num_x+x)*ncg*8 + 6*ncg + e];
+  #pragma omp parallel for private(current)
+  for (int y = 0; y < _num_y; y++) {
+    for (int x = 0; x < _num_x; x++) {
+      int cell = y*_num_x + x;
+      for (int s = 4; s < NUM_SURFACES; s++) {
```
wbinventor: Would this loop not be more intuitive if it looped from 1 to a new ``NUM_CORNERS`` constant?

wbinventor: Personally, I don't like lumping corners in with surfaces - if that is indeed what is being done here - since it's not at all intuitive to the layperson.

src/Cmfd.cpp
```diff
@@ -1872,71 +1476,71 @@ void Cmfd::generateKNearestStencils(){
...
+FP_PRECISION Cmfd::getUpdateRatio(int cell, int group, int fsr) {
 
   FP_PRECISION ratio = 0.0;
   FP_PRECISION total_distance;
   std::vector< std::pair<int, FP_PRECISION> >::iterator iter;
 
-  if (_centroid_update_on){
+  if (_centroid_update_on) {
 
     total_distance = _centroid_total_distances[fsr];
```
wbinventor: Instead of storing the distances, have you considered storing ``iter->second/total_distance`` to avoid lots of expensive, redundant divisions? Or maybe even better yet for code simplicity, store (1.0 - iter->second/total_distance). This might even cut into the 15% slowdown in CMFD with *k*-nearest neighbor stencils

src/Cmfd.cpp
```diff
@@ -2153,44 +1748,38 @@ void Cmfd::tallySurfaceCurrent(segment* curr_segment, FP_PRECISION* track_flux,
...
 
-      for (int p=0; p < _num_polar; p++)
-        surf_current += track_flux(p,e) * polar_weights[p];
+      int g = getCmfdGroup(e);
 
-      /* Atomically increment the Cmfd Mesh surface current from the
-       * temporary array using mutual exclusion locks */
-      omp_set_lock(&_surface_locks[surf_id]);
+      for (int p=0; p < _num_polar; p++)
+        surf_current += track_flux(p,e) * polar_weights[p] / 2.;
```
wbinventor: Divide by 2 after the loop!

src/Cmfd.cpp
```diff
@@ -2153,44 +1748,38 @@ void Cmfd::tallySurfaceCurrent(segment* curr_segment, FP_PRECISION* track_flux,
...
 
+      int g = getCmfdGroup(e);
+      
       for (int p=0; p < _num_polar; p++)
-        surf_current += track_flux(p,e) * polar_weights[p];
-
-      /* Atomically increment the Cmfd Mesh surface current from the
-       * temporary array using mutual exclusion locks */
-      omp_set_lock(&_surface_locks[surf_id]);
+        surf_current += track_flux(p,e) * polar_weights[p] / 2.;
```
wbinventor: ditto

src/Matrix.cpp
```diff
@@ -0,0 +1,465 @@
...
+ * @param group_from The origin group.
+ * @param cell_to The destination cell.
+ * @param group_from The destination group.
+ * @param val The value used to increment the row/column location.
+ */
+void Matrix::incrementValue(int cell_from, int group_from,
+                            int cell_to, int group_to, FP_PRECISION val) {
+
+  if (cell_from >= _num_x*_num_y || cell_from < 0)
+    log_printf(ERROR, "Unable to increment Matrix value for cell_from %i"
```
wbinventor: Change ``%i`` to ``%d``

src/Matrix.cpp
```diff
@@ -0,0 +1,465 @@
...
+/**
+ * @brief Convert the matrix lists of lists to compressed row (CSR) storage 
+ *        form.
+ */
+void Matrix::convertToCSR() {
+  
+  /* Get number of nonzero values */
+  int NNZ = getNNZ();
+  
+  /* Allocate memory for arrays */
```
wbinventor: this should actually be something like "Deallocate memory for arrays if previously allocated" or something similar

src/Matrix.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+
+/**
+ * @brief Set the number of cells in the x dimension.
+ * @param num_x The number of cells in the x dimension.
+ */
+void Matrix::setNumX(int num_x) {
+
+  if (num_x < 1)
+    log_printf(ERROR, "Unable to set Matrix num x to non-positive value %i",
```
wbinventor: ``%i`` -> ``%d`` here and elsewhere

src/Vector.cpp
```diff
@@ -0,0 +1,268 @@
+#include "Vector.h"
+
+/**
+ * @brief Constructor initializes Vector object as a floating point array
+ *        and sets the vector dimensions.
+ * @detail The vector is ordered by cell (as opposed to by group) on the
+ *         outside to be consistent with the Matrix object . Locks are used to
```
wbinventor: typo with extra whitespace in comment

src/Vector.cpp
```diff
@@ -0,0 +1,268 @@
+#include "Vector.h"
+
+/**
+ * @brief Constructor initializes Vector object as a floating point array
+ *        and sets the vector dimensions.
+ * @detail The vector is ordered by cell (as opposed to by group) on the
+ *         outside to be consistent with the Matrix object . Locks are used to 
+ *         make the vector object thread-safe against concurrent writes the 
+ *          same value. One lock locks out multiple rows of
```
wbinventor: extra space

src/Vector.cpp
```diff
@@ -0,0 +1,268 @@
...
+ *         the vector at a time reprsenting multiple groups in the same cell.
```
wbinventor: typo: "reprsenting" -> "representing"

src/Vector.cpp
```diff
@@ -0,0 +1,268 @@
...
+ *         row and column in the vector. If a value exists for the row,
+ *         the value is incremented by val; otherwise, it is set to val.
+ * @param cell The cell location.
+ * @param group The group location.
+ * @param val The value used to increment the row location.
+ */
+void Vector::incrementValue(int cell, int group, FP_PRECISION val) {
+
+  if (cell >= _num_x*_num_y || cell < 0)
+    log_printf(ERROR, "Unable to increment Vector value for cell %i"
```
wbinventor: ditto ``%i`` -> ``%d``

src/constants.h
```diff
@@ -43,6 +43,16 @@
  *  was selected based on analysis by Yamamoto's 2004 paper on the topic. */
 #define EXP_PRECISION FP_PRECISION(1E-5)
 
+/** The surfaces of a rectangle used in indexing CMFD data arrays. */
+#define NUM_SURFACES 8
```
wbinventor: ditto my comment about splitting up surface and corner constants

src/linalg.cpp
```diff
@@ -0,0 +1,319 @@
...
+ * @param M the fission gain Matrix object
+ * @param X the flux Vector object
+ * @param tol the power method and linear solve source convergence threshold
+ * @param SOR_factor the successive over-relaxation factor
+ * @return k_eff the dominant eigenvalue
+ */
+FP_PRECISION eigenvalueSolve(Matrix* A, Matrix* M, Vector* X, FP_PRECISION tol,
+                             FP_PRECISION SOR_factor) {
+
+  log_printf(INFO, "computing eigenvalue...");
```
wbinventor: Make this log msg more explicit and informative - something brief but it should be clear that this is not coming from the MOC solver

src/linalg.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+  /* Check for consistency of matrix and vector dimensions */
+  if (A->getNumX() != M->getNumX() || A->getNumX() != X->getNumX())
```
wbinventor: ditto ``%i`` -> ``%d``

src/linalg.cpp
```diff
@@ -0,0 +1,319 @@
+#include "linalg.h"
+
+/**
+ * @brief Solves a generalized eigenvalue problem using the Power method.
+ * @details This function takes in a loss + streaming Matrix (A),
+ *          a fission gain Matrix (M), a flux Vector (X), a tolerance used
+ *          for both the power method and linear solve convergence (tol), and
+ *          a successive over-relaxation factor (SOR_factor) and computes the
+ *          dominant eigenvalue and eigenvector using the Power method. The
+ *          eigenvalue is returned along and the input X Vector is modified in
```
wbinventor: typo: "returned along" -> "returned"

src/linalg.cpp
```diff
@@ -0,0 +1,319 @@
...
+  FP_PRECISION residual, _k_eff;
+  int iter;
+
+  /* Compute and normalize the initial source */
+  matrixMultiplication(M, X, &old_source);
+  old_source.scaleByValue(num_rows / old_source.getSum());
+  X->scaleByValue(num_rows / old_source.getSum());
+
+  /* Power iteration diffusion solver */
+  for (iter = 0; iter < 25000; iter++) {
```
wbinventor: Formalize the arbitrary "25000" as a constant perhaps?

src/linalg.cpp
```diff
@@ -0,0 +1,319 @@
...
+    old_source.scaleByValue(_k_eff);
+
+    /* Compute the residual */
+    residual = computeRMSE(&new_source, &old_source, true);
+
+    /* Normalize the new source to have an average value of 1.0 */
+    new_source.scaleByValue(num_rows / new_source.getSum());
+    new_source.copyTo(&old_source);
+
+    log_printf(INFO, "CMFD iter: %i, keff: %f, error: %g",
```
wbinventor: change "error" to "residual" in the log message. And change ``%g`` -> ``%d``?

src/linalg.cpp
```diff
@@ -0,0 +1,319 @@
...
+  FP_PRECISION* x = X->getArray();
+  FP_PRECISION* b = B->getArray();
+  int row, col;
+  Vector old_source(num_x, num_y, num_groups);
+  Vector new_source(num_x, num_y, num_groups);
+
+  /* Compute initial source */
+  matrixMultiplication(M, X, &old_source);
+
+  while (iter < 1000) {
```
wbinventor: make the arbitrary "1000" a constant perhaps?

src/linalg.cpp
```diff
@@ -0,0 +1,319 @@
...
+    /* Compute the residual */
+    residual = computeRMSE(&new_source, &old_source, true);
+
+    /* Copy the new source to the old source */
+    new_source.copyTo(&old_source);
+
+    /* Increment the interations counter */
+    iter++;
+
+    log_printf(INFO, "SOR iter: %i, res: %g", iter, residual);
```
wbinventor: ditto ``%g`` -> ``%f``

src/linalg.cpp
```diff
@@ -0,0 +1,319 @@
...
+      b[row] += a[i] * x[JA[i]];
+  }
+}
+
+
+/**
+ * @brief Computes the Root Mean Square Error of two Vectors.
+ * @details This function takes in two vectors (X and Y) and computes the
+ *          Root Mean Square Error of the Vector Y with respect to Vector X.
+ *          A boolean (integrated) must also be given to indicate whether the
```
wbinventor: What does "A boolean (integrated) ..." mean?

wbinventor: Oops, that comment was premature, I know what you mean now. I think that "The boolean integrated" might be clearer, but it is up to you.

wbinventor:
I've taken a fairly thorough look over this PR and it is a major step forward to cleaning up and generalizing CMFD. Although I agree with @geogunow that I think there is a lot of room for improvement here with built-in C++ types and the use of templates, I agree with you @samuelshaner that this should be raised as an issue for future development (add it @geogunow if you want to keep a track record of your wish list). 

Other than the comments which I have made, the one prerequisite to merging this in should be a thorough check with valgrind that this does not introduce any memory leaks. A few weeks ago @cjosey did some valgrind runs with OpenMOC which we addressed in PR #162. On a related note, it would also be to add in a section to the Developer's Guide on how to use run Valgrind on OpenMOC, something akin to the section on GDB ([here](https://mit-crpg.github.io/OpenMOC/devguide/debugging.html)) perhaps for future developers to reference.

src/Cmfd.cpp
samuelshaner: Cmfd requires that the Cmfd lattice be a regular grid. In other words, the cell width in a particular direction must be the same for all cells, but the cell width is not necessarily the same for all directions. For most CMFD cases (e.g. c5g7-cmfd.py) `length` and `length_perpen` will be the same irregardless of the value of `surface`. However, it the Cmfd lattice in `c5g7-cmfd.py` was changed from 51x51 to 51x3, `length` and `length_perpen` would be depend on whether `surface` is an x surface or a y surface.

The challenge with changing the surface enums from being unique for each surface to having two enums to identify the surface direction (x or y) and side (min or max) is that corners would require two directions and two sides. Also, this would increase the storage requirement for each segment to at least two enums. I only see one conditional block with an `X_MIN' or 'X_MAX' condition and would side on keeping the constants as is.

src/Cmfd.cpp
samuelshaner: Same as above; also, if `surface` is equal to `SURFACE_X_MIN` or `SURFACE_X_MAX`, `next_length_perpen` will be set to the `_cell_width` variable.

src/Cmfd.cpp
samuelshaner: Sure, sounds good.

src/Cmfd.cpp
samuelshaner: Yes, I'll update.

src/Cmfd.cpp
samuelshaner: I was just using the default emacs indent for continuation lines, which seems to indent statements within parentheses to the start of the parentheses and indent statements not in parentheses by two spaces (one tab). As far as I can see, the Google C++ style guide is ambiguous on line continuation. On continuation of function calls, the Google C++ style guide suggests to indent statements within parentheses to the start of the parentheses or by four spaces (two tabs).

I'm fine with any format that looks readable. The current code base indents function calls to the first argument. The majority of line continuations seem to be from code that I pushed, so they are indented in the same format as shown in the statement above. Personally, I like the line continuations within parentheses to be indented to the start of the parentheses. As for line continuations not within parentheses, I'm indifferent to indenting by one or two tabs.

What would you recommend for this PR? It seems like @geogunow's proposal is to use 2 tabs regardless of whether the continuation is within parentheses - is that correct?

src/Cmfd.cpp
samuelshaner: nope, one too many relaxations.

src/Cmfd.cpp
samuelshaner: Nothing anymore; I'll remove.

src/Cmfd.cpp
samuelshaner: All `log_printf` statements in the current branch of develop indent line continuations to the start to this location so I'd recommend we change all `log_printf` statements at the same time (in a future PR) to remain consistent.

src/Cmfd.cpp
samuelshaner: As it stands now, the enum values that represent the corners need to start at 4 and go up from there in order to correctly index into the `_surface_currents` vector. One option could be to add a `_corner_currents` vector and have each segment store a cmfd surface for each direction and a cmfd corner for each direction. This would separate the surface currents and corner currents and allow this loop to be more intuitive. What do you think about that?

src/Cmfd.cpp
samuelshaner: I'll change the second value in the pair to `(1.0 - cell_distance / total_distance)`.

src/Cmfd.cpp
wbinventor: That sounds good.

src/Cmfd.cpp
geogunow: @samuelshaner that is indeed what I'm proposing since that's how my vim is setup. But I am open to changing it, as long as I can edit my vimrc to adopt the style automatically (as it seems your emacs configuration does). I would just like to see a consistent style. I haven't put anything in the style guide yet about it because we are very far from having consistent style in that regard.

src/Cmfd.cpp
samuelshaner: Many of these changes to the Oxygen comments are to address #117. I agree that having the first word capitalized is a good choice. I tried to update all the comments to consistently have the first word capitalized.

geogunow:
This PR should be third in line to be merged after #187 and #162.

samuelshaner:
I just push some commits that addressed the issues raised from review of this PR. Namely:
- Typos
- Style guide issues. I fixed `){` issues, but for the most part kept the line continuation indentation as is.
- Changed `cell` to `cell_id` in many places in Cmfd.
- Created Cmfd corner constants and an array for storing corner currents separately from surface currents.
- Ran valgrind on the code and addressed memory leak issues introduced by these changes and a few lingering memory leaks. It is important to note that there are still several memory leaks in the code.
- Changed all doxygen comments in Cmfd to have the first word capitalized.

Should be ready for @wbinventor and @geogunow to review.

geogunow:
I think this PR should be next in line for merging. I'll start reviewing it.

src/Cmfd.cpp
```diff
@@ -331,58 +362,61 @@ void Cmfd::computeXS() {
...
     flux_next, f, f_next, length, length_perpen, next_length_perpen, \
-    sense, next_surface, cell, cell_next)
+    sense, next_surface, cell_id, cell_next_id, current_in, current_out)
   for (int y = 0; y < _num_y; y++) {
 
     /* Loop over Mesh cells in x direction */
     for (int x = 0; x < _num_x; x++) {
 
-      cell = y*_num_x+x;
+      cell_id = y*_num_x+x;
```
geogunow: spacing: see https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Horizontal_Whitespace

samuelshaner: @geogunow - can you clarify what you are referring to?

src/Cmfd.cpp
```diff
@@ -446,11 +480,16 @@ void Cmfd::computeDs(int moc_iteration) {
             d_hat = 2.0 * d * f * d_next * f_next / (length_perpen
                     * d * f + next_length_perpen * d_next*f_next);
```
geogunow: be consistent with the spacing here (per https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Horizontal_Whitespace separating operators with spaces is the default but can be removed for terms i.e. ``next_length_perpen*d_next*f_next`` but I don't think ``next_length_perpen * d_next*f_next`` is correct style.
t I don't think ``next_length_perpen * d_next*f_next`` is correct styl

samuelshaner: Ok, I'll make that consistent.

src/Cmfd.cpp
```diff
@@ -606,311 +601,140 @@ FP_PRECISION Cmfd::computeKeff(int moc_iteration) {
...
   /* Loop over cells */
-  #pragma omp parallel for private(value, volume, cell, row, material)
+  #pragma omp parallel for private(value, volume, cell_id, material)
   for (int y = 0; y < _num_y; y++) {
     for (int x = 0; x < _num_x; x++) {
 
-      cell = y*_num_x + x;
-      material = _materials[cell];
-      volume = _volumes[cell];
+      cell_id = y*_num_x + x;
```
geogunow: good style, use this same style for the previous comment above

samuelshaner: Why do you prefer this style? It isn't clear to me that the google style guide differentiates between different types of operators with regards to white space.

geogunow: They say these type of operators (+, -, divide, etc) should be separated by a space but they make an exception for terms (i.e. a*b + c). So my interpretation is always separate the operator with spaces unless you have a reason not to (i.e. a*b*c + d*e*f + h*g*i).

samuelshaner: Okay, I'll change in my PR; I will note that this isn't really consistent throughout OpenMOC.

src/Cmfd.cpp
geogunow: Change ``y*_num_x+x`` to ``y*_num_x + x``

geogunow:
After style is addressed, I think this PR is ready to merge. It passed all my tests.

samuelshaner:
@geogunow - I fixed the style guide issues you referred to.

PR #183, opened 2015-07-25
Keff calculation based on successive fission sources

wbinventor:
This PR revises the way that the eigenvalue is computed to use fission sources from successive MOC iterations. In the current formulation from PR #115, the eigenvalue is computed directly based on the total, fission, scattering and leakage rates to account for multiplication from both nu-fission and (n,xn) reactions. Although this is sound, it does lead to some odd convergence behavior, such as that noted in #131. This likely the result of the boundary flux lagging the FSR scalar flux and hence oscillations in the convergence of the leakage (just my postulation). The new successive source formulation avoids this behavior since it does not need the leakage in the calculation. 

The switch to using successive fission sources has allowed all code related to the leakage to be removed which (marginally) improves both performance and the overall memory footprint. These changes have been implemented and verified for the ``CPUSolver``, ``GPUSolver`` and ``VectorizedSolver``. The new methodology has *only* been implemented for pure MOC eigenvalue calculations and is not included in CMFD. (@samuelshaner, how difficult would this be to translate over to CMFD, perhaps in #182?)

The successive fission source methodology for the eigenvalues will also be consistent for adjoint calculations, which I will provide in a separate PR later today. The direct methodology would *not* have been consistent as it would have required transposing both the scattering and fission source calculation routines. Rather than having two different modes to compute the forward/adjoint eigenvalues, it will be best to have a single streamlined calculation that will work for both.

### DIFF
--- a/src/CPUSolver.cpp
+++ b/src/CPUSolver.cpp
@@ -256,8 +256,10 @@ void CPUSolver::normalizeFluxes() {
 
   #pragma omp parallel for schedule(guided)
   for (int r=0; r < _num_FSRs; r++) {
-    for (int e=0; e < _num_groups; e++)
+    for (int e=0; e < _num_groups; e++) {
       _scalar_flux(r,e) *= norm_factor;
+      _old_scalar_flux(r,e) *= norm_factor;
+    }
   }
 
   /* Normalize angular boundary fluxes for each Track */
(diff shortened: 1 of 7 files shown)


samuelshaner:
I tested this out with `c5g7-cmfd.py` and it seems to work just fine. We actually already use this methodology for computed k-eff in CMFD and looking over the code, I don't see any reason why CMFD would be affected.

samuelshaner:
Also, when CMFD is turned on, the converged k-eff from CMFD is used in the MOC solver for normalizing the fission source and computing the fission source residual, not the k-eff computed in the MOC solver classes.

wbinventor:
Excellent, if this is the way keff is compute in CMFD, then I don't know of anything else missing in this PR.

samuelshaner:
Just tested this out on the GPU and I think the PR looks good to go. Do you want me to merge it in before the k-nearest PR?

wbinventor:
If you are ready, go ahead! I plan to test the nearest neighbors out again this afternoon and merge it in if things look good. If you merge this PR, just update the nearest neighbors PR with the new code on develop and I'll give it a spin.

samuelshaner:
Alright, I'll merge this now and then update the nearest neighbors PR as necessary.

PR #184, opened 2015-07-26
Simplified TrackGenerator::initializeBoundaryConditions() method

samuelshaner:
The `TrackGenerator::initializeBoundaryCondtions()` method is ~300 lines and unnecessarily difficult to understand. This PR updates this method with a much more condensed scheme to set the reflecting tracks and track boundary conditions. Additionally, using the new scheme, it will be cleaner to implement periodic boundary conditions, as will be shown in a forthcoming PR. 

This PR builds on the PR #145 and PR  #182.


wbinventor:
Perhaps @geogunow can review this one since this will be most relevant to him?

samuelshaner:
Note that this PR will reduce to <100 lines in one file once PR #182 is merged, so it's not as daunting to review as it seems!

src/Cmfd.cpp
```diff
@@ -498,8 +475,8 @@ void Cmfd::computeDs(int moc_iteration) {
             d_tilde = 0.0;
           else
             d_tilde =
-                _materials[cell]->getDifTilde()[surface*_num_cmfd_groups + e] *
-                (1 - _relax_factor) + _relax_factor * d_tilde;
+              _materials[cell]->getDifTilde()[surface*_num_cmfd_groups + e] *
+              (1 - _relax_factor) + _relax_factor * d_tilde;
```
geogunow: I think this line should be indented

src/Cmfd.cpp
```diff
@@ -606,311 +539,140 @@ FP_PRECISION Cmfd::computeKeff(int moc_iteration) {
...
-  sum_new = pairwise_sum(_new_source, _num_x*_num_y*_num_cmfd_groups);
-  scale_val = _num_x*_num_y*_num_cmfd_groups / sum_new;
-  vector_scale(_new_flux, scale_val, _num_x*_num_y*_num_cmfd_groups);
-  matrix_multiplication(_M, _old_flux, _old_source, _num_x*_num_y, 
-          _num_cmfd_groups);
-  sum_old = pairwise_sum(_old_source, _num_x*_num_y*_num_cmfd_groups);
-  scale_val = _num_x*_num_y*_num_cmfd_groups / sum_old;
-  vector_scale(_old_flux, scale_val, _num_x*_num_y*_num_cmfd_groups);
+  matrixMultiplication(_M, _new_flux, _new_source);
+  matrixMultiplication(_M, _old_flux, _old_source);
```
geogunow: We might want to consider making a matrix class that supports operator overloading for matrix multiplication

src/Cmfd.cpp
```diff
@@ -1333,217 +1056,104 @@ void Cmfd::setNumFSRs(int num_fsrs) {
...
+            _surface_currents->incrementValue
+              (cell, SURFACE_Y_MAX*_num_cmfd_groups + e, current);
+            
+            if (x != _num_x - 1)
+              _surface_currents->incrementValue
+                (cell+1, SURFACE_Y_MAX*_num_cmfd_groups + e, current);
+            if (y != _num_y - 1)
+              _surface_currents->incrementValue
+                (cell+_num_x, SURFACE_X_MAX*_num_cmfd_groups + e, current);
+          }
```
geogunow: It seems these corner blocks have a lot in common. Would it be possible to abstract this for any corner case based on the surface index?

src/Cmfd.cpp
```diff
@@ -1551,31 +1161,29 @@ void Cmfd::splitCorners() {
...
   }
-
-  else if (surface_id == 2) {
+  else if (surface_id == SURFACE_X_MAX) {
     if (cell_num % _num_x != _num_x - 1)
       cell_next = cell_num + 1;
   }
-
-  else if (surface_id == 3) {
+  else if (surface_id == SURFACE_Y_MAX) {
```
geogunow: Ditto with these ... any way we could determine this based on surface number ... such as having an array contain [_num_x, _num_y, _num_z]

src/Cmfd.cpp
```diff
@@ -1872,71 +1476,71 @@ void Cmfd::generateKNearestStencils(){
...
       else if (iter->first == 7)
         ratio += (1.0 - iter->second/total_distance) *
-          getFluxRatio(cell + _num_x, group);
+          _flux_ratio->getValue(cell + _num_x, group);
 
       /* TOP RIGHT CORNER */
       else if (iter->first == 8)
         ratio += (1.0 - iter->second/total_distance) *
-          getFluxRatio(cell + _num_x + 1, group);
+          _flux_ratio->getValue(cell + _num_x + 1, group);
```
geogunow: Again, can we abstract all this?

src/Geometry.cpp
```diff
@@ -1081,10 +1081,10 @@ void Geometry::initializeCmfd() {
...
 
   /* Set CMFD mesh boundary conditions */
-  _cmfd->setBoundary(0, getMinXBoundaryType());
-  _cmfd->setBoundary(1, getMinYBoundaryType());
-  _cmfd->setBoundary(2, getMaxXBoundaryType());
-  _cmfd->setBoundary(3, getMaxYBoundaryType());
+  _cmfd->setBoundary(SURFACE_X_MIN, getMinXBoundaryType());
+  _cmfd->setBoundary(SURFACE_Y_MIN, getMinYBoundaryType());
+  _cmfd->setBoundary(SURFACE_X_MAX, getMaxXBoundaryType());
+  _cmfd->setBoundary(SURFACE_Y_MAX, getMaxYBoundaryType());
```
geogunow: Maybe do this in a for loop?

geogunow:
This is definitely an improvement over what we do now. If possible, I would like to see even more abstraction to avoid the long conditional blocks.

src/Cmfd.cpp
wbinventor: I tend to agree that this would be a nice feature too. But I would also be
fine without this for this PR if it's not how @samuelshaner wants to spend
his time.

src/Cmfd.cpp
wbinventor: Agreed, abstraction would be nice if possible, but isn't a must for me in
this PR if it isn't obvious how to go about it.

src/Cmfd.cpp
samuelshaner: The line is was indented 4 spaces and I just changed the indent back to 2 spaces.

src/Geometry.cpp
samuelshaner: I don't see an easy way to make this shorter with a loop since the `get___boundaryType()` methods are all unique.

wbinventor:
I'll wait to review this and your other PRs @samuelshaner after #182 is finalized.

geogunow:
This PR should be fourth in line to be merged after #187, #162, and #182. Let's those done by tonight/ tomorrow so we can start reviewing this PR.

geogunow:
I just looked it over. Great improvement! I think this is good to

geogunow:
I just looked it over. Great improvement! I think this is good to

PR #185, opened 2015-07-26
Periodic Boundary Conditions

samuelshaner:
This PR implements periodic boundary conditions. Along with the modifications to implement periodic BCs, some of the language of describing the next track in and track out was changed since the next track can either be a reflective track or periodic track. This PR has been tested for both the CPUSolver and GPUSolver. The `simple-lattice.py` input file has also been modified to now have all periodic BCs.

This PR depends on PRs #145, #182, and #184 and should thus be reviewed after those are merged.


wbinventor:
@geogunow can you take this one on as well? I can certainly help too. Both this and #184 should be much smaller once we merge in #182.

geogunow:
I would like to make a mini test suite to analyze it. I'm a bit uneasy about this PR right now, due to the issues in 3D MOC. I don't think they are related, but better safe than sorry.

src/CPUSolver.cpp
```diff
@@ -814,24 +816,22 @@ void CPUSolver::transferBoundaryFlux(int track_id,
...
-    start = _tracks[track_id]->isReflIn() * _polar_times_groups;
-    bc = (int)_tracks[track_id]->getBCIn();
+    start = _tracks[track_id]->isNextIn() * _polar_times_groups;
+    bc = std::min((int)_tracks[track_id]->getBCIn(), 1);
     track_out_id = _tracks[track_id]->getTrackIn()->getUid();
   }
 
   FP_PRECISION* track_out_flux = &_boundary_flux(track_out_id,0,0,start);
-
+
```
geogunow: remove space

src/Cmfd.cpp
```diff
@@ -681,6 +681,14 @@ void Cmfd::constructMatrices() {
                      [SURFACE_X_MIN*_num_cmfd_groups + e]) * _cell_height;
           _A->incrementValue(cell_id - 1, e, cell_id, e, value);
         }
+        else if (_boundaries[SURFACE_X_MIN] == PERIODIC) {
+          value = - (material->getDifHat()[SURFACE_X_MIN*_num_cmfd_groups + e]
+                     - material->getDifTilde()
+                     [SURFACE_X_MIN*_num_cmfd_groups + e])
+            * _cell_height;
```
geogunow: spacing is a bit weird here

src/Cmfd.cpp
```diff
@@ -1652,7 +1690,7 @@ FP_PRECISION Cmfd::getDistanceToCentroid(Point* centroid, int cell_id,
                  2.0);
     found = true;
   }
-
+
```
geogunow: remove space

geogunow:
I just reviewed this PR and it seems to be in pretty good shape. I noticed you changed the simple lattice to be periodic which makes sense. However, I would still like to see a simple sample-input case (much like the gradients case which I added) which would very clearly show this is working. Some of the details in this PR are difficult for me to wrap my head around but if there were a test case clearly showing the periodic behavior, that would really put my mind at ease.

src/CPUSolver.cpp
```diff
@@ -681,13 +681,13 @@ void CPUSolver::transportSweep() {
   if (_cmfd != NULL && _cmfd->isFluxUpdateOn())
     _cmfd->zeroCurrents();
 
-  /* Loop over azimuthal angle halfspaces */
-  for (int i=0; i < 2; i++) {
+  /* Loop over azimuthal angle and periodic track halfspaces */
+  for (int i=0; i < 6; i++) {
```
wbinventor: Maybe it's just me, but I found that "2" made relatively good intuitive sense for the term "halfspaces" but do not think that "6" does. I know this PR has been merged in already and I didn't have time to review it beforehand. Perhaps the term "halfspaces" is a bit of a misnomer now? Plus, I always dislike seeing somewhat arbitrary values hard-coded in places like this.

src/accel/cuda/GPUSolver.cu
```diff
@@ -1454,18 +1452,54 @@ void GPUSolver::transportSweep() {
   /* Initialize flux in each FSR to zero */
   flattenFSRFluxes(0.0);
 
-  /* Sweep the first halfspace of azimuthal angle space */
-  tid_offset = 0;
-  tid_max = (_tot_num_tracks / 2);
+  /* Sweep the first space of azimuthal angle and periodic track halfspaces */
+  tid_offset = _num_tracks_by_halfspace[0];
```
wbinventor: Again, I missed this during the initial review, but think that this function bloat could be conslidated into a nice for loop.

src/accel/cuda/clone.cu
```diff
@@ -98,11 +98,12 @@ void clone_track(Track* track_h, dev_track* track_d,
...
   new_track._num_segments = track_h->getNumSegments();
   new_track._azim_angle_index = track_h->getAzimAngleIndex();
-  new_track._refl_in = track_h->isReflIn();
-  new_track._refl_out = track_h->isReflOut();
-  new_track._bc_in = track_h->getBCIn();
-  new_track._bc_out = track_h->getBCOut();
+  new_track._next_in = track_h->isNextIn();
+  new_track._next_out = track_h->isNextOut();
 
+  new_track._bc_in = std::min((int)track_h->getBCIn(), 1);
```
wbinventor: Why the need for ``std::min`` here?

src/TrackGenerator.cpp
```diff
@@ -491,6 +528,7 @@ void TrackGenerator::generateTracks() {
     /* Allocate memory for the Tracks */
     try {
       _num_tracks = new int[_num_azim];
+      _num_tracks_by_halfspace = new int[7];
```
wbinventor: Seems pretty arbitrary without a comment

src/CPUSolver.cpp
samuelshaner: I agree, it is a bit of a misnomer now. At first glance, it appears that periodic boundary conditions could be separated into their own half space, but after thinking about it for a while I convinced myself that there are situations that require 3 half spaces (or rather 3 separate parallel transport sweeps). I changed it to `_num_parallel_track_groups` in my 3D-MOC branch. Any ideas on a term that would be most fitting?

src/accel/cuda/clone.cu
samuelshaner: VACUUM BCs (enum value 0) require the new_track._bc_in to be set to 0 while REFLECTIVE (enum value 1) and PERIODIC (enum value 2) reequire the new_track._bc_in to be 1. Would you prefer this replace with an if/else? ideas?

wbinventor: Ok, makes sense. I do think that while an ``if-else`` requires more lines of code, it is more transparent. Perhaps a ``switch-case`` block over the (currently 3) different BC types?

src/CPUSolver.cpp
wbinventor: I think that ``_num_parallel_track_groups`` is much clearer.

src/accel/cuda/clone.cu
samuelshaner: How about just an if/else? If the bc is vacuum, set bc to 0; else set bc to 1. This will work for the current BC options and will work for rotational BCs, if we ever decide to implement them. The only other BC I can think of is white BCs and I'm not sure how they would be implemented, but I don't see any need to even consider those BCs at the moment.

wbinventor: Sounds like good reasoning to me.

PR #186, opened 2015-07-26
Use of full fission matrix for source updates

wbinventor:
This PR introduces a new fission matrix ``_fiss_mat`` attribute for the ``Material`` class. The fission matrix is now used when computing the FSR sources from scattering and fission. The goal behind the fission matrix is to allow as much code as possible to be re-used for the upcoming adjoint solver (in this case, the scatter and fission matrices simply need to be transposed). This PR is a good follow-on to #183 and may be reviewed after that one. The fission matrix has been validated for the ``CPUSolver``, ``VectorizedSolver`` and ``GPUSolver``.

### DIFF
--- a/src/CPUSolver.cpp
+++ b/src/CPUSolver.cpp
@@ -246,7 +246,7 @@ void CPUSolver::normalizeFluxes() {
   /* Normalize scalar fluxes in each FSR */
   norm_factor = 1.0 / tot_fission_source;
 
-  log_printf(NORMAL, "Tot. Fiss. Src. = %f, Norm. factor = %f",
+  log_printf(DEBUG, "Tot. Fiss. Src. = %f, Norm. factor = %f",
              tot_fission_source, norm_factor);
 
   #pragma omp parallel for schedule(guided)
@@ -290,7 +290,7 @@ void CPUSolver::computeFSRSources() {
   FP_PRECISION* scatter_sources = new FP_PRECISION[size];
 
   /* For all FSRs and compute the total source */
-  #pragma omp parallel for private(tid, material, sigma_s, fiss_mat, \
+  #pragma omp parallel for private(tid, material, sigma_t, sigma_s, fiss_mat, \
     fission_source, scatter_source) schedule(guided)
   for (int r=0; r < _num_FSRs; r++) {
 
--- a/src/CPUSolver.cpp
+++ b/src/CPUSolver.cpp
@@ -278,49 +278,49 @@ void CPUSolver::normalizeFluxes() {
  */
 void CPUSolver::computeFSRSources() {
 
-  int tid;
-  Material* material;
-  FP_PRECISION* sigma_t;
-  FP_PRECISION sigma_s, fiss_mat;
-  FP_PRECISION scatter_source, fission_source;
+  #pragma omp parallel default(none)
+  {
+    int tid;
+    Material* material;
+    FP_PRECISION* sigma_t;
+    FP_PRECISION sigma_s, fiss_mat;
+    FP_PRECISION scatter_source, fission_source;
+    FP_PRECISION* fission_sources = new FP_PRECISION[_num_groups];
+    FP_PRECISION* scatter_sources = new FP_PRECISION[_num_groups];
 
-  int size = _num_threads * _num_groups;
-  FP_PRECISION* fission_sources = new FP_PRECISION[size];
-  FP_PRECISION* scatter_sources = new FP_PRECISION[size];
+    /* Compute the total source for each FSR */
+    #pragma omp parallel for schedule(guided)
+    for (int r=0; r < _num_FSRs; r++) {
 
-  /* Compute the total source for each FSR */
-  #pragma omp parallel for private(tid, material, sigma_t, sigma_s, fiss_mat, \
-    fission_source, scatter_source) schedule(guided)
-  for (int r=0; r < _num_FSRs; r++) {
+      tid = omp_get_thread_num();
+      material = _FSR_materials[r];
+      sigma_t = material->getSigmaT();
 
-    tid = omp_get_thread_num();
-    material = _FSR_materials[r];
-    sigma_t = material->getSigmaT();
+      /* Compute scatter + fission source for group G */
+      for (int G=0; G < _num_groups; G++) {
+        for (int g=0; g < _num_groups; g++) {
+          sigma_s = material->getSigmaSByGroup(g+1,G+1);
+          fiss_mat = material->getFissionMatrixByGroup(g+1,G+1);
+          scatter_sources[g] = sigma_s * _scalar_flux(r,g);
+          fission_sources[g] = fiss_mat * _scalar_flux(r,g);
+        }
 
-    /* Compute scatter + fission source for group G */
-    for (int G=0; G < _num_groups; G++) {
-      for (int g=0; g < _num_groups; g++) {
-        sigma_s = material->getSigmaSByGroup(g+1,G+1);
-        fiss_mat = material->getFissionMatrixByGroup(g+1,G+1);
-        scatter_sources(tid,g) = sigma_s * _scalar_flux(r,g);
-        fission_sources(tid,g) = fiss_mat * _scalar_flux(r,g);
+        scatter_source = pairwise_sum<FP_PRECISION>(scatter_sources, 
+                                                    _num_groups);
+        fission_source = pairwise_sum<FP_PRECISION>(fission_sources,
+                                                    _num_groups);
+        fission_source /= _k_eff;
+
+        /* Compute total (scatter+fission+fixed) reduced source */
+        _reduced_sources(r,G) = _fixed_sources(r,G);
+        _reduced_sources(r,G) += scatter_source + fission_source;
+        _reduced_sources(r,G) *= ONE_OVER_FOUR_PI / sigma_t[G];
       }
-
-      scatter_source = pairwise_sum<FP_PRECISION>(&scatter_sources(tid,0),
-                                                  _num_groups);
-      fission_source = pairwise_sum<FP_PRECISION>(&fission_sources(tid,0),
-                                                  _num_groups);
-      fission_source /= _k_eff;
-
-      /* Compute total (scatter+fission+fixed) reduced source */
-      _reduced_sources(r,G) = _fixed_sources(r,G);
-      _reduced_sources(r,G) += scatter_source + fission_source;
-      _reduced_sources(r,G) *= ONE_OVER_FOUR_PI / sigma_t[G];
     }
-  }
 
-  delete [] fission_sources;
-  delete [] scatter_sources;
+    delete [] fission_sources;
+    delete [] scatter_sources;
+  }
 }
 
 
--- a/src/CPUSolver.cpp
+++ b/src/CPUSolver.cpp
@@ -289,7 +289,7 @@ void CPUSolver::computeFSRSources() {
     FP_PRECISION* scatter_sources = new FP_PRECISION[_num_groups];
 
     /* Compute the total source for each FSR */
-    #pragma omp parallel for schedule(guided)
+    #pragma omp for schedule(guided)
     for (int r=0; r < _num_FSRs; r++) {
 
       tid = omp_get_thread_num();
--- a/src/CPUSolver.cpp
+++ b/src/CPUSolver.cpp
@@ -246,7 +246,7 @@ void CPUSolver::normalizeFluxes() {
   /* Normalize scalar fluxes in each FSR */
   norm_factor = 1.0 / tot_fission_source;
 
-  log_printf(DEBUG, "Tot. Fiss. Src. = %f, Norm. factor = %f",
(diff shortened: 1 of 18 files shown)


src/CPUSolver.cpp
```diff
@@ -279,51 +279,42 @@ void CPUSolver::normalizeFluxes() {
...
-  int size = _num_FSRs * _num_groups;
+  int size = _num_threads * _num_groups;
   FP_PRECISION* fission_sources = new FP_PRECISION[size];
-  size = _num_threads * _num_groups;
   FP_PRECISION* scatter_sources = new FP_PRECISION[size];
 
-  /* For all FSRs, find the source */
-  #pragma omp parallel for private(tid, material, nu_sigma_f, chi, \
-    sigma_t, fission_source, scatter_source) schedule(guided)
+  /* For all FSRs and compute the total source */
```
samuelshaner: Might take another look at this comment.

src/VectorizedSolver.cpp
```diff
@@ -322,76 +328,52 @@ void VectorizedSolver::normalizeFluxes() {
...
-  FP_PRECISION* sigma_t;
-  FP_PRECISION* chi;
   Material* material;
+  FP_PRECISION* sigma_t;
+  FP_PRECISION* sigma_s;
+  FP_PRECISION* fiss_mat;
+  FP_PRECISION scatter_source, fission_source;
 
-  int size = _num_FSRs * _num_groups * sizeof(FP_PRECISION);
+  int size = _num_threads * _num_groups * sizeof(FP_PRECISION);
```
geogunow: Can someone explain to me why we are making the memory size dependent on the number of threads? Is each thread getting a copy of its own data? And do we really need to do that?

src/accel/cuda/GPUSolver.cu
```diff
@@ -693,27 +686,28 @@ FP_PRECISION GPUSolver::getFSRSource(int fsr_id, int group) {
...
-  }
-
-  fission_source /= _k_eff;
-
-  /* Compute total scattering source for this FSR */
+  /* Compute total scattering and fission sources for this FSR */
   for (int g=0; g < _num_groups; g++) {
     scatter_source += sigma_s[(group-1)*(_num_groups)+g] 
-                    * fsr_scalar_fluxes[g];
+                      * fsr_scalar_fluxes[g];
```
geogunow: Which indentation style is correct? It seems we have two different styles between the old and new.

src/linalg.h
```diff
@@ -75,7 +80,27 @@ inline void matrix_multiplication(T** matrix, T* vector_x,
...
+ */
+template<typename T>
+inline void matrix_transpose(T* matrix, int dim1, int dim2) {
+
+  std::vector<T> temp;
+
+  for (int i=0; i < dim1; i++) {
+    for (int j=0; j < dim2; j++)
+      temp.push_back(matrix[j * dim1 + i]);
+  }
```
geogunow: It might not make a huge difference, but this is an inefficient way to make a transpose. We know the size of the matrix we need to make so we can allocate all the memory upfront. Also I believe the linear algebra folks would tell you to reference the same data - just with a different indexer to remove unnecessary extra storage of data.

geogunow:
I think this PR should be ready to merge soon - after the comments from Sam and myself are addressed. Some of them are stylistic or have to do with the code structure in general so a change might not necessarily need to be made. But I would be interested in an explanation.

src/CPUSolver.cpp
wbinventor: good call

src/VectorizedSolver.cpp
wbinventor: The reason is that the ``fission_sources`` and ``scatter_sources`` arrays are "scratchpad" arrays for each thread. The outer loop is over FSRs while the inner loops are such that each thread accumulates the source *into* each group *from* every group. 

Now one of the things that I realized when I was revisiting this code to make this change is that this code block will suffer from the same false sharing issue as was once an issue for the transport sweep. Since computing FSR sources takes up a relatively small fraction of time compared to the sweeps, I decided to keep this structure in place since it makes the code (slightly) more readable at the expense of poor parallel scaling. I'm happy to change this if you think it necessary.

Perhaps you have a better approach in mind. If so, I can put it into place or you can submit a separate PR.

src/accel/cuda/GPUSolver.cu
wbinventor: Good question! Right now, our Style Guide does not specify the indentation for continuation lines. OpenMC's [Style Guide](https://mit-crpg.github.io/openmc/devguide/styleguide.html#indentation) specifies that continuation lines should be indented an extra 5 spaces relative to the first line. I haven't seen anything in the Google C++ Style Guide or otherwise that indicates a particular convention. However, I do think that it would be good to specify one, though the current codebase would likely be far from meeting it at this point in time! What do you like best @geogunow?

src/linalg.h
wbinventor: Excellent point, the ``temp`` vector now calls ``resize()`` to allocate the appropriate amount of memory upon initialization. I agree that this is almost guaranteed to not be the most cache efficient implementation of a matrix transpose. But since this is definitely not a performance critical section of code I just went with the most intuitive implementation that came to mind.

wbinventor:
@geogunow since you raised a comment about performance I thought I should address this question more fully here. The introduction of a dense and explicit fission matrix is most certainly not the most efficient way to compute fission sources. However, as I recall from some of yours, @jtramm's, and my own performance analyses, for problems with few energy groups (i.e., <100), the compute time spent in the transport sweep (95+%) easily dominates the time spent computing sources (<2%). Furthermore, when like materials are replicated in many different spatial zones, the memory footprint for explicit fission matrices, although inefficient, is far outpaced by that for other data structures such as tracks. Of course as the number of energy groups increase, both the performance and memory penalty of explicitly stored fission matrices will change this picture somewhat.

Furthermore, it was not immediately apparent how I might expose a sparse fission matrix in a relatively seamless way to the ``VectorizedSolver`` and ``GPUSolver``. As you know, conditional "if-else" statements do not play nicely with SIMD vectorization. As a result, this PR adds some extra useless flops (to an extent) to iterate over explicit scattering and fission matrices at this point in time. If someone comes along and can do a good job of implementing this in a vectorized form with sparse matrices - which I am confident can be done, with (much) time and effort - then I'll be the first supporter! But for now, the code is simpler with explicit (smallish) sparse production matrices.

Anyway, that's what I told myself as I implemented this in an attempt to make adjoint calculations in #187 super straightforward

src/accel/cuda/GPUSolver.cu
geogunow: I really don't have a preference too much, except Vim I believe does have a default (I'm not sure about Emacs). I would opt for whatever the default is.

wbinventor: What is the default? Emacs does have one but I find myself disagreeing with
it quite often. But that's not a good reason. If you like vim's default,
add it to the style guide and I'll start abiding by it.

geogunow: Ok, I just looked into Vim - the default is just one more tab equivalent (so 2 spaces in that case)

wbinventor: Ok, I'll go with that if the other developers concur.

geogunow:
Ok, I see your point. I'm just worried about HPC concerns for the target problems we are interested in (~100 energy groups, solved on next generation architectures - maybe ~512 threads). At a high number of threads the added memory cost of the scratchpads might become an issue? I'm not entirely sure about that. I'll talk with @jtramm more about it tomorrow morning and try to come to a consensus.

wbinventor:
Ok that certainly makes sense. Do note that one more reason that it is done
this is that it allows for pairwise sums to be used to mitigate roundoff
error. You could get away without the scratchpads if you do away with
pairwise sums, but then this will lead to (very) erroneous results for very
large problems. Of course you guys may not have noticed these sorts of
issues in SimpleMOC :-) My opinion is that in terms of the menory
footprint, these scratchpad arrays should not be a concern because they are
still pretty tiny (even with 512 threads x 100 groups) relative to
everything else.

src/VectorizedSolver.cpp
```diff
@@ -322,76 +328,52 @@ void VectorizedSolver::normalizeFluxes() {
...
   FP_PRECISION* fission_sources = (FP_PRECISION*)MM_MALLOC(size, VEC_ALIGNMENT);
-  size = _num_threads * _num_groups * sizeof(FP_PRECISION);
   FP_PRECISION* scatter_sources = (FP_PRECISION*)MM_MALLOC(size, VEC_ALIGNMENT);
 
   /* For all FSRs, find the source */
-  #pragma omp parallel for private(material, nu_sigma_f, chi, \
-    sigma_s, sigma_t, fission_source, scatter_source) schedule(guided)
+  #pragma omp parallel for private(tid, material, sigma_t, sigma_s, fiss_mat, \
+    fission_source, scatter_source) schedule(guided)
```
jtramm: A possibly cleaner approach might be to make thread private allocations inside the parallel region, which should reduce the chances of false sharing. It also eliminates the need to manually index each thread into the shared array, making the code a little simpler. This can be accomplished by breaking the OMP parallel pragma and OMP for pragma apart. For instance,
```
#pragma omp parallel default(none) private(...) shared(...)
{
  int size = _num_groups * sizeof(FP_PRECISION);
  FP_PRECISION * fission_sources = (FP_PRECISION*) MM_MALLOC(size, VEC_ALIGNMENT);
  #pragma omp for schedule (guided)
  for( ){
  }
}
```

wbinventor: Thanks for the help @jtramm! I just pushed an update which used your approach for both ``CPUSolver::computeFSRSources()`` and ``VectorizedSolver::computeFSRSources()``. Although this doesn't address @geogunow's concern about memory limitations, it does address the false sharing issue which I alluded to above.

I tried to apply this approach to allocate the ``thread_fsr_fluxes`` array in the ``CPUSolver::transportSweep()`` once for each thread rather than over and over again for each track, but was unable to get it working. Sometimes the OpenMP is like voodoo to me. That is outside the scope of this PR, but might be something to look at it in the future (albeit it should hardly affect performance at all).

wbinventor:
@geogunow I think this addresses your comments save for the continuation line indentation. I think it might be best to save that for another PR once you've canonized a rule for it in our style guide.

jtramm:
If you do a:
```
#pragma omp parallel...
{
   #pragma omp for...
  for()
  {
  }
}
```
Then this should have the desired effect. It will decompose the work across threads automatically.

As you have it now:
```
#pragma omp parallel...
{
   #pragma omp parallel for...
  for()
  {
  }
}
```
Will cause issues, as each "parallel" declaration spawns a new set of threads (in this case, squaring the number of threads you have alive). The outer set of threads will each spawn their own set of threads, and then all the outer threads will essentially compute the same thing.

Basically, the best way to think about it is that:
```
#pragma omp parallel
```
spawns a set of threads, and
```
#pragma omp for
```
tells the compiler that the subsequent for loop should be decomposed across the threads active within the parallel region that you're in.

Often, these two pragmas are combined so that both commands can be given at once, but separating them allows for thread local variabes or allocations to be declared before the work loop actually starts.

Hope that makes things clearer!

wbinventor:
Thanks for the explanation @jtramm! I just removed the ``parallel`` in
place of ``#pragma omp for``. I think I recall learning about this when I
was using OpenMP a lot a few years back, but found it more transparent (and
less prone to errors) to just use the simple ``#pragma omp parallel for``
than create more generic parallel code regions. But in this case, it works,
it makes sense, it is clearly the way to go, and I think I'll structure my
parallel code in this way in the future to help avoid false sharing.

On Thu, Jul 30, 2015 at 9:04 AM, John Tramm <notifications@github.com>
wrote:

> If you do a:
>
> #pragma omp parallel...
> {
>    #pragma omp for...
>   for()
>   {
>   }
> }
>
> Then this should have the desired effect. It will decompose the work
> across threads automatically.
>
> As you have it now:
>
> #pragma omp parallel...
> {
>    #pragma omp parallel for...
>   for()
>   {
>   }
> }
>
> Will cause issues, as each "parallel" declaration spawns a new set of
> threads (in this case, squaring the number of threads you have alive). The
> outer set of threads will each spawn their own set of threads, and then all
> the outer threads will essentially compute the same thing.
>
> Basically, the best way to think about it is that:
>
> #pragma omp parallel
>
> spawns a set of threads, and
>
> #pragma omp for
>
> tells the compiler that the subsequent for loop should be decomposed
> across the threads active within the parallel region that you're in.
>
> Often, these two pragmas are combined so that both commands can be given
> at once, but separating them allows for thread local variabes or
> allocations to be declared before the work loop actually starts.
>
> Hope that makes things clearer!
>
> —
> <https://github.com/mit-crpg/OpenMOC/pull/186#issuecomment-126383579>.
>



-- 
Will Boyd
Nuclear Science & Engineering
Cell: 423.413.8469
wbinventor@gmail.com

wbinventor:
@geogunow this has been updated per yours and @jtramm's comments and is ready for any further review or testing you want to do.

geogunow:
Everything checks out to me

PR #187, opened 2015-07-26
New adjoint calculation mode

wbinventor:
This PR introduces a new adjoint calculation mode for both fixed source and criticality problems. The adjoint mode builds upon PRs #183 and #186 which made it surprisingly simple to this feature. There is a new ``solverMode`` enumerated type which can take on values ``FORWARD`` or ``ADJOINT``. All of the following routines have been modified to take a new ``mode`` parameter (which defaults to ``FORWARD``):

* ``Solver::computeFlux(...)``
* ``Solver::computeSource(...)``
* ``Solver::computeEigenvalue(...)``

The following snippet illustrates how easily one may now toggle between ``FORWARD`` and ``ADJOINT`` modes:

```python
solver.computeEigenvalue(max_iters=1000, mode=ADJOINT)
```

The adjoint calculation mode is made possible by the new ``Material::transposeProductionMatrices()`` routine. If in ``ADJOINT`` mode, the ``Solver`` calls the routine when the materials are initialized. At the end of a calculation performed by one of the ``Solvers`` three main methods above, the new ``Solver::resetMaterials()`` routine will return each ``Material's`` production matrices to its original state.

In addition, this PR illustrates the use of the ``ADJOINT`` mode in the homogeneous one and two group sample input scripts. Per an email conversation with @robertsj, I've included some new code in each of these sample inputs which uses NumPy to solve for the forward and adjoint eigenvalues to compare with OpenMOC. 

On a side note, this PR breaks ups the C5G7 inputs into an importable "geometry.py" file with separate scripts to run simulations in different modes (CPU/GPU, CMFD) akin to what was discussed in #165.

Both ``FORWARD`` and ``ADJOINT`` modes have been validated for the ``CPUSolver``, ``VectorizedSolver`` and ``GPUSolver``. However, I have not modified CMFD to work for an adjoint calculation. I believe that this would simply require making CMFD use the fission matrix per #186.


samuelshaner:
@wbinventor - This PR looks really interesting! How much of a priority is it to get this PR merged? I'm happy to review it, but probably couldn't really dig into it until next weekend.

wbinventor:
No rush! I'll review and test #182 and #188 this week, and help with any of the other PRs that the other developers are able to take responsibility for. Plus, PR #186 should be reviewed and merged in first since it will make this PR (nearly) trivial to review.

samuelshaner:
Ok, I'll focus on reviewing PR #186 then.

geogunow:
Let me know when PR #186 gets merged and I'll take a look at this one as well

sample-input/benchmarks/c5g7/c5g7-cmfd.py
```diff
@@ -1,10 +1,7 @@
-from openmoc import *
+import openmoc
 import openmoc.log as log
 import openmoc.plotter as plotter
```
geogunow: Can we remove these and just call ``openmoc.plotter``? That would allow us to just import openmoc in one line. Also it would be a bit more consistent in my opinion.

wbinventor: Sure, I removed the various submodule imports from all sample input files in place of a single ``import openmoc``. Note that this change greatly increases the size of the diff for this PR, though much of it will be relatively straightforward to review.

wbinventor:
This has been updated with respect to develop. You'll notice that perhaps 99% of the changes in the diff correspond to @geogunow's request to simply import the base ``openmoc`` module in our sample input files.

geogunow:
@wbinventor I just tried running test cases (specifically the homogeneous forward flux case) with this branch and I get the following error message:
```
  File "forward-flux.py", line 2, in <module>
    import openmoc
  File "/home/geogunow/.local/lib/python2.7/site-packages/openmoc/__init__.py", line 18, in <module>
    import krylov
ImportError: No module named krylov
```

wbinventor:
Sorry about that @geogunow, looks like a stray reference to the Krylov solver in #162 crept in here while I was working on both features. I just removed it from this branch.

geogunow:
@wbinventor I am seeing some very strange behavior when trying to merge this PR. There is still a reference to ``krylov`` in the ``openmoc/__init__.py`` file even though it's not present in this PR (at least it doesn't look like it). Let me explain: in the source code on github here, it seems there is no reference to ``krylov`` in ``openmoc/__init__.py`` however it somehow gets inserted when I try to merge this branch into my local develop. To reproduce what I am seeing, I will describe how I arrived here:

To remove all changes and start fresh with the mit-crpg develop branch, I do:
```
git reset --hard upstream/develop
```
where upstream refers to this mit-crpg OpenMOC repository. Next, I remove all build files, including those installed in the .local directory. So at this point I should be working with a clean copy of the OpenMOC develop branch. Next I try to merge your branch into my local develop branch using:
```
git merge Will/adjoint
```
where ``Will`` refers to your github fork of OpenMOC. If I then open the ``openmoc/__init__.py`` file there are suddenly references to ``krylov``! Here is the file that is produced from the merge:
```
import sys, os
import random
import datetime
import signal

# For Python 2.X.X
if (sys.version_info[0] == 2):
  from openmoc import *
# For Python 3.X.X
else:
  from openmoc.openmoc import *

import log
import options
import materialize
import plotter
import process
import krylov

# Tell Python to recognize CTRL+C and stop the C++ extension module
# when this is passed in from the keyboard
signal.signal(signal.SIGINT, signal.SIG_DFL)

# Set a log file name using a date and time
now = datetime.datetime.now()
current_time = str(now.month).zfill(2) + '-' + str(now.day).zfill(2) + '-' + str(now.year) + '--'
current_time = current_time + str(now.hour).zfill(2) + ':' + str(now.minute).zfill(2)
current_time = current_time + ':' + str(now.second).zfill(2)
initialize_logger()
set_log_filename('openmoc-' + current_time + '.log');

Timer = Timer()
```

Notice the reference to ``krylov``. This causes every OpenMOC input file to fail at the beginning of execution since it has no idea what ``krylov`` is referring to!

geogunow:
Disregard my previous comment. I forgot to fetch.

geogunow:
Now ``OpenMOC/sample-input/benchmarks/c5g7/c5g7.py`` and ``OpenMOC/sample-input/benchmarks/c5g7/c5g7-cmfd.py`` are working just fine. However, ``OpenMOC/sample-input/benchmarks/homogeneous-one-group/forward-flux.py`` is not. I receive the following output:

```
[  TITLE  ]  *******************************************************************
[  TITLE  ]         Simulating a one group homogeneous infinite medium...       
[  TITLE  ]  *******************************************************************
[  HEADER ]  *****************  The reference keff = 1.43...  ******************
[  NORMAL ]  Creating materials...
[  NORMAL ]  Creating surfaces...
[  NORMAL ]  Creating cells...
[  NORMAL ]  Creating universes...
[  NORMAL ]  Creating geometry...
[  NORMAL ]  Initializing the track generator...
[  NORMAL ]  Ray tracing for track segmentation...
[  NORMAL ]  Dumping tracks to file...
[  NORMAL ]  Running MOC forward eigenvalue simulation...
  File "forward-flux.py", line 117, in <module>
    solver.computeEigenvalue(max_iters, mode=openmoc.FORWARD)
AttributeError: 'module' object has no attribute 'FORWARD'
```

wbinventor:
Try installing it twice...

On Tue, Aug 4, 2015 at 10:06 AM, Geoffrey Gunow <notifications@github.com>
wrote:

> Now OpenMOC/sample-input/benchmarks/c5g7/c5g7.py and
> OpenMOC/sample-input/benchmarks/c5g7/c5g7-cmfd.py are working just fine.
> However,
> OpenMOC/sample-input/benchmarks/homogeneous-one-group/forward-flux.py is
> not. I receive the following output:
>
> [  TITLE  ]  *******************************************************************
> [  TITLE  ]         Simulating a one group homogeneous infinite medium...
> [  TITLE  ]  *******************************************************************
> [  HEADER ]  *****************  The reference keff = 1.43...  ******************
> [  NORMAL ]  Creating materials...
> [  NORMAL ]  Creating surfaces...
> [  NORMAL ]  Creating cells...
> [  NORMAL ]  Creating universes...
> [  NORMAL ]  Creating geometry...
> [  NORMAL ]  Initializing the track generator...
> [  NORMAL ]  Ray tracing for track segmentation...
> [  NORMAL ]  Dumping tracks to file...
> [  NORMAL ]  Running MOC forward eigenvalue simulation...
> Traceback (most recent call last):
>   File "forward-flux.py", line 117, in <module>
>     solver.computeEigenvalue(max_iters, mode=openmoc.FORWARD)
> AttributeError: 'module' object has no attribute 'FORWARD'
>
> —
> <https://github.com/mit-crpg/OpenMOC/pull/187#issuecomment-127677355>.
>



-- 
Will Boyd
Nuclear Science & Engineering
Cell: 423.413.8469
wbinventor@gmail.com

geogunow:
@wbinventor that fixed it. I'm going to check every sample input file before I sign off on this PR.

geogunow:
This is probably not an issue with this PR, but the LRA benchmark does not converge within the default iterations (1000). Could we up the default number of iterations to 2000 for that case?

geogunow:
Ditto on ``bundled-lattice.py``

geogunow:
Ditto on ``point-source.py``

geogunow:
After those three files are changed so the default case actually converges, I think this PR is ready to merge. All cases check out.

geogunow:
Let's try to merge this PR tonight.

wbinventor:
@geogunow I can see your point about enforcing that the sample input files converge to within the tolerance. However, I'm not yet sold that we should enforce this. The reason is that it would require us to hard-code either 1) a lower convergence threshold or 2) a larger maximum number of iterations into a few select input files. This would over-ride any values the user might select for these parameters at runtime through the command line options (without an ``if-else`` block to check for this, which would clutter the example input files). Does this make sense? My instinct tells me that it is better to allow users to transparently set these parameters at runtime, but I'm open to changing this if that's a change others agree with.

wbinventor:
This has been updated with respect to #162. In addition, I simplified the finalization process for the following routines to streamline the call to print a timer report and reset material data:

* ``Solver::computeFlux(...)``
* ``Solver::computeSource(...)``
* ``Solver::computeEigenvalue(...)``

Finally, @cjosey's new ``IRAMSolver`` has been configured to solve for *n* forward or adjoint eigenmodes. I thought this was as simple as adding a new ``solution_mode`` parameter to be set to either ``openmoc.FORWARD`` (the default) or ``openmoc.ADJOINT``, and then passing that parameter to the ``Solver::initializeMaterials(...)`` routine to transpose the production matrices, if necessary. However, when I run this (on an AWS GPU), it does not find adjoint eigenvalues which are very nearly the forward eigenvalues. Most odd, the eigenmodes for groups 1-4 (fast) look reasonable, but those for groups 5-7 (thermal) are completely amiss:

**First adjoint eigenmode (group 1)**
fsr-flux-group-1

**First adjoint eigenmode (group 5)**
fsr-flux-group-5

I'm not sure what is going on yet, and may not look into this further till next week. I can remove the adjoint capability for the ``IRAMSolver`` for this PR if you guys want to move it in sooner.

geogunow:
@wbinventor let's concentrate on making a good PR before merging. As far as the sample inputs are concerned, I think it doesn't make sense to include examples that don't converge. If 1000 iterations is so limiting, perhaps we should change the default to 2000?

wbinventor:
Sure, I like that idea more than changing hard-coding parameters in the examples. Similarly, @samuelshaner might argue that we could relax the default convergence threshold since it is pretty low compared to some other codes.

cjosey:
I tried running this to see what was up with the interaction between the Krylov stuff and the adjoint stuff, but in the process I found this PR is incompatible with Python3:
```
  File "c5g7.py", line 1, in <module>
    import openmoc
  File "/home/cjosey/.local/lib/python3.4/site-packages/openmoc/__init__.py", line 14, in <module>
    import options
ImportError: No module named 'options'
```
It works just fine with python2.

wbinventor:
Thanks for pointing this out @cjosey - it was an easy fix for Python 3 compatibility, see my latest commit.

wbinventor:
Well, it looks like the adjoint eigenmode calculation works with ``CPUSolver``. I get the following first 5 forward eigenvalue:

[ 1.18555284, 1.04515135, 1.04515100, 0.91254997,  0.96217489]

and the following first 5 adjoint eigenvalues:

[ 1.18555617, 0.91254884, 0.96217430, 1.04515159, 1.04515183]

I'm actually unsure if the lower forward and adjoint eigenvalues should match, or if it should simply be the dominant mode. The adjoint fluxes also look reasonable. Perhaps there is some sort of synchronization issue in the ``GPUSolver`` in which a portion of the flux array is dirtied before it is copied to/from the GPU.

cjosey:
I was just about to post that I had the CPU solver working.  

If you treat it purely as a matrix math problem, then the eigenvalues of a transpose are equivalent to the eigenvalues of the original, which you can see at this [stackexchange link](https://math.stackexchange.com/questions/123923/a-matrix-and-its-transpose-have-the-same-set-of-eigenvalues).  The subtle variations are numerics and the fact that IRAM initializes the vector with a random one at the beginning of simulation (so as to ensure to non-zero probability that every eigenmode exists in the initial vector to ensure possibility of convergence).

Similarly, ARPACK does not return sorted eigenvalues, which is something I should probably fix for the output at some point.

cjosey:
So, I ran a quick test.  One thing that occurs with left (equivalently, eigenvectors of the transpose and thus adjoint) and right (normal) eigenvectors is that if you sort them by eigenvalue, form them into matrices, and transpose the left one so that you have X_L and X_R, the matrix product X_L X_R is necessarily diagonal ([mathematica link] (http://mathworld.wolfram.com/Eigenvector.html)).

Now, I don't have the computer horsepower to fully converge 2 C5G7 runs with eigs, but here's the result converged for 5 eigenmodes to 0.01:

```
[[  6.18528306e-01+0.j  -1.31521502e-03+0.j   8.94186960e-04+0.j  2.38844106e-04+0.j   5.17509226e-03+0.j]
 [ -1.27592206e-03+0.j  -6.24084651e-01+0.j   2.58791610e-04+0.j  8.55523511e-04+0.j  -3.55589174e-04+0.j]
 [ -2.12355866e-04+0.j   7.33355351e-04+0.j   6.24464869e-01+0.j  6.10202774e-02+0.j   3.49509879e-04+0.j]
 [  1.31630967e-03+0.j  -1.13755220e-03+0.j   6.86113816e-03+0.j  6.32287085e-01+0.j  -4.07282787e-04+0.j]
 [  5.18455682e-03+0.j  -4.95420070e-04+0.j   1.12200016e-03+0.j  4.42534685e-04+0.j  -6.82219565e-01+0.j]]
```
I'd call that diagonal.  It's not a sufficient condition to prove correctness, but at the very least it came out affirmative.

cjosey:
test
Here is a plot of a 20-eigenvector run converged to 10^-5.  It's a bit better, but there are also a few cases where it isn't diagonal and forms a block.  I suspect that these occur when two eigenvectors are exactly identical but rotated along a line of symmetry.  The eigenvalues should precisely equal one another, but due to precision, the last digits are random and will get sorted randomly.  Thus, what should be vector N and N+1 being multiplied by adjoint vector N and N+1 instead is multiplied to N+1 and N, forming the blocks.

This is just speculation though, so I hope someone else can figure out a better test to ensure the functionality of the adjoint.

wbinventor:
Very interesting @cjosey. Could you perhaps break the symmetry in the problem, perhaps use three MOX assemblies with on UO2 assembly?

cjosey:
I just found a better, and in theory more consistent, method to test this.  The [dot product test] (http://sepwww.stanford.edu/sep/prof/pvi/conj/paper_html/node9.html) basically states that given two random vectors, x and y, the following equation holds true:
y'Ax = (A'y)'x
I ran this test by first initializing the eigenvalue solvers such that ``_F_op`` is initialized in both adjoint and forward modes, and running the following code:
```
# Create random vectors
x = np.random.ranf(iram_forward._F_op.shape[0])
y_tilde = iram_forward._F_op * x

y = np.random.ranf(iram_adjoint._F_op.shape[0])
x_tilde = iram_adjoint._F_op * y

print("x_tilde * x", str(np.dot(x_tilde, x)))
print("y * y_tilde", str(np.dot(y, y_tilde)))
```
What I found is that this test is incredibly sensitive to numerics.  If run in single precision, it will not succeed for the F operator, and will be within 1% for the M and A operators.

If run in double precision, the accuracy of the M and A operators does not improve (this is odd), but the F operator, converged to 10^-15, improved a lot.  For several entirely different random vectors, the two values were within about 0.1%.  

Based on this knowledge, I'll eventually rerun the earlier, more computationally expensive test to machine precision convergence.  Then, I'll have the code swap all the pairs and output plots for all possible sets.  It'll take a very long time to do so, though, so don't expect this result for a day or two.

I suspect that we do not have a perfectly adjoint set of operators, as numerics are interfering.

wbinventor:
Thanks @cjosey for working on the correctness of the forward/adjoint eigenmodes. This is a mucb more thorough way or ensuring that the adjoint solver is working as promised than what I used (albeit, for all vacuum BCs). Definitely curious to hear what your new combinatorial test produces. It might actually be worthwhile to include a version of this test as a sample input for "krylov," what do you think?

I spent some time last night trying to fix the `GPUSolver` for adjoint eigenmodes. It doesn't appear to be a host/device synchronization issue, so I'll have to dig a little deeper. I think this is really the only major hurdle to cross with this PR, pending any other comments.

cjosey:
I continued testing.   I got repeated failures to actually prove functionality.  Then, I did a null test by duplicating the exact same problem twice, but swapping the forward and adjoint.  As far as I could tell, they were identical.

It turns out that, at least right now, if I simply make two CPUSolvers and two IRAMSolvers, they step on each other's memory and the end result is that both calculations are performed with whichever is initialized second.

Further, when I deliberately did them one at a time, the dot product test gave a negative result for the A and M operators.  I'll continue investigating, but either there is memory corruption by simply loading OpenMOC twice in the same file, something's wrong with the adjoint stuff, or something's wrong with the krylov stuff.

cjosey:
Ok, I've figured at least one problem out.  By switching the fission and scattering cross section indices, we get the A^T and M^T operators.  However, in linear eigenvalue calculation, we get the following problem:
```
Av = 1/k Mv
kAv = Mv
kv = (A^-1 M)v
kv = Fv
```
We want the transpose of the operator ``F``:
```
F^T
(A^-1 M)^T
M^T (A^-1)^T
M^T (A^T)^-1
```
For Krylov, we calculate ``(A^-1) Mv``, which needs to be converted to ``M^T(A^T)^-1v``.  I've put a PR on this on your repo.  I've yet to determine if the A^T and M^T operators are correct yet, but once I do for both the CPU and GPU, I can give the thumbs up. It appears that the A^T operator on the GPU does not match that on the CPU.

As a nice picture, here's (supposedly) the 4th group adjoint flux:
fsr-flux-group-4
This does not match the plot that comes out of the non-Krylov solver.

cjosey:
I'm sorry I've not spent much time on this.  @wbinventor, you basically know the gist of the situation.

I spoke with both Ben and Kord about this particular problem.  Ben made mention that although the transpose of indices is necessary, he suspects that there is a sign flip that does not cancel.  I suspect that it may cancel in the 1-D case, which is why it's frequently ignored in textbooks, but does not in 2-D.  Once my other things are dealt with, I'll sit down and give this a proper and thorough derivation, something worthy of being merged in as documentation.

wbinventor:
Thanks for the update @cjosey, I recognize that this is definitely a side project for you. I've poked at the GPU adjoint calculation and think I will have that resolved soon (see #200 as a starting point). I've also confirmed that the first five eigenvalues out of the ``IRAMSolver`` for both forward and adjoint cases match to at least 6 significant figures, albeit they are returned in an arbitrary order and must first be sorted in descending order. I believe you noted this in an earlier comment. It also appeared to me (at least to the naked eye) that the dominant adjoint eigenvector out of the ``IRAMSolver`` is very nearly the same as that computed with power iteration.

wbinventor:
I've fixed the ```GPUSolver`` for adjoint Krylov calculations. Now both CPU and GPU compute the same forward/adjoint eigenmodes. In addition, the dominant eigenmodes are the same as those computed using plain vanilla power iteration. I have not taken the time to fully explore the orthogonality (or correctness) of the adjoint eigenmodes as @cjosey has discussed in earlier comments, and likely won't devote any time to that at this point. I do think this PR should be fully reviewed (amd tested) by a third developer before merging. The vast majority of the changes in the diff are to the sample input files to address @geogunow's request, while the changes to the C++/CUDA source is relatively minor.

wbinventor:
This is now sync'ed up with the develop branch and is ready for final review and testing.

wbinventor:
Fyi @geogunow @samuelshaner perhaps a quick review and a trial run of the example inputs would be worthwhile, but this PR has been ready to go for quite some time.

src/CPUSolver.cpp
```diff
@@ -900,8 +901,9 @@ void CPUSolver::computeFSRFissionRates(double* fission_rates, int num_FSRs) {
   #pragma omp parallel for private (nu_sigma_f) schedule(guided)
```
samuelshaner: `volume` should be private as well.

src/accel/cuda/GPUSolver.cu
```diff
@@ -211,29 +211,26 @@ __global__ void computeFSRFissionSourcesOnDevice(int* FSR_materials,
...
-    /* Compute total fission source for current FSR */
-    for (int g_prime=0; g_prime < *num_groups; g_prime++)
-      fission_source += scalar_flux(tid,g_prime) * nu_sigma_f[g_prime];
+      for (int g_prime=0; g_prime < *num_groups; g_prime++)
+        fission_source += fiss_mat[g*(*num_groups)+g_prime] * scalar_flux(tid,g_prime);
 
-    /* Set the reduced source for FSR r in each group g */
-    for (int g=0; g < *num_groups; g++) {
-      reduced_sources(tid,g) = fission_source * chi[g];
+      /* Compute total (scatter+fission+fixed) reduced source */
```
samuelshaner: It looks like this is just computing the fission reduced source. Is that correct?

sample-input/simple-lattice/simple-lattice.py
```diff
@@ -17,33 +12,33 @@
...
 boundaries = [left, right, bottom, top]
 
-large_circle = Circle(x=0.0, y=0.0, radius=0.4, name='large pin')
-medium_circle = Circle(x=0.0, y=0.0, radius=0.3, name='medium pin')
-small_circle = Circle(x=0.0, y=0.0, radius=0.2, name='small pin')
+large_circle = openmoc.Circle(x=0.0, y=0.0, radius=0.4, name='large pin')
+medium_circle = openmoc.Circle(x=0.0, y=0.0, radius=0.3, name='medium pin')
+small_circle = openmoc.Circle(x=0.0, y=0.0, radius=0.2, name='small pin')
 
 for boundary in boundaries: boundary.setBoundaryType(PERIODIC)
```
samuelshaner: The BC should be changed to `openmoc.PERIODIC`.

samuelshaner:
I'm curious, is there any way to indicate how progress on the `adjoint-eigenmodes.py` simulation? If the number of sweeps on the A or M operator is predetermined, I feel like it would be helpful if OpenMOC printed out `Performed A operator sweep number --- of 500` or something.

sample-input/benchmarks/LRA/geometry.py
```diff
@@ -0,0 +1,201 @@
...
+
+num_threads = options.getNumThreads()
+track_spacing = options.getTrackSpacing()
+num_azim = options.getNumAzimAngles()
+tolerance = options.getTolerance()
+max_iters = options.getMaxIterations()
+
+openmoc.log.set_log_level('NORMAL')
+
+openmoc.log.py_printf('TITLE', 'Simulating the LRA Benchmark Problem...')
```
samuelshaner: The title gets printed twice since it is also in the main input file so I'd take this out.

samuelshaner:
This PR looks good to merge once the items are addressed.

src/CPUSolver.cpp
wbinventor: Good catch!

src/accel/cuda/GPUSolver.cu
wbinventor: That is right, I've denoted this in the doxygen comment now. Do you want me to change the method signature?

sample-input/simple-lattice/simple-lattice.py
wbinventor: Good catch, thanks.

sample-input/benchmarks/LRA/geometry.py
wbinventor: I removed it.

src/accel/cuda/GPUSolver.cu
samuelshaner: I think the message signature is fine. I would just recommend changing the comment on line 232 since it includes scatter and fixed source in it.

wbinventor:
@samuelshaner all of your comments have been addressed save for the Krylov solver's progress reporting. There isn't an immediately straightforward method to do this since the Krylov solver is buried within SciPy's ``eigs`` routine. I think this could be opened as an issue and dealt with in a future PR if the need persists.

src/accel/cuda/GPUSolver.cu
wbinventor: Done.

samuelshaner:
Everything looks good to merge. I'll add the Krylov solver progress as an issue.

PR #188, opened 2015-07-27
Modifications to support 3D geometry features

samuelshaner:
This PR makes changes to many of the objects to support 3D geometry features. In particular, the z-component is added to points and lattices and the polar angle is introduced in functions to propagate the tracks forward in the z-direction. Surfaces and boundary conditions are added in the z-direction, but are required to be REFLECTIVE since they are not actually used in the Solver. The CMFD objects also is organized for a 3D mesh, but the z-direction is required to be of dimension 1 and have REFLECTIVE BCs. Many 3D features were intentionally left out of this PR such as reconfiguring the PolarQuad to support different polar angles/weights for each azimuthal angle.

This PR depends on PRs #182, #184, and #185 and will be updated as those are modified.


geogunow:
@samuelshaner let me know if the issue in the 3D MOC branch would be present here as well

openmoc/map_to_dict.i
```diff
@@ -96,10 +96,10 @@
 }
 
 
-/* Typemap for Lattice::setUniverses(int num_x, int num_y, Universe** universes)
+/* Typemap for Lattice::setUniverses(int num_y, int num_x, Universe** universes)
```
geogunow: Ok, I'm a bit confused here. It looks intentional so you're probably correct, but at a first glance it looks like the old line was correct.

src/CPUSolver.cpp
```diff
@@ -530,13 +530,13 @@ void CPUSolver::transportSweep() {
   if (_cmfd != NULL && _cmfd->isFluxUpdateOn())
     _cmfd->zeroSurfaceCurrents();
 
-  /* Loop over azimuthal angle halfspaces */
-  for (int i=0; i < 2; i++) {
+  /* Loop over azimuthal angle and periodic track halfspaces */
+  for (int i=0; i < 4; i++) {
```
geogunow: Is this specific to 2D or generalization to 3D? If it's specific to 2D I would like to see a variable rather than the number 4 to make everything more translatable to 3D.

src/CPUSolver.cpp
```diff
@@ -663,24 +665,22 @@ void CPUSolver::transferBoundaryFlux(int track_id,
...
-    start = _tracks[track_id]->isReflIn() * _polar_times_groups;
-    bc = (int)_tracks[track_id]->getBCIn();
+    start = _tracks[track_id]->isNextIn() * _polar_times_groups;
+    bc = std::min((int)_tracks[track_id]->getBCIn(), 1);
     track_out_id = _tracks[track_id]->getTrackIn()->getUid();
   }
 
   FP_PRECISION* track_out_flux = &_boundary_flux(track_out_id,0,0,start);
-
+
```
geogunow: Remove space

src/CPUSolver.cpp
```diff
@@ -663,24 +665,22 @@ void CPUSolver::transferBoundaryFlux(int track_id,
...
-  /* Extract boundary conditions for this Track */
-
   /* For the "forward" direction */
   if (direction) {
-    start = _tracks[track_id]->isReflOut() * _polar_times_groups;
-    bc = (int)_tracks[track_id]->getBCOut();
+    start = _tracks[track_id]->isNextOut() * _polar_times_groups;
+    bc = std::min((int)_tracks[track_id]->getBCOut(), 1);
     track_out_id = _tracks[track_id]->getTrackOut()->getUid();
   }
```
geogunow: Not for this PR, but in the future I would like to see an abstraction for direction

src/Cmfd.h
```diff
@@ -70,29 +63,28 @@ class Cmfd {
...
+  Vector* _new_flux;
 
   /** Vector representing the flux for each cmfd cell and cmfd enegy group at 
    * the beginning of a CMFD solve */
-  FP_PRECISION* _old_flux;
-
-  /** Vector representing the flux during the previous iteration of a 
-   * cmfd solve */
-  FP_PRECISION* _flux_temp;
+  Vector* _old_flux;
```
geogunow: Ok, I just thought of something - why are we implementing our own vector class? What does this Vector class do that std::vector cannot?

src/Geometry.cpp
```diff
@@ -1069,26 +1081,26 @@ void Geometry::initializeCmfd() {
   double width = getWidth();
   double cell_width = width / num_x;
   double cell_height = height / num_y;
-
+
```
geogunow: Remove space

src/Material.cpp
```diff
@@ -279,9 +279,9 @@ FP_PRECISION* Material::getDifHat() {
 
   if (_dif_hat == NULL) {
 
-    _dif_hat = new FP_PRECISION[4*_num_groups];
+    _dif_hat = new FP_PRECISION[6*_num_groups];
 
-    for (int e = 0; e < 4*_num_groups; e++)
+    for (int e = 0; e < 6*_num_groups; e++)
```
geogunow: What about 2D problems - does changing from 4 to 6 affect them? I assume not, but why?

src/Surface.cpp
```diff
@@ -850,12 +902,27 @@ int Circle::intersection(Point* point, double angle, Point* points) {
...
-      points[num].setCoords(xcurr, ycurr);
-      if (angle < M_PI && ycurr > y0)
-        num++;
-      else if (angle > M_PI && ycurr < y0)
-        num++;
+      ycurr = y0 + m * (points[num].getX() - x0);
+      zcurr = z0 + sqrt(pow(ycurr - y0, 2.0) + pow(xcurr - x0, 2.0))
+        * tan(M_PI_2 - polar);
+      points[num].setCoords(xcurr, ycurr, zcurr);
+      if (azim < M_PI && ycurr > y0){
```
geogunow: spacing

src/Surface.cpp
```diff
@@ -798,31 +806,75 @@ int Circle::intersection(Point* point, double angle, Point* points) {
...
-        num++;
-      else if (angle > M_PI && ycurr < y0)
-        num++;
-      return num;
+      zcurr = z0 + sqrt(pow(ycurr - y0, 2.0) + pow(xcurr - x0, 2.0))
+        * tan(M_PI_2 - polar);
+      points[num].setCoords(xcurr, ycurr, zcurr);
+
+      /* Check that point is in same direction as angle */
+      if (azim < M_PI && ycurr > y0){
```
geogunow: spacing

src/Surface.cpp
```diff
@@ -798,31 +806,75 @@ int Circle::intersection(Point* point, double angle, Point* points) {
...
+        if (zcurr > z0 && polar < M_PI_2)
+          num++;
+        else if (zcurr < z0 && polar > M_PI_2)
+          num++;
+        else if (fabs(zcurr - z0) < 1.e-10 && fabs(polar - M_PI_2) < 1.e-10)
+          num++;
+      }
+      else if (azim > M_PI && ycurr < y0){
```
geogunow: ditto

src/Surface.cpp
```diff
@@ -798,31 +806,75 @@ int Circle::intersection(Point* point, double angle, Point* points) {
...
       ycurr = (-b + sqrt(discr)) / (2 * a);
-      points[num].setCoords(xcurr, ycurr);
-      if (angle < M_PI && ycurr > y0)
-        num++;
-      else if (angle > M_PI && ycurr < y0)
-        num++;
+      zcurr = z0 + sqrt(pow(ycurr - y0, 2.0) + pow(xcurr - x0, 2.0))
+        * tan(M_PI_2 - polar);
+      points[num].setCoords(xcurr, ycurr, zcurr);
+      if (azim < M_PI && ycurr > y0){
```
geogunow: Spacing. I'll stop with these comments - just search for ``){`` and replace with ``) {``

src/TrackGenerator.cpp
```diff
@@ -491,6 +528,7 @@ void TrackGenerator::generateTracks() {
     /* Allocate memory for the Tracks */
     try {
       _num_tracks = new int[_num_azim];
+      _num_tracks_by_halfspace = new int[5];
```
geogunow: Why 5?

geogunow:
I just finished my first review of this code. Looks pretty good - just a few stylistic comments.

openmoc/map_to_dict.i
samuelshaner: There are three locations where the setUniverses method needs to be defined: map_to_dict.i, Universe.h, and Universe.cpp. There was an inconsistency in the current develop branch on the method definitions, so this change and a change in Universe.h makes them consistent.

src/CPUSolver.cpp
samuelshaner: This is specific to 2D. 3D requires 16 iterations. I'll change to a variable.

src/Cmfd.h
samuelshaner: This vector class is convenient because it allows us to set values given a cell and energy group, which makes things a bit cleaner. Previously, one would have to compute the index by multiplying the cell by the number of groups each time we want to set/increment a value. This vector class also has thread-safe set/increments, so it is kind of like your replacement of the unordered_maps with the parallel hash maps.

src/Material.cpp
samuelshaner: With this update, we are now solving CMFD in 3D. The z-height of the cmfd lattice is just required to be 1 and the z-max and z-min BCs are required to be reflective. This effectively makes the problem 2D.

src/TrackGenerator.cpp
samuelshaner: There are 4 halfspaces and it is easiest to have the index for the beginning and end of each half space stored, so hence 5 values. I created a variable for the number of independent transport sweep loops required, so I'll change this to that value plus 1.

samuelshaner:
I have addressed the comments and merged with all dependent branches. This PR should be up to date, but should wait to be merged/further reviewed until after PRs #182 #184 and #185 are merged.

samuelshaner:
I have updated this PR and in doing so, tried to include the minimum amount of modifications to achieve the desired goal: to allow users to input a 3D geometry and performed a 2D MOC simulation on one of the xy-planes.

Important aspects of this PR include:
- Lattices are now 3D instead of 2D. 2D lattices can still be input as before, but they will just be assumed to be infinite in the z-direction.
- Points and LocalCoords objects are now assumed to be in 3D.
- Tracks are propagated forward in only x and y. The z-dimension is assumed to stay constant.
- CMFD is kept in 2D; I contemplated extending it to 3D, but it adds quite a bit of code that is unnecessary for running any simulations with this version of OpenMOC.
- Plots of materials and cells can now be done at different z-levels; However, plots are restricted to the xy-plane. Adding the ability to plot materials and cells in the xz and yz planes can be implemented if others find it useful.
- A 3D C5G7 benchmark problem was added as an example of inputting a 3D geometry and running a 2D simulation at a desired z-level.

The 3D C5G7 rodded B case was added as a sample input file to demonstrate the input of a 3D lattice geometry and 2D simulation at a desired z-level. The plots below (taken from Brendan Kocunas' thesis) show the C5G7 rodded B configuration with distinct z-levels from (-32.13, -17.85), (-17.85, -3.57), (-3.57, 10.71), and (10.71, 32.13). The top z-level contains no fissionable materials, so it cannot be run. All other planes can be run and produce the desired results. 

c5g7_rodded_b

The plots below show the subtle differences in the materials and fluxes between the different levels. From left to right, the plots represent the materials and thermal fluxes plotted a z-levels of (z = 10.0, ), (z = -10.0), and (z = -20.0).

c5g7-rodded-b-z-levels-w-flux

This PR begs for the documentation to be updated to describe 3D geometry input, not to mention the documentation updates to describe k-nearest neighbor flux updating. The KNN documentation is coming along and I'll put a PR in for that soon. The 3D geometry documentation update will be a task for some weekend in October after this PR is merged.

samuelshaner:
@geogunow and @wbinventor - I just updated this PR so it is ready for your further review.

wbinventor:
Thanks for putting this together @samuelshaner! I'll start reviewing this over the next few days. Hopefully @geogunow can also review it since he has probably seen more of these features before than I have.

openmoc/plotter.py
```diff
@@ -225,12 +229,22 @@ def plot_materials(geometry, gridsize=250, xlim=None, ylim=None):
...
     py_printf('ERROR', 'Unable to plot the Materials since ' + \
-              'since the gridsize %d is not an integer', gridsize)
+              'the gridsize %d is not an integer', gridsize)
 
   if gridsize <= 0:
     py_printf('ERROR', 'Unable to plot the Materials ' + \
               'with a negative gridsize (%d)', gridsize)
-
+    
+  if z_level is None:
```
geogunow: Sometimes I've noticed differences with "is" and "==", not sure if it makes a difference here

sample-input/benchmarks/c5g7/c5g7-3d-rodded-B.py
```diff
@@ -0,0 +1,526 @@
...
+zmax = openmoc.ZPlane(z= 32.13, name='zmax')
+
+xmin.setBoundaryType(openmoc.REFLECTIVE)
+xmax.setBoundaryType(openmoc.VACUUM)
+ymin.setBoundaryType(openmoc.VACUUM)
+ymax.setBoundaryType(openmoc.REFLECTIVE)
+
+# Create Circles for the fuel as well as to discretize the moderator into rings
+fuel_radius = openmoc.Circle(x=0.0, y=0.0, radius=0.54)
+moderator_inner_radius = openmoc.Circle(x=0.0, y=0.0, radius=0.62)
```
geogunow: fix inner to be less than outer

src/Point.h
```diff
@@ -49,9 +48,10 @@ class Point {
  * @param x x-coordinate
  * @param y y-coordinate
  */
-inline void Point::setCoords(const double x, const double y) {
-  _x = x;
-  _y = y;
+inline void Point::setCoords(const double x, const double y, const double z) {
+  _xyz[0] = x;
+  _xyz[1] = y;
+  _xyz[2] = z;
```
geogunow: Is there a reason we've gone to _xyz rather than _x and _y and _z?

src/Surface.cpp
```diff
@@ -340,53 +351,36 @@ inline int Plane::intersection(Point* point, double angle, Point* points) {
...
-      /* Check that point is in same direction as angle */
-      if (angle < M_PI && ycurr > y0)
-        num++;
-      else if (angle > M_PI && ycurr < y0)
-        num++;
-      return num;
-    }
+  double xcurr, ycurr, zcurr; /* coordinates of current intersection point */
+
+  /* The track and plane are parallel */
```
geogunow: Is this comment supposed to be on this line?

src/Surface.cpp
```diff
@@ -800,19 +803,22 @@ int Circle::intersection(Point* point, double angle, Point* points) {
...
-      points[num].setCoords(xcurr, ycurr);
+      zcurr = z0;
+      points[num].setCoords(xcurr, ycurr, zcurr);
+
+      /* Check that point is in same direction as angle */
       if (angle < M_PI && ycurr > y0)
         num++;
       else if (angle > M_PI && ycurr < y0)
-        num++;
-      return num;
```
geogunow: why did you remove the return here?

src/TrackGenerator.cpp
```diff
@@ -429,19 +429,19 @@ bool TrackGenerator::containsTracks() {
  *
  * @code
  *          num_tracks = track_generator.getNumTracks()
- *          coords = track_generator.retrieveTrackCoords(num_tracks*4)
+ *          coords = track_generator.retrieveTrackCoords(num_tracks*6)
```
geogunow: Are we always returning 3D coordinates or should the 4 and 6 be replaced with something like NUM_SURFACES

src/TrackGenerator.cpp
```diff
@@ -469,22 +471,22 @@ void TrackGenerator::retrieveTrackCoords(double* coords, int num_tracks) {
...
+  if (num_segments != 7*getNumSegments())
     log_printf(ERROR, "Unable to retrieve the Track segment coordinates since "
                "the TrackGenerator contains %d segments with %d coordinates "
                "but an array of length %d was input",
-               getNumSegments(), 5*getNumSegments(), num_segments);
+               getNumSegments(), 7*getNumSegments(), num_segments);
 
   segment* curr_segment = NULL;
-  double x0, x1, y0, y1;
+  double x0, x1, y0, y1, z;
```
geogunow: It looks to me like there is an inconsistency here. I understand that you just have ``z`` instead of ``z0`` and ``z1`` because it is just a 2D slide at a given z-height, but you are returning an array that has two extra variable slots (you changed from 5 to 7). Perhaps we should just include ``z0`` and ``z1`` to make this consistent and make the transition to 3D a bit smoother.

src/Universe.cpp
```diff
@@ -893,7 +877,7 @@ Universe* Lattice::getUniverse(int lat_x, int lat_y) const {
  * @brief Return a 2D vector of the Universes in the Lattice.
  * @return 2D vector of Universes
  */
-std::vector<std::vector<std::pair<int, Universe*>>>
+std::vector< std::vector<std::vector<std::pair<int, Universe*>>>>
```
geogunow: Remove the extra space

src/Universe.cpp
```diff
@@ -1050,20 +1046,93 @@ void Lattice::setUniverses(int num_y, int num_x, Universe** universes) {
...
+  setNumZ(num_z);
+
+  Universe* universe;
+
+  /* The Lattice cells are assumed input in row major order starting from the
+   * upper left corner. This double loop reorders the Lattice cells from the
+   * to start from the lower left corner */
+  for (int k = 0; k < _num_z; k++) {
+    _universes.push_back
+      (std::vector< std::vector< std::pair<int, Universe*> > >());
```
geogunow: make sure style is consistent on nested vectors

src/Universe.cpp
```diff
@@ -1197,40 +1286,37 @@ Cell* Lattice::findCell(LocalCoords* coords) {
  *          in the direction of the track.
  *          Returns distance to nearest Lattice cell boundary.
  * @param point a pointer to a starting point
- * @param angle the azimuthal angle of the track
+ * @param angle the angle of the track
```
geogunow: Not the azimuthal angle?

src/constants.h
```diff
@@ -68,16 +68,16 @@
...
  */
 #define NUM_SURFACES 4
 #define SURFACE_X_MIN 0
-#define SURFACE_Y_MIN 1
-#define SURFACE_X_MAX 2
+#define SURFACE_X_MAX 1
+#define SURFACE_Y_MIN 2
 #define SURFACE_Y_MAX 3
 
 #define NUM_CORNERS 4
```
geogunow: What about for 3D MOC?

geogunow:
I just finished my review of the code. A fresh set of eyes would be nice. I think this is close to being able to merge.

openmoc/map_to_dict.i
```diff
@@ -107,15 +107,15 @@
     return NULL;
   }
 
-  $1 = PySequence_Length($input);  // num_x
-  $2 = PySequence_Length(PyList_GetItem($input,0)); // num_y
+  $1 = PySequence_Length($input);  // num_y
```
wbinventor: Was this wrong all along?

openmoc/map_to_dict.i
```diff
@@ -126,13 +126,65 @@
...
+                    "for the Lattice cells");
+    return NULL;
+  }
+
+  $1 = PySequence_Length($input);  // num_z
+  $2 = PySequence_Length(PyList_GetItem($input,0)); // num_y
+  $3 = PySequence_Length(PyList_GetItem(PyList_GetItem($input,0), 0)); // num_x
+  $4 = (Universe**) malloc(($1 * $2 * $3) * sizeof(Universe*)); // universes
+
+  /* Loop over y */
```
wbinventor: I think this comment should be z not y

openmoc/map_to_dict.i
```diff
@@ -126,13 +126,65 @@
...
       void *p1 = 0;
       SWIG_ConvertPtr(o, &p1, SWIGTYPE_p_Universe, 0 | 0);
-      $3[i*$2+j] = (Universe*) p1;
+      $3[j*$2+i] = (Universe*) p1;
+    }
+  }
+}
+
+
+/* Typemap for Lattice::setUniverses3D(int num_z, int num_y, int num_x,
```
wbinventor: Perhaps it would be cleaner to require all lattices to be 3D so that we simply have a single `Lattice::setUniverses(...)` routine? This would require adding an extra level of brackets around the 2D lattice universe lists to make them 3D in all input files. What do you think @samuelshaner @geogunow?

openmoc/map_to_dict.i
```diff
@@ -126,13 +126,65 @@
...
+    PyObject* outer_outer_list = PyList_GetItem($input,k);
+    
+    /* Loop over y */
+    for (int j = 0; j < $2; j++) {
+
+      /* Get the inner list in the nested list for the lattice */
+      PyObject* outer_list = PyList_GetItem(outer_outer_list, j);
+
+      /* Check that the length of this list is the same as the length
+       * of the first list */
```
wbinventor: should be "second list" rather than "first list"?

openmoc/map_to_dict.i
```diff
@@ -126,13 +126,65 @@
...
+  /* Loop over y */
+  for (int k = 0; k < $1; k++) {
+
+    /* Get the inner list in the nested list for the lattice */
+    PyObject* outer_outer_list = PyList_GetItem($input,k);
+    
+    /* Loop over y */
+    for (int j = 0; j < $2; j++) {
+
+      /* Get the inner list in the nested list for the lattice */
```
wbinventor: "inner list" should be "second list"?

openmoc/plotter.py
```diff
@@ -78,12 +78,12 @@ def plot_tracks(track_generator):
   num_azim = track_generator.getNumAzim()
   spacing = track_generator.getTrackSpacing()
   num_tracks = track_generator.getNumTracks()
-  coords = track_generator.retrieveTrackCoords(num_tracks*4)
+  coords = track_generator.retrieveTrackCoords(num_tracks*6)
```
wbinventor: I really think that this "6" should be defined as a variable to make it transparent

openmoc/plotter.py
```diff
@@ -143,20 +143,23 @@ def plot_segments(track_generator):
   spacing = track_generator.getTrackSpacing()
   num_segments = track_generator.getNumSegments()
   num_fsrs = track_generator.getGeometry().getNumFSRs()
-  coords = track_generator.retrieveSegmentCoords(num_segments*5)
+  coords = track_generator.retrieveSegmentCoords(num_segments*7)
```
wbinventor: Ditto, define "7" as some variable

openmoc/plotter.py
```diff
@@ -143,20 +143,23 @@ def plot_segments(track_generator):
...
+  z = numpy.zeros(num_segments*2)
   fsrs = numpy.zeros(num_segments)
 
   for i in range(num_segments):
-    fsrs[i] = coords[i*5]
-    x[i*2] = coords[i*5+1]
-    y[i*2] = coords[i*5+2]
-    x[i*2+1] = coords[i*5+3]
-    y[i*2+1] = coords[i*5+4]
+    fsrs[i] = coords[i*7]
```
wbinventor: Ditto, use whatever variables you define for "7" above

openmoc/plotter.py
```diff
@@ -179,13 +182,13 @@ def plot_segments(track_generator):
   plt.xlim([x.min(), x.max()])
   plt.ylim([y.min(), y.max()])
 
-  title = 'Segments for ' + str(num_azim) + ' angles and ' + str(spacing) + \
-        ' cm spacing'
+  title = 'Segments for ' + str(num_azim) + ' angles, ' + str(spacing) + \
+        ' cm spacing, and ' + str(z[0]) + ' z-level'
```
wbinventor: Perhaps the z-level could be a [subtitle](http://matplotlib.sourcearchive.com/documentation/0.98.3/classmatplotlib_1_1figure_1_1Figure_c5a0bfb9075702f39846bad94aee5463.html)?

openmoc/plotter.py
samuelshaner: The convention for comparing with `None` seems to be with `is` and not `==`, but I don't see any strict guidance in the PEP8 guidelines.

openmoc/plotter.py
wbinventor: Also, I don't know about the notation you and @geogunow are using for "z levels", but this seems a bit confusing to me. Is the "z[0]" here an integer index for some "z level?" Or is it (as I expect) the z-coordinate for the *xy*-plane plotted here? If it's the latter, perhaps "z-str(z[0])" would be more accurate than "str(z[0])-z-level"?

openmoc/plotter.py
```diff
@@ -179,13 +182,13 @@ def plot_segments(track_generator):
...
 
   plt.title(title)
 
   filename = directory + 'segments-' + str(num_azim) + '-angles-' + \
-      str(spacing) + '-spacing.png'
+      str(spacing) + '-spacing-' + str(z[0]) + '-z-level.png'
```
wbinventor: Ditto, if it's agreed that "z-str(z[0])" is more appropriate then use it here as well.

openmoc/plotter.py
```diff
@@ -208,7 +211,8 @@ def plot_segments(track_generator):
 # @param gridsize an optional number of grid cells for the plot
 # @param xlim optional list/tuple of the minimim/maximum x-coordinates
 # @param ylim optional list/tuple of the minimim/maximum y-coordinates
-def plot_materials(geometry, gridsize=250, xlim=None, ylim=None):
+# @param z_level optional the z level (default is Geometry's _z_level)
```
wbinventor: Typo "_z_level" should be "z_level"

openmoc/plotter.py
```diff
@@ -208,7 +211,8 @@ def plot_segments(track_generator):
...
+def plot_materials(geometry, gridsize=250, xlim=None, ylim=None, z_level=None):
```
wbinventor: If `z_level` is actually a coordinate rather than a level index, perhaps `zcoord` would be more appropriate? I won't make this comment in the code below since it looks like it `z_level` is the notation used everywhere right now.

openmoc/plotter.py
wbinventor: Yeah, I think `is` is the more common notation from what I've seen (and from the more formal Python code standards we are adhering to in OpenMC).

openmoc/plotter.py
```diff
@@ -321,12 +336,22 @@ def plot_cells(geometry, gridsize=250, xlim=None, ylim=None):
...
   if not is_integer(gridsize):
     py_printf('ERROR', 'Unable to plot the Cells since ' + \
-                'since the gridsize %d is not an integer', gridsize)
+                'the gridsize %d is not an integer', gridsize)
 
   if gridsize <= 0:
     py_printf('ERROR', 'Unable to plot the Cells ' + \
               'with a negative gridsize (%d)', gridsize)
 
+  if z_level is None:
```
wbinventor: While going through this I've realized just how much redundant error checking code we have in these plotting routines. I recognize this isn't something that must be solved in this PR, but if you are going to change `z_level` to `zcoord` you might consider extracting and generalizing a lot of this error checking to some sort of (private) "openmoc.plotter._check_input(gridsize, xlim, ylim, zcoord)` routine.

openmoc/process.py
```diff
@@ -246,6 +247,7 @@ def store_simulation_state(solver, fluxes=False, sources=False,
   num_FSRs = geometry.getNumFSRs()
   num_materials = geometry.getNumMaterials()
   num_groups = geometry.getNumEnergyGroups()
+  z_level = geometry.getZLevel()
```
wbinventor: Ditto `z_level` notation here

openmoc/thisown.i
```diff
@@ -90,12 +90,12 @@
 /* A Lattice owns the memory for each Universe it contains */
 %pythonappend Lattice::setUniverses %{
   # SWIG 3
-  if 'num_x' in locals():
-    universes = locals()['num_x']
-  elif 'args' in locals() and 'num_x' in locals()['args']:
-    universes = locals()['args']['num_x']
-  elif 'kwargs' in locals() and 'num_x' in locals()['kwargs']:
-    universes = locals()['kwargs']['num_x']
+  if 'num_y' in locals():
```
wbinventor: If we decide to move to a streamlined 3D lattice universes input, make sure to update this

openmoc/thisown.i
```diff
@@ -106,6 +106,26 @@
       universes[i][j].thisown = 0
 %}
 
+/* A Lattice owns the memory for each Universe it contains */
+%pythonappend Lattice::setUniverses3D %{
+  # SWIG 3
```
wbinventor: Just curious, has this been tested with both SWIG 2 and 3? If not, I have both installed and can quickly test it out.

src/Surface.cpp
samuelshaner: Nope, it should be before the if statement.

sample-input/benchmarks/c5g7/c5g7-3d-rodded-B.py
```diff
@@ -0,0 +1,526 @@
+import openmoc
+
+###############################################################################
```
wbinventor: Can this be turned into a "geometry3d.py" file which can be imported into "c5g7.py", "c5g7-gpu.py" or "c5g7-cmfd.py"? Just doesn't seem to align with the format we have switched to which minimized code reuse by abstracting the geometry construction into a single shared and importable Python module.

Or perhaps you could take this geometry construction code and append it to "geometry.py" and name the two different `Geometry` objects as `geometry2d` and `geometry3d`, either of which could be imported from `geometry` and added to a `TrackGenerator`.

sample-input/benchmarks/c5g7/c5g7-3d-rodded-B.py
```diff
@@ -0,0 +1,526 @@
...
+num_threads = options.getNumThreads()
+azim_spacing = options.getTrackSpacing()
+num_azim = options.getNumAzimAngles()
+tolerance = options.getTolerance()
+max_iters = options.getMaxIterations()
+
+openmoc.log.set_log_level('NORMAL')
+
+openmoc.log.py_printf('TITLE', \
+                      'Simulating the OECD\'s C5G7 Benchmark Problem...')
```
wbinventor: Perhaps "C5G7 3D"?

src/Surface.cpp
samuelshaner: Good call, I think the return should still be there.

sample-input/benchmarks/c5g7/c5g7-3d-rodded-B.py
```diff
@@ -0,0 +1,526 @@
...
+
+
+###############################################################################
+###########################   Creating Materials   ############################
+###############################################################################
+
+openmoc.log.py_printf('NORMAL', 'Importing materials data from py...')
+
+materials = openmoc.materialize.materialize('../../c5g7-materials.py')
```
wbinventor: Using the Python dictionary of materials rather than HDF5 for any particular reason? The Python dict support for materials is really only there for weird machines which don't support `h5py`

sample-input/benchmarks/c5g7/c5g7-3d-rodded-B.py
```diff
@@ -0,0 +1,526 @@
...
+xmax = openmoc.XPlane(x= 32.13, name='xmax')
+ymin = openmoc.YPlane(y=-32.13, name='ymin')
+ymax = openmoc.YPlane(y= 32.13, name='ymax')
+zmin = openmoc.ZPlane(z=-32.13, name='zmin')
+zmax = openmoc.ZPlane(z= 32.13, name='zmax')
+
+xmin.setBoundaryType(openmoc.REFLECTIVE)
+xmax.setBoundaryType(openmoc.VACUUM)
+ymin.setBoundaryType(openmoc.VACUUM)
+ymax.setBoundaryType(openmoc.REFLECTIVE)
```
wbinventor: Are there not boundary conditions on the `ZPlanes`?

sample-input/benchmarks/c5g7/c5g7-3d-rodded-B.py
```diff
@@ -0,0 +1,526 @@
...
+reflector_cell.setFill(materials['Water'])
+
+reflector = openmoc.Universe(name='Reflector')
+reflector.addCell(reflector_cell)
+
+refined_reflector_cell = openmoc.Cell(name='Semi-Finely Spaced Reflector')
+refined_reflector = openmoc.Universe(name='Semi-Finely Spaced Moderator')
+refined_reflector.addCell(refined_reflector_cell)
+
+# Cells
```
wbinventor: Perhaps elaborate as "Lattice-filled Cells"

sample-input/benchmarks/c5g7/c5g7-3d-rodded-B.py
```diff
@@ -0,0 +1,526 @@
...
+lattices[-1].setUniverses(template)
+assembly_rfl_unrod_cell_cnr.setFill(lattices[-1])
+
+
+###############################################################################
+#########################   Creating Core Lattice   ###########################
+###############################################################################
+
+# 3 x 3 x 9 core to represent 3D core
+lattices.append(openmoc.Lattice(name='Full Geometry'))
```
wbinventor: Can this be shortened as:

```python
lattices[-1].setUniverses
([[[assembly_rfl_rod      , assembly_rfl_rod      , assembly_rfl_unrod_rgt],
   [assembly_rfl_rod      , assembly_rfl_rod      , assembly_rfl_unrod_rgt],
   [assembly_rfl_unrod_btm, assembly_rfl_unrod_btm, assembly_rfl_unrod_cnr]]*3,
  [[assembly_uo2_rod      , assembly_mox_rod      , assembly_rfl_unrod_rgt],
   [assembly_mox_rod      , assembly_uo2_unrod    , assembly_rfl_unrod_rgt],
   [assembly_rfl_unrod_btm, assembly_rfl_unrod_btm, assembly_rfl_unrod_cnr]]*2,
  [[assembly_uo2_unrod    , assembly_mox_unrod    , assembly_rfl_unrod_rgt],
   [assembly_mox_unrod    , assembly_uo2_unrod    , assembly_rfl_unrod_rgt],
   [assembly_rfl_unrod_btm, assembly_rfl_unrod_btm, assembly_rfl_unrod_cnr]]*4])
```

Again, this would assume we stick with two different ``Lattice::setUniverses(...)`` routines.

sample-input/benchmarks/c5g7/c5g7-3d-rodded-B.py
```diff
@@ -0,0 +1,526 @@
...
+
+###############################################################################
+##########################   Creating the Geometry   ##########################
+###############################################################################
+
+openmoc.log.py_printf('NORMAL', 'Creating geometry...')
+
+geometry = openmoc.Geometry()
+geometry.setRootUniverse(root_universe)
+geometry.setZLevel(-20.0)
```
wbinventor: Perhaps this might be changed to `Geometry:setZCoord(...)`? I'm not sure if this is the best notation, but the term "level" connotes some integer index to me, esp. considering the features that are coming down the pipeline.

sample-input/benchmarks/c5g7/c5g7-3d-rodded-B.py
```diff
@@ -0,0 +1,526 @@
...
+geometry.setCmfd(cmfd)
+geometry.initializeFlatSourceRegions()
+
+###############################################################################
+########################   Creating the TrackGenerator   ######################
+###############################################################################
+
+openmoc.log.py_printf('NORMAL', 'Initializing the track generator...')
+
+track_generator = openmoc.TrackGenerator(geometry, num_azim, azim_spacing)
```
wbinventor: If you decide to abstract the 3D geometry construction to "geometry.py" as some new "geometry3d" variable, then perhaps rename this file and keep these last 100 lines or so as a "c5g7-3d-slice.py" or something?

sample-input/benchmarks/c5g7/c5g7-3d-rodded-B.py
wbinventor: Is there a default z level in the geometry? I guess I'll find out more about this as I move on down.

src/Cell.cpp
```diff
@@ -64,6 +64,12 @@ Cell::Cell(int id, const char* name) {
   _max_y = std::numeric_limits<double>::infinity();
   _min_z = -std::numeric_limits<double>::infinity();
   _max_z = std::numeric_limits<double>::infinity();
+
+  /* Set the default boundaries to be REFLECTIVE */
+  _min_x_bc = REFLECTIVE;
+  _max_x_bc = REFLECTIVE;
+  _min_y_bc = REFLECTIVE;
+  _max_y_bc = REFLECTIVE;
```
wbinventor: I assume you don't want to include BCs for the z-axis here since that will come in a later PR with full 3D support?

src/Geometry.cpp
```diff
@@ -20,6 +20,9 @@ Geometry::Geometry() {
 
   /* Initialize CMFD object to NULL */
   _cmfd = NULL;
+
+  /* Initialize the z-level to zero */
+  _z_level = 0.0;
```
wbinventor: Ok, I see the default here.

src/Geometry.cpp
```diff
@@ -82,6 +85,15 @@ double Geometry::getWidth() {
 
 
 /**
+ * @brief Returns the total depth (z extent) of the Geometry in cm.
+ * @return the total depth of the Geometry (cm)
+ */
+double Geometry::getDepth() {
```
wbinventor: I can see these height, width and depth routines coming back to haunt us in the future - what about `Geometry::getDeltaX()`, `Geometry::getDeltaY()` and `Geometry::getDeltaZ()`?

src/Geometry.h
```diff
@@ -76,21 +76,28 @@ class Geometry {
 
 private:
 
-  /** The boundary conditions at the top of the bounding box containing
-   *  the Geometry. False is for vacuum and true is for reflective BCs. */
-  boundaryType _top_bc;
+  /** The boundary conditions at the x-min surface of the bounding box 
+   * containing the Geometry. False is for vacuum and true is for reflective
```
wbinventor: Is the second sentence here about "False" vs. "true" for vacuum/reflective BCs needed? I know this was here prior to this PR, but upon a second look at this, it seems that this is unneeded since these attributes are of type `boundaryType` rather than booleans.

src/Geometry.h
```diff
@@ -131,8 +139,7 @@ class Geometry {
   boundaryType getMaxXBoundaryType();
   boundaryType getMinYBoundaryType();
   boundaryType getMaxYBoundaryType();
-  boundaryType getMinZBoundaryType();
-  boundaryType getMaxZBoundaryType();
```
wbinventor: Could these routines still be useful (albeit, perhaps more so in the future) even if we only do xy-planar ray tracing on some z level?

src/LocalCoords.cpp
```diff
@@ -77,6 +77,16 @@ int LocalCoords::getLatticeY() const {
 
 
 /**
+ * @brief Return the third index of the Lattice cell within which this
+ *        LocalCoords resides.
+ * @return the third Lattice cell index
+ */
+int LocalCoords::getLatticeZ() const {
+  return _lattice_z;
```
wbinventor: Seeing that a 3D lattice coordinate is now built into `LocalCoords` makes me even more strongly believe that we should only accept 3D lists of lattice universes.

src/LocalCoords.h
```diff
@@ -93,14 +99,16 @@ class LocalCoords {
...
   void setX(double x);
   void setY(double y);
+  void setZ(double z);
   void setNext(LocalCoords *next);
   void setPrev(LocalCoords* coords);
 
   LocalCoords* getLowestLevel();
   LocalCoords* getHighestLevel();
-  void adjustCoords(double delta_x, double delta_y);
+  void adjustCoords(double delta_x, double delta_y, double delta_z=0.0);
```
wbinventor: Is `LocalCoords:adjustCoords(...)` used anywhere with the default `delta_z` argument? If not, I would remove the default value and require it to be set everywhere (or make all args have a default).

src/Material.cpp
```diff
@@ -299,9 +299,9 @@ FP_PRECISION* Material::getDifHat() {
 
   if (_dif_hat == NULL) {
 
-    _dif_hat = new FP_PRECISION[4*_num_groups];
+    _dif_hat = new FP_PRECISION[NUM_SURFACES*_num_groups];
 
-    for (int e = 0; e < 4*_num_groups; e++)
+    for (int e = 0; e < NUM_SURFACES*_num_groups; e++)
```
wbinventor: This is just me being picky and I think this would be more relevant for a different PR, but it seems like this loop would be more intuitive if broken up into two for one over surfaces and one over energies (and if `_dif_hat` had a 2D indexing macro accepting both surface and energy group indices).

src/Material.cpp
```diff
@@ -1214,9 +1214,9 @@ void Material::setDifHat(double* xs, int num_groups) {
                _id, _num_groups);
 
   if (_dif_hat == NULL)
-    _dif_hat = new FP_PRECISION[4*_num_groups];
+    _dif_hat = new FP_PRECISION[NUM_SURFACES*_num_groups];
 
-  for (int i=0; i < _num_groups*4; i++)
+  for (int i=0; i < _num_groups*NUM_SURFACES; i++)
```
wbinventor: Again, just being picky, but "i" and "e" are used interchangeably as loop iteration variables for these loops over surfaces/energy groups for dif_hat.

wbinventor: If leaving this as a single loop, I would use "i" since "e" would seem to connote energy group index.

src/Material.cpp
```diff
@@ -1535,13 +1535,13 @@ Material* Material::clone() {
     if (_buckling != NULL)
       clone->setBucklingByGroup((double)_buckling[i], i+1);
 
-    for (int j=0; j < 4; j++) {
+    for (int j=0; j < NUM_SURFACES; j++) {
 
       if (_dif_hat != NULL)
-        clone->setDifHatByGroup((double)_dif_hat[i*4+j], i+1, j);
+        clone->setDifHatByGroup((double)_dif_hat[i*NUM_SURFACES+j], i+1, j);
```
wbinventor: Here _dif_hat_ is indexed with 2D indices

src/Point.h
```diff
@@ -19,26 +19,25 @@
...
-  /** The Point's y-coordinate */
-  double _y;
+  /** The Point's coordinates */
+  double* _xyz;
 
 public:
   Point();
   virtual ~Point();
-  void setCoords(const double x, const double y);
+  void setCoords(const double x, const double y, const double z=0.0);
```
wbinventor: Ditto, I'm not sure it's necessary to make `z` the only parameter with a default value? Seems like none of them or all of them should have defaults unless I'm missing something.

src/Point.h
wbinventor: I personally like this notation better, it follows with what is used in OpenMC, but recognize that it is just a personal preference issue (I don't see any performance issues either way).

src/Solver.h
```diff
@@ -107,6 +107,9 @@ class Solver {
   /** The number of fissionable flat source regions */
   int _num_fissionable_FSRs;
 
+  /** The number of independent transport sweeps */
+  int _num_independent_sweeps;
```
wbinventor: Is this to allow multiple sweeps per outer source iteration?

src/Surface.cpp
wbinventor: Yeah, not sure why it was there in the first place (that was a long time ago!) but this is worrisome.

src/Surface.h
```diff
@@ -197,9 +197,12 @@ class Plane: public Surface {
   /** The coefficient for the linear term in y */
   double _B;
 
-  /** The constant offset */
+  /** The coefficient for the linear term in y */
```
wbinventor: This comment mirrors the one above for `_B` - I think this should be "linear term in z"

src/Track.cpp
```diff
@@ -372,7 +375,11 @@ bool Track::contains(Point* point) {
         ((point->getY() <= _start.getY()+1.0E-2 &&
           point->getY() >= _end.getY()-1.0E-2)
       || (point->getY() >= _start.getY()-1.0E-2 &&
-          point->getY() <= _end.getY()+1.0E-2)))) {
+          point->getY() <= _end.getY()+1.0E-2)) &&
```
wbinventor: Oh man, what the heck is this?! We should really make this "1.0E-2" a macro or something. I know this is old code, but it should be cleaned up now that we are looking over it once again here.

src/TrackGenerator.cpp
wbinventor: Agreed

src/TrackGenerator.cpp
```diff
@@ -429,19 +429,19 @@ bool TrackGenerator::containsTracks() {
  *
  * @code
  *          num_tracks = track_generator.getNumTracks()
- *          coords = track_generator.retrieveTrackCoords(num_tracks*4)
+ *          coords = track_generator.retrieveTrackCoords(num_tracks*6)
  * @endcode
  *
- * @param coords an array of coords of length 4 times the number of Tracks
+ * @param coords an array of coords of length 6 times the number of Tracks
```
wbinventor: The "6" here should be defined perhaps in a doxygen "@details" comment block

src/TrackGenerator.cpp
```diff
@@ -429,19 +429,19 @@ bool TrackGenerator::containsTracks() {
...
  * @param num_tracks the total number of Tracks
  */
 void TrackGenerator::retrieveTrackCoords(double* coords, int num_tracks) {
 
-  if (num_tracks != 4*getNumTracks())
+  if (num_tracks != 6*getNumTracks())
```
wbinventor: Replace "6" with a macro

src/TrackGenerator.cpp
```diff
@@ -469,22 +471,22 @@ void TrackGenerator::retrieveTrackCoords(double* coords, int num_tracks) {
...
  * @endcode
  *
- * @param coords an array of coords of length 5 times the number of segments
+ * @param coords an array of coords of length 7 times the number of segments
  * @param num_segments the total number of Track segments
  */
 void TrackGenerator::retrieveSegmentCoords(double* coords, int num_segments) {
 
-  if (num_segments != 5*getNumSegments())
+  if (num_segments != 7*getNumSegments())
```
wbinventor: Ditto "7" as a macro

src/TrackGenerator.cpp
```diff
@@ -632,15 +637,17 @@ void TrackGenerator::initializeTrackFileDirectory() {
...
                   << "x" << _geometry->getCmfd()->getNumY()
-                  << ".data";
+                  << ")_cmfd.data";
     }
   else{
     test_filename << directory.str() << "/"
                   << _num_azim*2.0 << "_angles_"
-                  << _spacing << "_cm_spacing.data";
+                  << _spacing << "_cm_spacing_"
+                  << _geometry->getZLevel() << "_z_level.data";
```
wbinventor: I'm not sure that "z_level" is necessary here, how about this instead:

```C++
<< "_z_" << _geometry->getZLevel() << ".data";
```

src/TrackGenerator.cpp
```diff
@@ -677,6 +684,7 @@ void TrackGenerator::initializeTracks() {
   double* phi_eff = new double[_num_azim];
 
   double x1, x2;
+  double z = _geometry->getZLevel();
```
wbinventor: @samuelshaner I was wondering your thoughts - it seems to me that the z level concept is more applicable to the `TrackGenerator` than to the `Geometry`. Would you agree? Could it be easily transferred to the `TrackGenerator` from the `Geometry`?

src/TrackGenerator.cpp
```diff
@@ -747,13 +755,13 @@ void TrackGenerator::initializeTracks() {
 
       /* If Track points to the upper right */
       if (i < _num_azim / 2)
-        _tracks[i][_num_x[i]+j].getStart()->setCoords(0,
-                                     dy_eff[i] * (0.5 + j));
+        _tracks[i][_num_x[i]+j].getStart()->setCoords
+          (0, dy_eff[i] * (0.5 + j), z);
```
wbinventor: I think the opening paranthesis is typically put on the preceding rather than the trailing line

src/TrackGenerator.cpp
```diff
@@ -1044,17 +1056,18 @@ void TrackGenerator::computeEndPoint(Point* start, Point* end,
                                      const double height) {
 
   double m = sin(phi) / cos(phi);             /* slope */
-  double yin = start->getY();                 /* y-coord */
   double xin = start->getX();                 /* x-coord */
-
+  double yin = start->getY();                 /* y-coord */
+  double zin = start->getZ();                 /* Z-coord */
```
wbinventor: capital "Z-coord" comment here doesn't follow that above: "z-coord"

src/Universe.cpp
```diff
@@ -734,10 +690,12 @@ Lattice::Lattice(const int id, const char* name): Universe(id, name) {
   _offset.setCoords(0.0, 0.0);
 
   /* Default width and number of Lattice cells along each dimension */
-  _num_y = -1;
-  _num_x = -1;
+  _num_x = 0;
```
wbinventor: There was a reason that the default number of lattice cells was set to "-1" at initialization, but I can't recall it. Is there a good reason to make it 0.0 now?

src/Universe.cpp
```diff
@@ -746,10 +704,14 @@ Lattice::Lattice(const int id, const char* name): Universe(id, name) {
  */
 Lattice::~Lattice() {
 
-  /* Clear the vector of vector of Universes */
-  std::vector< std::vector< std::pair<int, Universe*> > >::iterator iter;
-  for (iter = _universes.begin(); iter != _universes.end(); ++iter)
-    iter->clear();
+  std::map<int, Universe*> unique_universes = getUniqueUniverses();
```
wbinventor: I don't see `unique_universes` being used here at all

wbinventor: or the iterator `iter` for that matter

src/Universe.cpp
wbinventor: Looks like there is still one extra space here

src/Universe.cpp
```diff
@@ -1050,20 +1046,93 @@ void Lattice::setUniverses(int num_y, int num_x, Universe** universes) {
   /* Set the Lattice dimensions */
   setNumX(num_x);
   setNumY(num_y);
-
+  setNumZ(1);
+  
   Universe* universe;
 
+  _universes.push_back
+    (std::vector< std::vector< std::pair<int, Universe*> > >());
```
wbinventor: Ditto comment about opening parenthesis on continuation line

src/Universe.cpp
```diff
@@ -1050,20 +1046,93 @@ void Lattice::setUniverses(int num_y, int num_x, Universe** universes) {
...
+                             Universe** universes) {
+
+  std::map<int, Universe*> unique_universes = getUniqueUniverses();
+  std::map<int, Universe*>::iterator iter;
+
+  /* Remove all Universes in the Lattice */
+  for (iter = unique_universes.begin(); iter != unique_universes.end(); ++iter)
+    removeUniverse(iter->second);
+
+  /* Clear all Univers maps in the Lattice (from a previous run) */
```
wbinventor: typo: "Univers" -> "Universe"

src/Universe.cpp
```diff
@@ -1050,20 +1046,93 @@ void Lattice::setUniverses(int num_y, int num_x, Universe** universes) {
...
+
+  /* Set the Lattice dimensions */
+  setNumX(num_x);
+  setNumY(num_y);
+  setNumZ(num_z);
+
+  Universe* universe;
+
+  /* The Lattice cells are assumed input in row major order starting from the
+   * upper left corner. This double loop reorders the Lattice cells from the
```
wbinventor: typo: "cells from the to start" -> "cells to start"

src/Universe.h
```diff
@@ -141,29 +139,38 @@ class Lattice: public Universe {
...
 
+  /** The width of each Lattice cell (cm) along the z-axis */
+  double _width_z;
+  
   /** The coordinates of the offset for the Universe */
   Point _offset;
 
   /** A container of Universes ? */
-  std::vector< std::vector< std::pair<int, Universe*> > > _universes;
+  std::vector< std::vector< std::vector< std::pair<int, Universe*> > > >
```
wbinventor: Extra spaces between brackets

src/Universe.h
```diff
@@ -141,29 +139,38 @@ class Lattice: public Universe {
...
+    _universes;
 
 public:
 
   Lattice(const int id=-1, const char* name="");
   virtual ~Lattice();
 
-  void setOffset(double x, double y);
+  void setOffset(double x, double y, double z=0.0);
```
wbinventor: Default value for `z` argument needed?

src/Universe.h
```diff
@@ -171,16 +178,20 @@ class Lattice: public Universe {
   double getMinZ();
   double getMaxZ();
 
-  Universe* getUniverse(int lat_x, int lat_y) const;
-  std::vector< std::vector< std::pair<int, Universe*> > > getUniverses() const;
+  Universe* getUniverse(int lat_x, int lat_y, int lat_z=0) const;
```
wbinventor: Default value for `lat_z` argument needed?

wbinventor:
Thanks for this PR @samuelshaner, this is awesome! Everything works out for me, I think this will be ready to go quickly as @geogunow and mine comments are addressed.

openmoc/map_to_dict.i
samuelshaner: There was just an inconsistency with the template for the `setUniverses` method. The template for the method exists in three places: `map_to_dict.i`, 'Universe.h`, and `Universe.cpp`. In `map_to_dict.i` and `Universe.h` the template was:

`(int num_x, int num_y, Universe** universes)`

In `Universe.cpp` the template was:

`(int num_y, int num_x, Universe** universes)`

Note the switch of `num_x` and `num_y` in the templates. These modifications make all the templates consistent. The previous templates produced the desired result; their was just an inconsistency that could cause confusion.

openmoc/map_to_dict.i
samuelshaner: I'm on board with this, but if this is done I'm in favor of changing the `Circle` class to `ZCylinder` to make it clear that we are requiring geometries to be 3D instead of assuming they are axially extruded.

openmoc/map_to_dict.i
samuelshaner: correct

openmoc/map_to_dict.i
wbinventor: I'm fine with that as well.

openmoc/map_to_dict.i
samuelshaner: Sure, the comment seems to confusing to me now that I read it again. Perhaps something more physical like `"Get the list of universes in the j-th row in the k-th xy-plane"`.

openmoc/map_to_dict.i
samuelshaner: Also, I think something like `"Check that the number of universes in the j-th row of the k-th xy-plane is the same as the number of cells in the 1st row of the 1st z-plane"` would more accurately describe what this is check for, physically.

openmoc/plotter.py
samuelshaner: Ok. I'll make it a constant and note that allowable values are 4 and 6.

openmoc/map_to_dict.i
wbinventor: I agree, sounds good

On Mon, Oct 12, 2015 at 6:53 PM, sam <notifications@github.com> wrote:

> In openmoc/map_to_dict.i
> <https://github.com/mit-crpg/OpenMOC/pull/188#discussion_r41809937>:
>
> > +  $4 = (Universe**) malloc(($1 * $2 * $3) * sizeof(Universe*)); // universes
> > +
> > +  /* Loop over y */
> > +  for (int k = 0; k < $1; k++) {
> > +
> > +    /* Get the inner list in the nested list for the lattice */
> > +    PyObject* outer_outer_list = PyList_GetItem($input,k);
> > +
> > +    /* Loop over y */
> > +    for (int j = 0; j < $2; j++) {
> > +
> > +      /* Get the inner list in the nested list for the lattice */
> > +      PyObject* outer_list = PyList_GetItem(outer_outer_list, j);
> > +
> > +      /* Check that the length of this list is the same as the length
> > +       * of the first list */
>
> Also, I think something like "Check that the number of universes in the
> j-th row of the k-th xy-plane is the same as the number of cells in the 1st
> row of the 1st z-plane" would more accurately describe what this is check
> for, physically.
>
> —
> <https://github.com/mit-crpg/OpenMOC/pull/188/files#r41809937>.
>



-- 
Will Boyd
Nuclear Science & Engineering
Cell: 423.413.8469
wbinventor@gmail.com

openmoc/plotter.py
samuelshaner: I added constants to `constants.h` as follows:

`#define NUM_VALUES_PER_RETRIEVED_TRACK 6`
`#define NUM_VALUES_PER_RETRIEVED_SEGMENT 7`

These values are used in the `plotter` and `TrackGenerator` in setting the length of the coords array used to retrieve the tracks and segments. In the future, if developers want to bundle more info in the arrays they just need to change this value in one location and add/extract the necessary data to the array. I'm open to other suggestions if you see a better method.

openmoc/thisown.i
samuelshaner: I've tried it with SWIG 3 on my mac and SWIG 2 on AWS and it has been working fine in both environments.

sample-input/benchmarks/c5g7/c5g7-3d-rodded-B.py
samuelshaner: I contemplated adding another geometry file, but it won't reduce the code in the input files since the rodded B 3d geometry input is used in only one place. In the future when we have the full 3D implemented, we will have the three versions of the C5G7 geometry (`c5g7`, `c5g7-rodded-A`, and `c5g7-rodded-B`). How about I add the rodded A geometry and then have 3 geometry files: `geometry.py`, `geometry_rodded_A.py`, and `geometry_rodded_B.py`. The `c5g7.py`, `c5g7-gpu.py`, and `c5g7-cmfd.py` will used `geometry.py` and input files for the rodded A and rodded B cases will use their respective input files. @wbinventor - what do you think of that?

wbinventor: Ok, I didn't realize that this comment was opening up a can of worms. If
it's not too much trouble, go ahead with your plan; otherwise just ignore
my comment for this PR.

sample-input/benchmarks/c5g7/c5g7-3d-rodded-B.py
samuelshaner: Nope, I just needed the control rod material and added it to the python materials input file. I just added it to the hdf5 materials file and will update this with the hdf5 file.

sample-input/benchmarks/c5g7/c5g7-3d-rodded-B.py
samuelshaner: good catch

sample-input/benchmarks/c5g7/c5g7-3d-rodded-B.py
samuelshaner: yep, I like that condensed notation.

sample-input/benchmarks/c5g7/c5g7-3d-rodded-B.py
samuelshaner: I like the `ZCoord` notation over the `ZLevel` notation so I'll change it to that throughout.

src/Cell.cpp
samuelshaner: correct, the full transition to 3D capabilities will be presented over several PRs to \*hopefully\* make them manageable to review!

src/Geometry.cpp
samuelshaner: I completely agree - the height, width, and depth can be a bit confusing since they don't explicitly refer to a dimension. Unless there is any objections, I'm going to go with the notation `Geometry::getWidthX()`, `Geometry::getWidthY()`, `Geometry::getWidthZ()` since that is the notation used in the `Lattice` object in OpenMOC and OpenCG.

src/Geometry.h
samuelshaner: Correct, I'll change the comments.

src/Geometry.h
samuelshaner: As discussed, I'd recommend saving for a future PR.

src/LocalCoords.h
samuelshaner: It is never called with anything other than the default, so I'll just remove it.

src/Point.h
samuelshaner: I'll remove the default.

src/Point.h
samuelshaner: I don't have a preference, but I like the idea of it being an ordered array so I'm going to keep it as `_xyz`.

src/Solver.h
samuelshaner: This was just a different notation I was using with the 3D Solver. I'll remove this.

src/TrackGenerator.cpp
samuelshaner: It could be transferred and I agree that intuitively, it seems more fitting to have it in the `TrackGenerator`. However, doing this would then require that the plotter either use some other default zcoord (e.g. 0.0) for plotting the materials, cells, etc. I'll make it a `TrackGenerator` attribute and have the default zcoords in the plotter be 0.0.

src/Material.cpp
samuelshaner: Just created issue #211 where this will be addressed.

src/Material.cpp
samuelshaner: This will be addressed in issue #211.

src/Track.cpp
samuelshaner: This function is actually never used. I believe it was just used in debugging a long time ago. @wbinventor - how do you feel about removing it altogether?

src/TrackGenerator.cpp
samuelshaner: That is correct. I'll also make the indent 4 lines. Interestingly, the google C++ style guide recommends a 4 line indent if no function arguments are given on the line that starts the function, but recommends aligning the arguments with the start of the parentheses if any arguments are given on the line that starts the function.

wbinventor: Alright, I'll start abiding by that in my own code from now on. If it is good enough for Google, it is good enough for us!

src/Track.cpp
wbinventor: If it's not used I say remove it.

src/Universe.cpp
samuelshaner: I think it is because in `Lattice::findCell` and `Lattice::getUniverse` (which should actually have the error checking modified to be the same as `findCell`), their is error checking to see if the input `lat_x` is within the bounds of 0 and `_num_x`. A given input of `lat_x = 0` would be valid even if the lattice has not been initialized. The `_num_x = -1` ensures that these functions error out or return NULL if `_num_x = -1` for all input values of `lat_x`. I'll change it back to -1.

src/Universe.cpp
samuelshaner: The Google C++ style guide does not have one correct way to format this line. To summarize for an example template `std::vector<std::vector<int>>`:
- C++11 is okay with `>>` so `std::vector<std::vector<int>>` is fine.
- C++03 requires a space between the ending angle brackets `> >` so at least `std::vector<std::vector<int> >` is required.
- Google C++ style guide is okay with either of these or symmetric spacing like `std::vector< std::vector<int> >`.

The precedent in OpenMOC looks to be symmetric spacing, so I'm going to change to that.

src/constants.h
samuelshaner: I figured that for 3D MOC, this will be modified to include the z-surfaces and corners/edges that include z.

wbinventor:
@samuelshaner all of your recent commits look great to me, and all of my tests checkout. This PR is awesome, I can't wait to start using this!

PR #190, opened 2015-07-30
Removing unnecessary Makefile flags from C++ alternative build

geogunow:
This is a simple PR to remove un-needed Makefile options from the alternative C++ build system such as MPI and PAPI (not used currently in OpenMOC). This should be merged soon, barring any objections.


geogunow:
@jtramm visually reviewed this PR. Merging now.

Issue #192, opened 2015-08-04
GPU solver on multi-GPU system memory allocation issue


cjosey:
So, I was testing #162 on the GPU.  The only machine with GPUs I have available has 3 GPU installed simultaneously:
ID 0 is a GTX Titan X
ID 1 is a Quadro K600
ID 2 is a Tesla K20c
I compiled the current develop branch of OpenMOC with CUDA 7.5, and ran c5g7-gpu.py.  I get the following result:
```
[  NORMAL ]  Computing the eigenvalue...
Traceback (most recent call last):
  File "c5g7-gpu.py", line 328, in <module>
    solver.computeEigenvalue(max_iters)
  File "/home/cjosey/.local/lib/python3.4/site-packages/openmoc/cuda/openmoc_cuda.py", line 153, in computeEigenvalue
    def computeEigenvalue(self, *args, **kwargs) -> "void" : return _openmoc_cuda.Solver_computeEigenvalue(self, *args, **kwargs)
RuntimeError: Could not allocate memory for fluxes on GPU
```
I then traced that error message to ``initializeFluxArrays``, specifically, the instant ``_boundary_flux.resize(size);`` is called, an error is thrown and the code fails.

As a slight modification, I modified the code so that ``cudaSetDevice(n);`` was called at the beginning of ``initializeFluxArrays``.  If ``n`` is set to 0 or 2, it fails identically to the above.  If it's set to 1, I get the following (with some debug code to output the size of the arrays allocated and GPU used):

```
[  NORMAL ]  Computing the eigenvalue...
CUDA Device:             Quadro K600
76440
936012
[  NORMAL ]  Initializing FSRs on the GPU...
terminate called after throwing an instance of 'thrust::system::system_error'
  what():  cudaFree in free: an illegal memory access was encountered
Aborted (core dumped)
```

I am unfamiliar with GPU coding, so I do not know if this is my fault, the code's fault, or some combination of the two.  @wbinventor, any ideas?

wbinventor:
First, the Python interface in "openmoc.cuda" already includes an
"attach_gpu(...)" routine to select which GPU you wish to use.

I tested this code out with a GPU using CUDA 7.0 an AWS server last night
and it worked fine. I think your comments here should be raised as a
separate issue from this PR. I'm not sure how to fix them, but have seen
issues similar to them in the past when debugging CUDA code. You might try
running the "c5g7-gpu.py" case to see if it works or not.

cjosey:
[...]

PR #193, opened 2015-08-04
Changed GPU compile flags to a generic one for newer machines

cjosey:
This changes the compile flags for NVCC from ones that will only compile for ``sm_20`` and ``sm_30`` to one that will JIT compile based on the user's GPU.

This allows for higher shader model machines, and might slightly reduce compile time as fewer binaries are created.  Code was tested on 3 GPUs successfully, an ``sm_30`` device (Quadro K600), an ``sm_35`` device (Tesla K20), and an ``sm_52`` device (Titan X).  Prior to this patch, only the first one was supported.

In testing with C5G7 (with 32 angles), there was no performance difference between any ``compute_nn``, so ``compute_20``, which supports the most devices, was selected.

This also fixes issue #192.

### DIFF
--- a/config.py
+++ b/config.py
@@ -206,8 +206,7 @@ class configuration:
   compiler_flags['nvcc'] =  ['--relocatable-device-code', 'true',
                              '-c', '-O3',  '-std=c++11', 
                              '--compiler-options', '-fpic',
-                             '-gencode=arch=compute_20,code=sm_20',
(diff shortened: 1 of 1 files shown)


wbinventor:
Thanks for fully investigating this @cjosey! Perhaps @geogunow and @jtramm can apply your knowledge to the C++ alternative build system for anyone who wants to take the CUDA code to the next level someday.

PR #195, opened 2015-08-05
Fixed FSR keys to be unique

geogunow:
This PR corrects for the potential bug of two unique ``fsr_key`` strings hashing to the same key. Currently OpenMOC converts an FSR region into a unique string identifier and then hashes that string to an integer (``size_t``). That integer is then used as the key to be inserted into the hash table. This is incorrect since the hash table already uses an internal hashing function. This leads to potentially loading the incorrect FSR.

This PR corrects this behavior to instead use the unique string identifiers as keys. In this way, we gain the benefit of hashing (present internally in the hash table) but also safeguard against inconsistency by directly comparing the unique string identifiers.

### DIFF
--- a/src/Geometry.cpp
+++ b/src/Geometry.cpp
@@ -565,22 +565,21 @@ int Geometry::findFSRId(LocalCoords* coords) {
   int fsr_id;
   LocalCoords* curr = coords;
   curr = coords->getLowestLevel();
-  std::hash<std::string> key_hash_function;
 
   /* Generate unique FSR key */
-  std::size_t fsr_key_hash = key_hash_function(getFSRKey(coords));
+  std::string fsr_key = getFSRKey(coords);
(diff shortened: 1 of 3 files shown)


geogunow:
To provide more motivation for this PR, I did a little math on the current hashing technique. In the current form, OpenMOC should fail 1% of the time if the number of FSRs exceeds 690 million on 64-bit machines. However, if we run on 32-bit machines a 1% failure rate will occur with only 9291 FSRs!!!

So for anyone running OpenMOC on 32-bit machines, this PR is imperative. Also, for 3D MOC we could envision large numbers of FSRs to converge a full core problem and we might even hit that 64-bit threshold for 1% failure.

samuelshaner:
Tested it on mac and linux and everything looks good.

PR #196, opened 2015-08-07
Cmfd shared memory optimization

jtramm:
In this PR I've optimized the shared memory locking implementation for CMFD. When using many threads, I observe performance improvements of 1.2x - 2x on c5g7-cmfd depending on machine and problem size.

The optimization is the result of reducing the number of times the locks are accessed. Rather than preparing data and locking in the same step for each energy group and polar angle, we now prepare our contributions in advance for all energy groups, then lock once and write all data in one transaction.

I've also added simd pragmas inside the update loop, which don't offer much improvement for low group numbers, but should speed things up for larger problem sizes. I chose the "omp simd" flavor as this is a little more portable.

### DIFF
--- a/src/Cmfd.cpp
+++ b/src/Cmfd.cpp
@@ -2155,7 +2155,7 @@ void Cmfd::tallySurfaceCurrent(segment* curr_segment, FP_PRECISION* track_flux,
 
     surf_id = curr_segment->_cmfd_surface_fwd;
 
-    /* Prepare surface current */
+    /* Prepare Cmfd Mesh surface current */
     for (int e=0; e < _num_moc_groups; e++) {
       surf_current[e] = 0.f;
 
@@ -2165,7 +2165,7 @@ void Cmfd::tallySurfaceCurrent(segment* curr_segment, FP_PRECISION* track_flux,
       surf_current[e] = surf_current[e] / 2.f;
     }
 
-    /* SHM write surface current */
+    /* SHM write Cmfd Mesh surface current */
     omp_set_lock(&_surface_locks[surf_id]);
     #pragma omp simd
     for (int e=0; e < _num_moc_groups; e++)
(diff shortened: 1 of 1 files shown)


geogunow:
I just reviewed this PR and it looks straightforward. Seems there are merge conflicts though. Once those are resolved, this should be good to merge.

samuelshaner:
The locks for the currents (now both surface and corner currents) have been moved to the `Vector` object   to avoid having redundant code with the same lock structure. In order to take advantage of the simd vectorization you would probably have to add a function in the `Vector` object to set multiple values at one time.

@jtramm - what do you think about just having `#pragma omp atomic` in the `Vector::setValue` and `Vector::incrementValue` routines instead of having locks altogether?

jtramm:
I think atomics will be significantly slower (perhaps over 10x) based on experience with SimpleMOC, as the OpenMP atomics are not implemented very efficiently -- though I will be sure to try that out first to make certain. I think locks will definitely be the way to go here. The issue is making sure to lock once and then write for all energy groups, rather than locking for each energy group (as they are all using the same lock).

I agree that putting the locking code in the Vector class allows for cleaner, easier to read code and reduces code redundancy -- but I think the potential 2x overall performance improvement makes a little code readability worth giving up in the top hotspot functions of the source code.

I'll work on merging my optimizations with PR #182.

samuelshaner:
@jtramm - that makes a lot of sense. Thanks for the detailed explanation.

How about adding two methods, `Vector::incrementValues(int cell, int group_start, int group_end, FP_PRECISION* vals)` and `Vector::setValues(int cell, int group_start, int group_end, FP_PRECISION* vals)`, that perform simd set and increment operations for multiple values at a time? I feel this should give the same performance as your PR modifications while keeping the locks within the `Vector` object.

jtramm:
That should work nicely -- I'll try that route.

geogunow:
@jtramm what's the status of this PR? Do you think you could sync it up sometime or should we close it?

jtramm:
I haven't had a chance to work on getting it rolled in yet. I would still
like to add this optimization at some point, but may not get around to it
until January or so. Feel free to close it, I can always reopen it down the
line or start a fresh one once I have some time to work on this.

wbinventor:
Is this PR necessary or did @samuelshaner incorporate these changes into #254?

samuelshaner:
PR #254 addresses this issue so I'll go ahead and close this PR.

PR #197, opened 2015-08-07
Bug fix in track generation

samuelshaner:
PR #184 made an assumption in the track generation process that turned out not be true. Namely, it assumed that tracks with phi (0, PI/2) were generated with start points going `left to right` along the x-axis and then bottom to top along the y-axis. However, at the time OpenMOC was generating tracks with (0, PI/2) with start points going `right to left` along the x-axis and then bottom to top along the y-axis. This PR changes the track ordering for tracks with phi (0, PI/2) to be generated going `left to right` along the x-axis and then bottom to top along the y-axis in order to preserve the condensed function to set the boundary conditions from PR #184.

This PR does not depend on any other PRs and should be integrated before any other PRs.


geogunow:
This looks simple and reasonable. Are there any tests I can run to confirm the correctness?

samuelshaner:
You can run c5g7-cmfd.py with 16 angles and 0.05 spacing before and after to see how the current develop branch fails. In order to confirm the correctness, I don't have any specific example. When setting the track in and track out, you can print out the start and end points if you want.

geogunow:
Good enough for me, this PR fixes that issue

wbinventor:
Did you test the ``GPUSolver`` with this before merging? It is now segfaulting.

wbinventor:
Disregard my previous comment, I thought I was on the "develop" branch but was actually on my own experimental branch which produced the segfaults.

wbinventor:
Actually, I take that back - PR #184 is causing the ``GPUSolver`` to segfault. @geogunow did you test any of these PRs out on a GPU before merging?

samuelshaner:
I know what the issue is. I took out the `setTrackInI`, `setTrackInJ`, `setTrackOutI`, and `setTrackOutJ` functions since they did not seem to be used or needed. I'll add them back and submit another PR. Sorry guys.

wbinventor:
That's right, just opened an issue for this. Thanks for looking at it @samuelshaner!

Issue #198, opened 2015-08-07
GPUSolver now segfaults in track initialization


wbinventor:
It appears that PR #184 is now leading the ``GPUSolver`` to segfault. This PR removed calls to ``Track::setTrackInI(...)``, ``Track::setTrackInJ(...)``, ``Track::setTrackOutI()`` and ``Track::setTrackOutJ()`` when initializing BCs. As a result, the ``GPUSolver`` now segfaults when initializing tracks, since it needs the ``i`` and ``j`` indices for each track (in particular, the ``GPUSolver::computeTrackScalarIndex(...)`` called [here](https://github.com/mit-crpg/OpenMOC/blob/develop/src/accel/cuda/GPUSolver.cu#L1288)) needs the indices.

PR #199, opened 2015-08-08
Fixed segfault for GPUSolver caused by not setting track i and j

samuelshaner:
This track fixes the issue #198 due to a change in PR #184 in not setting the track i and j indices. I have tested on the GPU for c5g7 and it looks to produce correct results.


wbinventor:
This makes sense to me. Thanks for the quick turnaround on this @samuelshaner!

PR #201, opened 2015-08-16
Update to Periodic Boundary Conditions PR

samuelshaner:
This PR updates PR #185 by cleaning up the additions to the TrackGenerator, Solver, GPUSolver, and clone. The changes in this PR include:
- Fixed typo in `config.py` for the icpc source files.
- Added `_num_tracks_by_parallel_group` array storing the track UIDs that separate the groups of tracks that can be looped over in parallel in the Solver's transportSweep method.
- Added `_num_parallel_track_groups` attribute to indicate the number of parallel track groups.
- Added `_tracks_by_parallel_groups` 1D array of track pointers that stores the tracks arranged by parallel track group. This array was created to avoid the redundant loops in the `TrackGenerator` and `Solver` that set the Track UIDs and form the 1D array of track pointers that stores the tracks arranged by parallel track group.
- Added method `TrackGenerator::initializeTrackPeriodicCycleIndices`. Formerly, the code in this method was included in the `TrackGenerator::initializeTrackUIDs` method.
- Shortened `GPUSolver::transportSweep` method by changing hard coded transport sweeps over parallel groups to a loop over the parallel groups.

### DIFF
--- a/src/CPUSolver.cpp
+++ b/src/CPUSolver.cpp
@@ -681,13 +681,13 @@ void CPUSolver::transportSweep() {
   if (_cmfd != NULL && _cmfd->isFluxUpdateOn())
     _cmfd->zeroCurrents();
 
-  /* Loop over azimuthal angle and periodic track halfspaces */
-  for (int i=0; i < 6; i++) {
+  /* Loop over the parallel track groups */
+  for (int i=0; i < _num_parallel_track_groups; i++) {
 
     /* Compute the minimum and maximum Track IDs corresponding to
-     * this azimuthal angle and periodic track halfspace */
-    min_track = _num_tracks_by_halfspace[i];
-    max_track = _num_tracks_by_halfspace[i+1];
+     * this parallel track group */
+    min_track = _num_tracks_by_parallel_group[i];
+    max_track = _num_tracks_by_parallel_group[i+1];
 
     /* Loop over each thread within this azimuthal angle halfspace */
     #pragma omp parallel for private(curr_track, azim_index, num_segments, \
@@ -696,7 +696,7 @@ void CPUSolver::transportSweep() {
 
       tid = omp_get_thread_num();
 
-      /* Use local array accumulator to prevent false sharing*/
+      /* Use local array accumulator to prevent false sharing */
       FP_PRECISION* thread_fsr_flux;
       thread_fsr_flux = new FP_PRECISION[_num_groups];
 
@@ -819,14 +819,36 @@ void CPUSolver::transferBoundaryFlux(int track_id,
   /* For the "forward" direction */
   if (direction) {
     start = _tracks[track_id]->isNextOut() * _polar_times_groups;
-    bc = std::min((int)_tracks[track_id]->getBCOut(), 1);
+
+    /* If the outgoing boundary condition is vacuum, set bc to 0 indicating the
+     * track flux of the outgoing track will be zeroed out */
+    if (_tracks[track_id]->getBCOut() == VACUUM)
+      bc = 0;
+
+    /* If the outgoing boundary condition is anything other than vacuum, set
+     * bc to 1 indicating the flux passed to the outgoing track should be the
+     * current track's flux */
+    else
+      bc = 1;
+
     track_out_id = _tracks[track_id]->getTrackOut()->getUid();
   }
 
   /* For the "reverse" direction */
   else {
     start = _tracks[track_id]->isNextIn() * _polar_times_groups;
-    bc = std::min((int)_tracks[track_id]->getBCIn(), 1);
+
+    /* If the incoming boundary condition is vacuum, set bc to 0 indicating the
+     * track flux of the incoming track will be zeroed out */
+    if (_tracks[track_id]->getBCIn() == VACUUM)
+      bc = 0;
+
+    /* If the incoming boundary condition is anything other than vacuum, set
+     * bc to 1 indicating the flux passed to the incoming track should be the
+     * current track's flux. */
+    else
+      bc = 1;
+
     track_out_id = _tracks[track_id]->getTrackIn()->getUid();
   }
 
--- a/src/CPUSolver.cpp
+++ b/src/CPUSolver.cpp
@@ -666,12 +666,13 @@ void CPUSolver::computeKeff() {
 void CPUSolver::transportSweep() {
 
   int tid;
-  int min_track, max_track;
   int azim_index, num_segments;
   Track* curr_track;
   segment* curr_segment;
   segment* segments;
   FP_PRECISION* track_flux;
+  int min_track = 0;
+  int max_track = 0;
 
   log_printf(DEBUG, "Transport sweep with %d OpenMP threads", _num_threads);
 
@@ -686,8 +687,8 @@ void CPUSolver::transportSweep() {
 
     /* Compute the minimum and maximum Track IDs corresponding to
      * this parallel track group */
-    min_track = _num_tracks_by_parallel_group[i];
-    max_track = _num_tracks_by_parallel_group[i+1];
+    min_track = max_track;
+    max_track += _track_generator->getNumTracksByParallelGroup(i);
 
     /* Loop over each thread within this azimuthal angle halfspace */
     #pragma omp parallel for private(curr_track, azim_index, num_segments, \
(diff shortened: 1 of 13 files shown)


src/Solver.cpp
```diff
@@ -356,48 +358,11 @@ void Solver::setTrackGenerator(TrackGenerator* track_generator) {
...
-          }
-          else if (period_halfspace == 2 && index % 2 == 0 && index != 0) {
-            _tracks[counter] = track;
-            counter++;
-          }
-        }
-      }
-    }
-  }
+  _tracks = _track_generator->getTracksByParallelGroup();
```
wbinventor: Not a big deal, but the routine about which returns an array pointer ends in "Array" while this one returns an Array but does not include that suffix. Perhaps either include "Array" as a suffix in both method names for consistency, or remove it from both (seems fine, then nams are both each pretty long anyway).

src/Solver.h
```diff
@@ -126,9 +126,9 @@ class Solver {
   /** The total number of Tracks */
   int _tot_num_tracks;
 
-  /** An integer array with the Track uid separating the azimuthal and periodic
-   * halfspaces */
-  int* _num_tracks_by_halfspace;
+  /** An integer array with the Track uid separating the parallel track
```
wbinventor: I found this comment confusing to read at first glance. Does the "uid" really add any relevant info here? I would remove that piece of the comment, or elaborate on it further if needed (if anything, one can't always assume that a uid is a monotonically increasing global integer counter as is the case for our Tracks).

src/Solver.cpp
samuelshaner: My vote is to remove the `Array` suffix so I'll remove the two instances of it.

src/TrackGenerator.cpp
```diff
@@ -123,18 +153,17 @@ int* TrackGenerator::getNumTracksArray() {
 
 
 /**
- * @brief Return an array with the Track uid separating the azimuthal and
- * periodic halfspaces
- * @return array with the Track uid separating the azimuthal and periodic 
- *         halfspaces
+ * @brief Return an array with the Track uid separating the parallel track
```
wbinventor: Ditto my comment above that this can be confusing to a novice like me

src/TrackGenerator.cpp
```diff
@@ -559,7 +605,9 @@ void TrackGenerator::generateTracks() {
     }
   }
 
+  /* Initialize the track boundary conditions and set the track UIDs */
```
wbinventor: Also I would say "UID" as opposed to "uid" in comments to be more clear as you do here

src/TrackGenerator.cpp
```diff
@@ -559,7 +605,9 @@ void TrackGenerator::generateTracks() {
...
   initializeBoundaryConditions();
+  initializeTrackPeriodicCycleIndices();
```
wbinventor: Maybe it's just me, but some of the camelCase method names can be pretty long. Does "Periodic" really add anything here to the method name?

src/Solver.h
samuelshaner: It is not really needed here, but the fact that the Solver's `_tracks` array is arranged so the Track UID corresponds to its index in the Solver's `_tracks` array is not coincidence as the UID is used as the index in the `dev_tracks` array in the `GPUSolver::initializeTracks` method. I'll remove the mention of uid here.

src/accel/cuda/GPUSolver.cu
```diff
@@ -1452,59 +1452,18 @@ void GPUSolver::transportSweep() {
...
+  for (int g=0; g < _num_parallel_track_groups; g++) {
+
+    tid_offset = _num_tracks_by_parallel_group[g] * _num_groups;
+    tid_max = _num_tracks_by_parallel_group[g+1];
+
+    transportSweepOnDevice<<<_B, _T, shared_mem>>>(scalar_flux, boundary_flux,
+                                                   reduced_sources,
+                                                   _materials, _dev_tracks,
+                                                   tid_offset, tid_max);
+  }
```
wbinventor: Make sure to add a call to "cudaDeviceSynchronize()" after the CUDA kernel call here. I added this in #187 recently, but may as well add it in here since this will be aeege conflict anyway.

src/TrackGenerator.cpp
samuelshaner: I'll remove the mention of uid here too.

src/TrackGenerator.cpp
samuelshaner: ok

src/TrackGenerator.cpp
samuelshaner: In 3D MOC we have both periodic and reflective cycles, so I decided to use the "Periodic" specifier here.

src/accel/cuda/GPUSolver.cu
samuelshaner: ok

src/CPUSolver.cpp
```diff
@@ -819,14 +819,36 @@ void CPUSolver::transferBoundaryFlux(int track_id,
   /* For the "forward" direction */
   if (direction) {
     start = _tracks[track_id]->isNextOut() * _polar_times_groups;
-    bc = std::min((int)_tracks[track_id]->getBCOut(), 1);
+
+    /* If the outgoing boundary condition is vacuum, set bc to 0 indicating the
+     * track flux of the outgoing track will be zeroed out */
```
wbinventor: I guess I thought that the if-else would be in the TrackGenerator's BC initialization code. It would be nice to avoid the if-else here since it is inside the core computationally expensive piece of the MOC algorithm. lbeit, I would not expect this to degrade performance in any substantial way, so leave it as you see fit.

src/TrackGenerator.cpp
wbinventor: Ok so there will be a separate routine to create reflective cycles?

wbinventor: Or a single routine which takes in a parameter to determine the cycle type?

src/Solver.h
wbinventor: That makes perfect sense, and I realized it myself but only after digging through some of the other code to figure out how the UIDs are set. It's not a big deal, just wasn't exactly clear what was meant by the comment at first sight.

src/CPUSolver.cpp
samuelshaner: Ok, I'll have to change the Track's `_bc_in` and `_bc_out` attributes to int types then. What do you think about changing these attributes to bool types named something more like `_transfer_flux_out` and `_transfer_flux_in` and then casting them as ints in the solver?

src/Solver.cpp
wbinventor: Sounds good to me

src/CPUSolver.cpp
wbinventor: Are you sure you have to cast them? I thought you could simply multiply a float by a boolean, but I may be wrong (def worthwhile to make sure both GNU and Intel allow for this). If it is not allowed, yes, I think that your proposal is a good compromise.

wbinventor:
This looked good to me yesterday, pending changes per the comments I made. I didn't have a chance to test it out yet and likely won't till tonight, but if @geogunow wants to review and test this out first, that is fine by me.

src/TrackGenerator.cpp
geogunow: I second @wbinventor's suggestion of a single with a parameter as long as it wouldn't be too tough

src/CPUSolver.cpp
geogunow: I believe almost every compiler will do the casting automatically (like when you divide a double by an int)

src/Solver.h
```diff
@@ -126,9 +126,9 @@ class Solver {
...
+   *  groups */
+  int* _num_tracks_by_parallel_group;
```
geogunow: Can someone explain to me why we are using a pointer to an ``int`` rather than just an ``int``?

src/TrackGenerator.cpp
```diff
@@ -821,44 +873,106 @@ void TrackGenerator::initializeTrackUIDs() {
...
+  /* Allocate memory for the num tracks by parallel group array */
+  _num_tracks_by_parallel_group = new int[_num_parallel_track_groups + 1];
+  _num_tracks_by_parallel_group[0] = 0;
+
+  /* Allocate memory for the tracks by parallel group array */
+  int num_tracks = getNumTracks();
+  _tracks_by_parallel_group = new Track*[num_tracks];
+
+  /* Loop over the parallel track groups */
+  for (int g = 0; g < _num_parallel_track_groups; g++) {
```
geogunow: This is fine for now, but eventually we will want to change this to be parallel over all tracks, using point-jacobi updating of angular fluxes

src/TrackGenerator.h
```diff
@@ -47,9 +47,12 @@ class TrackGenerator {
   /** An integer array of the number of Tracks for each azimuthal angle */
   int* _num_tracks;
 
-  /** An integer array with the Track uid separating the azimuthal and periodic
-   * halfspaces */
-  int* _num_tracks_by_halfspace;
+  /** An integer array with the Track uid separating the parallel groups */
+  int* _num_tracks_by_parallel_group;
```
geogunow: ditto on the pointer

geogunow:
I just did a quick review of the code. I ran all the non-GPU sample inputs on my local computer and they check out. I left a few comments on the code structure. I don't see any glaring typos or bugs. Someone else should be responsible for merging this one since I'm moving back to Boston tomorrow morning, arriving on the 20th. So I'll be pretty busy this week.

src/TrackGenerator.cpp
samuelshaner: In 2D and 3D, we only need the periodic track indices if periodic BCs are present. In 3D, we always need reflective cycle indices irregardless of the BCs. Currently, in 3D-MOC, there is not a separate routine specifically for creating the reflective cycle indices. The reflective cycle indices for 3D tracks are set in the routine where the 3D tracks are created.

I like the idea of a single routine called something line `initializeTrackCycleIndices(boundaryType bc)` that accepts a `boundaryType` enum with the type of cycle to set indices for. Then, the track will have both `_periodic_track_index` and `_reflective_track_index` even though we don't need the `_reflective_track_index` at the moment. What do you guys think of that?

src/CPUSolver.cpp
samuelshaner: Looking at the C standard, it appears a bool will be implicitly converted to an integer. I'll test with gnu and intel compilers to be sure.

src/Solver.h
samuelshaner: This is a pointer to an array of ints.

src/TrackGenerator.cpp
samuelshaner: agreed

src/TrackGenerator.h
samuelshaner: ditto on array of ints

src/TrackGenerator.cpp
samuelshaner: The track object has methods `getUid` and `setUid`. Do you think methods should use `UIDs` or `Uids` in their name? I would vote to have acronyms all capitalized as this seems to be the norm in the rest of OpenMOC for uses of `MOC` and `FSR`.

wbinventor: I like all caps for acronyms as well.

wbinventor:
I will test this out this evening and hopefully merge it in.

wbinventor:
Looks good with GNU, Intel, and CUDA.

PR #160, opened 2015-06-24
Hot fixes for Cell initialization

wbinventor:
This PR fixes a few bugs which led to segmentation faults for certain use cases when initializing new ``Cells``. I ran into these situations while updating the online docs and executing the inline code examples. In particular, the elimination of ``CellBasic`` (filled by ``Materials``) and ``CellFill`` (filled by ``Universes``) in lieu of a single ``Cell`` class filled by either a ``Material`` or ``Universe`` in PR #142.

PR #203, opened 2015-09-22
Hotfix for OpenCG/OpenMC compatibility

wbinventor:
This PR makes the default ``Universe`` ID=-1 rather than 0 to support OpenCG compatibility with OpenMC geometries which must have a universe with ID=0 (at the root). In addition, this PR removes some elements of ``openmoc.compatible.opencg_compatible`` for memory management of geometric primitives (*e.g.*, setting the SWIG ``thisown`` attributes) which was resolved in #176 and #160.

### DIFF
--- a/src/Universe.h
+++ b/src/Universe.h
@@ -155,7 +155,7 @@ private:
 
 public:
 
-  Lattice(const int id=0, const char* name="");
+  Lattice(const int id=-1, const char* name="");
   virtual ~Lattice();
 
   void setOffset(double x, double y);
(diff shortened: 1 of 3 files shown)


src/Universe.h
```diff
@@ -83,7 +83,7 @@ class Universe {
 
 public:
```
samuelshaner: I believe you also need to set the default const int id for `Lattice` to -1. Otherwise, all lattices where the user doesn't specific the id will have an id of `0`.

wbinventor: Thanks for picking up on that @samuelshaner, the latest commit makes the ``Lattice`` default ID -1 as well.

PR #205, opened 2015-09-29
Hotfix for advanced OpenCG compatibility

wbinventor:
This PR fixes a few issues I encountered when trying to interactively run multiple OpenMOC simulations in the same IPython Notebook with different multi-group cross sections defined on some OpenCG mesh (from OpenMC). This fixes a bug such that the ``thisown`` attribute for SWIG-wrapped ``Material`` Python objects is set to zero when a ``Material`` is added to a ``Cell`` (effectively delegating responsibility to the underlying C++ ``Cell`` object to delete the ``Material``). For some odd reason this bug crept in my PR #161. The second bug that this PR fixes is the deletion of ``Material`` objects for distinct calls of the ``opencg_compatible.get_openmoc_geometry(...)`` routine.

### DIFF
--- a/openmoc/compatible/opencg_compatible.py
+++ b/openmoc/compatible/opencg_compatible.py
@@ -744,6 +744,8 @@ def get_opencg_geometry(openmoc_geometry):
     raise ValueError(msg)
 
   # Clear dictionaries and auto-generated IDs
+  OPENMOC_MATERIALS.clear()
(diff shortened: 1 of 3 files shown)


wbinventor:
It would be good for @geogunow to confirm whether or not this will break the sample input for the C++-only build system (*i.e.*, if those inputs use dynamic allocation for ``Materials``).

PR #206, opened 2015-09-29
Added logger message to print final keff as a RESULT message

wbinventor:
This PR simply makes it such that the final keff value in an eigenvalue calculation is reported in a ``RESULT`` logger message. In an interactive environment like IPython Notebook, it can be helpful to throttle much of OpenMOC's output (*e.g.*, ``NORMAL`` messages showing eigenvalue convergence) and instead simply report the final converged eigenvalue. This one line PR makes it easy to do so by ensuring that the final eigenvalue (converged or unconverged) is reported to the screen as a ``RESULT`` message.


wbinventor:
I'm going to remove this since the user can easily query the final keff using the ``Solver.getKeff()`` routine.

PR #210, opened 2015-10-10
Refactor openmoc.materialize module for OpenMC MGXS compatiblity

wbinventor:
This PR refactors the `openmoc.materialize` module to enable it to be compatible with the new `openmc.mgxs` module for multi-group cross section generation in OpenMC. Some other changes introduced in this PR include:

* Now use four spaces in the `log`, `krylov` and `materialize` modules for PEP8 conformance
* Removed support for Python dictionaries of multi-group cross sections in `materialize`
* Renamed cross section types to shorter strings (*e.g.*, 'Total XS' -> 'total')
* Updated the C5G7 and LRA sample HDF5 materials data files

Many of these items will break backwards compatibility. Once this PR is merged, I will immediately update the documentation to reflect these changes to the `openmoc.materialize` API. Fortunately, the `Material` class API has not changed so anyone setting MGXS manually using `Material` class routines will not be affected.

I do not think this PR should be merged until `openmc.mgxs` is finalized for OpenMC. However, by opening this here I thought I could open it up for review in order to make any proactively make any necessary changes.


openmoc/materialize.py
```diff
@@ -2,192 +2,319 @@
...
-# Determine which OpenMOC module is being used. This is important for
-# the materialize module since it must instantiate Material objects
-# with the same floating point precision as that compiled into the
-# openmoc module used in the main Python input script to OpenMOC.
+# @date October 9, 2015
 
 import sys
+import os
+
+import numpy as np
```
samuelshaner: Adding the import of numpy here seems to imply that numpy is now a hard requirement for OpenMOC, whereas `config.py` has a flag indicating whether OpenMOC should be compiled with numpy typemaps. Do you want to just make numpy a hard requirement for OpenMOC now?

wbinventor: We did previously support an option to build OpenMOC without embedding the NumPy C API into the code. I gutted most of this functionality a while back (though it looks like I didn't *quite* finish the job then, so I did with my most recent commit here) since it was only there to support compilation on BGQ. Since BGQ really isn't (shouldn't in my opinion) be a target platform for us with the Python API, I think it is fine to make numpy a hard requirement since it is so widely available these days. And if anyone is interested in deploying the code on funky hardware like BGQ, they can simply use the Python-less alternative build system from #159.

samuelshaner: Sounds good to me.

openmoc/materialize.py
```diff
@@ -2,192 +2,319 @@
...
+            if isinstance(domain_spec, int):
+                material = openmoc.Material(id=domain_spec)
+            else:
+                material = openmoc.Material(name=domain_spec)
+
+        # Add material to the collection
+        materials[domain_spec] = material
+        material.setNumEnergyGroups(num_groups)
+
+        # Search for the total/transport cross section
```
samuelshaner: For the inexperienced user, I'm worried that having two options (transport and total) that do the same thing could be confusing. In `src/Material.h` and `src/Material.cpp` sigmaT is described as the total cross section. I think adopting one descriptor for sigmaT throughout the code would be wise to avoid confusion. Could this be done? If so, which descriptor do you think would be more fitting, total or transport?

openmoc/materialize.py
```diff
@@ -2,192 +2,319 @@
...
+        elif 'scatter matrix' in domain_group:
+            sigma = _get_numpy_array(domain_group, 'scatter matrix', suffix)
+            material.setSigmaS(sigma)
+        else:
+            py_printf('WARNING', 'No "scatter matrix" or "nu-scatter matrix" '
+                                 'found for "%s %s"', domain_type, domain_spec)
+
+        # Search for chi (fission spectrum)
+        if 'chi' in domain_group:
+            sigma = _get_numpy_array(domain_group, 'chi', suffix)
```
samuelshaner: Could `sigma` be changed to `chi`?

openmoc/materialize.py
```diff
@@ -2,192 +2,319 @@
...
+        # If not Geometry, instantiate a new Material with the ID/name
+        else:
+            material = openmoc.Material(id=domain.id)
+
+        # Add material to the collection
+        materials[domain.id] = material
+        material.setNumEnergyGroups(num_groups)
+
+        # Search for the total/transport cross section
+        if 'transport' in mgxs_lib.mgxs_types:
```
samuelshaner: Same thing here about transport and total.

wbinventor:
I updated this PR with a minor but important fix last night that I wanted to note here. This PR refactors the `openmoc.materialize` module with two core routines:

* `openmoc.materialize.load_from_hdf5(...)`
* `openmoc.materialize.load_openmc_mgxs_lib(...)`

Each of these may take an optional `geometry` parameter so that the routines can insert `Materials` into the appropriate `Cells` in the `Geometry`. 

Previously, an issue would arise if a user creates a geometry with multiple `Cells` filled by the same (dummy) `Material`. If the cross sections in the HDF5 file or OpenMC MGXS library were defined on a per-`Cell` basis, then the cross sections for the (dummy) `Material` were overridden multiple times by the cross sections for each cell filled by that material. This might seem like an arcane situation, but it is a realistic one. For example, this situation would arise if a user created an OpenMC geometry with a bunch of `Cells` filled by the same `Material` to compute an MGXS library (*e.g.*, a discretized fuel pin with distinct MGXS in each zone).

Now, if cross sections are defined by cell, both routines in `openmoc.materialize` will instantiate new `Material` objects for each cell. In the case that a user filled each `Cell` in the `Geometry` with a (dummy) `Material`, the routines will extract the string name (if any) given to the original `Material` and insert it in to the `Material` copy/clone. This allows for cross sections to be defined on a per-cell basis, but albeit may lead to perplexing behavior if a user unknowingly thinks that their dummy `Materials` are actually being used in the model under-the-hood (which they won't be).

wbinventor:
Also, I think that this PR should be reviewed once we finalize #216 since the two will have some merge conflicts, and I think #216 is nearer to being ready to merge (this one should wait on the `openmc.mgxs` PR to OpenMC).

wbinventor:
Now that the new multi-group cross section generation module `openmc.mgxs` has been formalized within OpenMC [PR #471](https://github.com/mit-crpg/openmc/pull/471), I think this PR is ready for final review.

openmoc/materialize.py
wbinventor: Good call, done

openmoc/materialize.py
wbinventor: I added some "INFO" log messages to `openmoc.materialize` in an attempt to help make it clear to the user which cross section types are loaded at runtime. I think our code comments on sigmaT are pretty clear, but perhaps that's not the case to a new users. Our online documentation on the `openmoc.materialize` is due for an update with both this PR and #216, and I think that this is something we can carefully note there.

openmoc/krylov.py
```diff
@@ -8,18 +8,17 @@
 # @date June 30, 2015
 
 import sys
-import copy
+
 import numpy as np
-import scipy.sparse.linalg as linalg
```
samuelshaner: Is there any reason this was moved into the functions and imported in two places? It seems more appropriate to keep the import here.

wbinventor: This makes `scipy` an optional dependency for the `openmoc` module, and only a required dependency if a user wants to use the `openmoc.krylov.IRAMSolver`.

openmoc/materialize.py
```diff
@@ -2,170 +2,356 @@
...
+
+            elif domain_type == 'cell':
+                cell = _get_domain(domains, domain_spec)
+                material = cell.getFillMaterial()
+
+                # If the user filled the Cell with a Material, clone it
+                if material != None:
+                    material = material.clone()
+
+                # If the Cell does not contain a Material, creat one for it
```
samuelshaner: creat -> create

openmoc/materialize.py
```diff
@@ -2,170 +2,356 @@
...
+
+        # Add material to the collection
+        materials[domain_spec] = material
+        material.setNumEnergyGroups(num_groups)
+
+        # Search for the total/transport cross section
+        if 'transport' in domain_group:
+            sigma = _get_numpy_array(domain_group, 'transport', suffix)
+            material.setSigmaT(sigma)
+            py_printf('INFO', 'Loaded "transport" MGXS for "%s %s"',
```
samuelshaner: I believe this will error out if `domain_spec` is an int, so you should probably cast `domain_spec` as a str.

wbinventor: Good call, I cast them to strings now

openmoc/materialize.py
```diff
@@ -2,170 +2,356 @@
...
+        py_printf('ERROR', 'The OpenMC code must be installed on your system')
+
+    if not isinstance(mgxs_lib, openmc.mgxs.Library):
+        py_printf('ERROR', 'Unable to load cross sections from %s which is not '
+                           'an openmc.mgxs.Library object', str(type(mgxs_lib)))
+
+    if geometry and 'openmoc.Geometry' not in str(type(geometry)):
+        py_printf('ERROR', 'Unable to load cross sections for "%s" which '
+                           'is not an OpenMOC Geometry', str(type(geometry)))
+
```
samuelshaner: It looks like there is a lot of repeated code from the `load_openmc_mgxs_lib()` and `load_from_hdf()` functions. Could a helper function be used to include the common code blocks in one location?

wbinventor: I agree that it the two routines mirror each other in a lot of ways. However, it seems to me that in no place is there more than ~2 lines of code in a particular block that is identical between the two routines. In other words, there are single lines of codes in the two routines that match, but not large blocks. So I'm not convinced there is a good way to make use of a helper function here.

wbinventor:
@samuelshaner let me know what you think of this when you get a chance. I have a PR with an SPH calculator waiting in the wings that builds upon this PR, and will be easiest to review (perhaps by @PrezNattyGibbs) once this PR is finalized.

samuelshaner:
@wbinventor I just tested out this PR and everything looks good. I just have one request before it is merged it - could you change `LRA-materials-hdf5.py` to `LRA-mgxs-hdf5.py` to be consistent with the cross section files between C5G7 and LRA?

wbinventor:
Just renamed "LRA-materials-hdf5.py" to "LRA-mgxs-hdf5.py"

PR #213, opened 2015-10-18
Hotfix for OpenCG compatibility with 3D geometries

wbinventor:
This PR fixes a few small issues in the OpenCG compatibility module as a result of #188, namely the following:

* Use of `ZCylinder` in place of `Circle` surfaces
* Use of 3D rather than 2D `Lattices`

I've verified that these fixes allow for full 3D modeling of fuel assemblies extracted from the full core 3D OpenCG BEAVRS model (or at least 2D axial "slices" of any assembly anywhere along their axial extent). This should be a quick merge, unless I've missed anything that @samuelshaner might notice.

PR #214, opened 2015-10-18
Updated IPython Notebook Example

wbinventor:
This PR adds the IPython Notebook linked by the README file as a new example input. The Notebook has been updated to reflect changes to the API to create 3D geometries. My one concern with merging this in is that the example is *insanely* slow now. This is a result of CMFD - it runs much more quickly without CMFD turned on. I have no idea why this is the case, CMFD does not seem to be a bottleneck for our other example inputs.


wbinventor:
Closing this since I realized that it was tied to some other PRs of mine like #213 and #210. I'll open a new one that applies these changes independently.

Issue #211, opened 2015-10-14
Compute the CMFD surface diffusion coefficients on-the-fly


samuelshaner:
Currently, the materials have surface diffusion coefficients (`_dif_hat` and `_dif_tilde`) for each energy group and surface of a cmfd cell. Materials for the MOC problem don't need these variables and they don't really represent material properties - they are intermediate terms used in setting up the diffusion matrices. I would recommend we change the `Cmfd::computeDs(int moc_iteration)` to two new methods:

- `Cmfd::computeSurfaceDiffusionCoefficient(int cmfd_cell, int surface)` - Compute the surface diffusion coefficient based on the boundary conditions and diffusion coefficients in the neighboring cells.
- `Cmfd::computeSurfaceDiffusionCoefficientCorrection(int cmfd_cell, int surface, int moc_iteration)` - Compute the surface diffusion coefficient correction based on the surface currents tallied during the last MOC iteration, the boundary conditions, and the surface diffusion coefficient for the given cmfd cell and surface.

wbinventor:
I like this idea a lot - the slimmer we can make the `Material` class the better in my opinion. We don't need to give the user a front-facing API to get/set diffusion coefficients for CMFD anyway, so computing them on the fly seems like a cleaner approach.

wbinventor:
On a related note, I think CMFD should compute sigma_a on-the-fly as well. The absorption cross-section isn't used anywhere else in the code, and could be gutted from the ``Material``. In addition, when using scattering matrices that account for neutron multiplication it isn't appropriate to the use the standard sigma_ but rather sigma_t - sum(sigma_s), which I think is one reason I've seen some instabilities when using CMFD with my MC-generated MGXS.

wbinventor:
Closed with #216.

PR #216, opened 2015-10-19
Removal of diffusion coefficients from Materials and simplification of Cmfd methods

samuelshaner:
This PR seeks to address issue #211 and update the Material class to remove unnecessary components including the absorption cross section (`_sigma_a`), the diffusion coefficients (`_dif_coef`, `_dif_hat`, and `_dif_tilde`), and buckling (`_buckling`). With these deletions, the cross section input files have been modified to no longer include these terms.

With these modifications to the Material class, the Cmfd class was modified as follows:
- Since the materials no longer have diffusion coefficients, the diffusion coefficients are computed on-the-fly with `computeDiffusionCoefficient` and `computeSurfaceDiffusionCoefficient`.
- Since the diffusion coefficients were removed from the Material class, damping on the surface diffusion coefficient correction term was removed and replaced with the optically thick correction factor (which has been renamed Larsen's effective diffusion coefficient correction factor) and k-nearest updating to maintain stability.
- Several helper functions were added to Cmfd to remove redundant or unclear code blocks.
- The `constructMatrices` and `splitCurrents` methods were modified to loop over the surfaces/corners and apply a general method instead of having multiple cases for the different surfaces/corners of the Cmfd cell.
- As mentioned above, the optically thick correction factor was renamed Larsen's effective diffusion coefficient correction factor and a detailed description was added.

These modifications should make Cmfd consistent with the MOC solver for problems where the scattering matrix has nu-scatter included in it, which did not work before.


src/Cmfd.cpp
```diff
@@ -362,198 +338,191 @@ void Cmfd::computeXS() {
 
 
 /**
- * @brief Compute the surface diffusion coefficients for each cell surface.
+ * @brief Computes the diffusion coefficient for a given cmfd cell and cmfd
```
wbinventor: typo: cmfd shows up here twice in a row

src/Cmfd.cpp
```diff
@@ -362,198 +338,191 @@ void Cmfd::computeXS() {
...
+
+  /* Get diffusivity and flux for Mesh cell */
+  FP_PRECISION dif_coef = getDiffusionCoefficient(cmfd_cell, group);
+  FP_PRECISION flux = _old_flux->getValue(cmfd_cell, group);
+  int cmfd_cell_next = getCellNext(cmfd_cell, surface);
+  FP_PRECISION delta_interface = getSurfaceWidth(surface);
+  FP_PRECISION delta = getPerpendicularSurfaceWidth(surface);
+  int sense = getSense(surface);
+
+  /* Correct the diffusion coefficient with Larsen's effective diffusion
```
wbinventor: Is there a good reference for Larsen's effective diff coeff? If so, you might include it here or in the doxygen comment.

samuelshaner: I had the same thought and you'll see that I added a detailed explanation and references for Larsen's effective diffusion coefficient in the doxygen comment for the `Cmfd:computeLarsensEDCFactor()` method.

src/Cmfd.cpp
```diff
@@ -1165,105 +1081,44 @@ void Cmfd::splitCorners() {
...
-                (cell_next_id, SURFACE_Y_MAX * _num_cmfd_groups + e, current);
-
-            cell_next_id = getCellNext(cell_id, SURFACE_Y_MAX);
-            if (cell_next_id != -1)
-              _surface_currents->incrementValue
-                (cell_next_id, SURFACE_X_MAX * _num_cmfd_groups + e, current);
-          }
-        }
+  int cell_cw, cell_ccw;
+  int surface_cw, surface_ccw;
```
wbinventor: I think these need to be declared as private in the parallel for loop below?

src/Cmfd.cpp
```diff
@@ -1311,39 +1166,24 @@ int Cmfd::getCellNext(int cell_id, int surface_id) {
...
+  return _larsens_edc_on;
 }
 
 
 /**
- * @brief Set whether optically thick diffusion correction factor is in use.
- * @param optically_thick Boolean indicating whether optically thick diffusion
- *        correction factor is in use.
+ * @brief Set whether Larsen's effective diffusion coefficient correction factor
+ *        is in use.
```
wbinventor: You might add a detail comment block highlighting those situations in which the factor should be turned on/off, and whether it is up to the user to turn it on/off or if CMFD does it automatically (it's not yet clear to me from reading the code).

wbinventor:
@samuelshaner this PR looks great to me! My only substantive comment is that the references to the absorption cross section should be removed from the GPU materials. Once this is done I will test this code out and we can move quickly to merge it in.

src/Cmfd.cpp
samuelshaner: correct, nice catch!

src/Cmfd.cpp
samuelshaner: It is recommended that Larsen's EDC always be on as it improves stability and, in theory, provides a better approximation of the diffusion coefficient for CMFD-accelerated MOC. This function is nice for testing, but there isn't really any value in the user having this function so I'm just going to take it out completely.

samuelshaner:
I have addressed the comments and tested the code on mac, linux, and on the gpu and everything look good to merge.

wbinventor:
I'm testing this out now and see that the "sample-input/gradients" cases all reference the now non-existent absorption cross-section by calling `Material.setSigmaA(...)`. Can you remove these calls?

wbinventor:
Ditto for the following inputs:

* "sample-input/benchmarks/romano"
* "sample-input/benchmarks/homogeneous-one-group"
* "sample-input/benchmarks/homogeneous-two-group"

wbinventor:
Everything checks out for me so once the absorption cross section references are removed from those sample inputs, this is ready to merge.

samuelshaner:
Just grep'ed `sigma_a` and `setsigmaa` in the sample inputs directory and removed all the references of the absorption cross section from the input files.

PR #217, opened 2015-10-21
Made openmoc.compatible module Python 3 compatible

wbinventor:
This PR just makes a few trivial changes to the "__init__.py" file for the `openmoc.compatible` submodule such that it is Python 3 compatible.

### DIFF
--- a/openmoc/compatible/__init__.py
+++ b/openmoc/compatible/__init__.py
@@ -1,2 +1,2 @@
-from casmo import *
-from opencg_compatible import *
+from openmoc.compatible.casmo import *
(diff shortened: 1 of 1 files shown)


geogunow:
Looks extremely simple. Could you tell me why the previous version doesn't work with Python 3 but works with Python 2.7?

wbinventor:
Submodule imports changed in Python 3 as shown here.

wbinventor:
You can Google around for a detailed answer to this, but the short of it is
that imports within a module must include the full module name in Python 3.
This is more strict than Python 2, but is also more consistent and
transparent (I think).

PR #218, opened 2015-11-03
changed Chi to chi in c5g7 fission chamber mgxs

samuelshaner:


wbinventor:
Oh wow, I thought I had picked this up in PR #210 but I guess not. Thanks @samuelshaner!

PR #219, opened 2015-11-03
SuPerHomogenisation (SPH) Factor Calculator

wbinventor:
This PR adds the capability to compute SuPerHomogenisation (SPH) factors for arbitrary geometries from OpenMC multi-group cross section (MGXS) libraries. SPH factors are used to correct for energy and angle condensation and spatial homogenisation errors made during MGXS generation. In the case of this PR, the SPH scheme can be used to produce a "corrected" OpenMC MGXS library designed to preserve the flux in fissionable spatial regions, given an accurate source (*i.e.*, the source from OpenMC) in all spatial zones.

The SPH scheme was first proposed and documented by Hebert:

* Hebert, A., "A Consistent Technique for the Pin-by-Pin Homogenization of a Pressurized Water Reactor Assembly." *Nuclear Science and Engineering*, 113 (3), pp. 227-233, 1993.

The scheme implemented here is very similar to the generic scheme proposed by Hebert. The SPH scheme is implemented in the `openmoc.materialize` module. In particular, the following routines have been added to that module:
* `openmoc.materialize.compute_sph_factors(...)`
* `openmoc.materialize._load_openmc_src(...)`
* `openmoc.materialize._apply_sph_factors(...)`

In addition a `openmoc.process.get_scalar_flux(...)` routine was added to extract a NumPy array of the FSR fluxes from a `Solver` object. This is a helper routine used during SPH iteration. Furthermore, the `TrackGenerator::initializeVolumes()` routine which assigns average volume/areas to each `Cell` and `Material`, along with the number of "instances" of each `Cell` and `Material` found in the `Geometry` during ray tracing. The volume and number of instances are needed by the SPH calculation scheme to normalize the volume-integrated fluxes from OpenMC.

A new sample input file "sample-input/sph-factors/sph-factors.py" has been added, along with a 2-group OpenMC MGXS library for the 1.6% enriched BEAVRS fuel pin. This example input file illustrates how easy it is to compute SPH factors and the improvement SPH factors make on the OpenMOC eigenvalue with respect to the ("correct") OpenMC eigenvalue.

@PrezNattyGibbs - wanna take a look and let me know what you think? @samuelshaner @geogunow - since I had to add a few new helper routines to the C++ source code to make this work (nothing major though), you may also want to look over this at some point.


openmoc/materialize.py
```diff
@@ -272,7 +273,9 @@ def load_openmc_mgxs_lib(mgxs_lib, geometry=None):
 
                 # If the user filled the Cell with a Material, clone it
                 if material != None:
-                    material = material.clone()
+                    # FIXME!!!
+#                    material = material.clone()
+                    material = material
```
samuelshaner: What needs to be fixed about clone?

openmoc/materialize.py
```diff
@@ -355,3 +358,288 @@ def load_openmc_mgxs_lib(mgxs_lib, geometry=None):
...
+# @param sph_tol the tolerance on the SPH factor convergence
+# @param fix_src_tol the tolerance on the MOC fixed source calculations
+# @param num_azim number of azimuthal angles to use in fixed source calculations
+# @param track_spacing the track spacing to use in fixed source calculations
+# @return a NumPy array of SPH factors and a new openmc.mgxs.Library object
+#         with the SPH factors applied to each MGXS object
+def compute_sph_factors(mgxs_lib, max_iters=10, sph_tol=1E-6, fix_src_tol=1E-5,
+                        num_azim=4, track_spacing=0.1, zcoord=0.0):
+
+    import openmc.mgxs
```
samuelshaner: Can we have a try/catch around this so it can tell users about the dependency if they don't have openmc installed?

wbinventor:
Also, @PrezNattyGibbs if you want to review this and run the new sample input, you will need to install both OpenCG and the Python API for OpenMC. This is actually very easy - the following is all you need to do to install OpenCG

```bash
git clone https://github.com/mit-crpg/OpenCG
cd OpenCG
python setup.py install --user
```

and likewise for OpenMC (Python API **only**):

```bash
git clone https://github.com/mit-crpg/openmc
cd openmc
python setup.py install --user
```

This will install the `opencg` and `openmc` Python modules on your system such that they are accessible when you run "sample-input/sph-factors/sph-factors.py".

openmoc/materialize.py
wbinventor: Yes, this is something I'm still working on, I'll post here once it is fixed. Material cloning is needed here when one has an OpenMC MGXS library tallied on cells when multiple OpenMC `Cells` use the same `Material` (*e.g.*, different spatial discretizations, such as rings an sectors in a fuel pin). In this case, we must create `Material` *clones* for each `Cell` in the coresponding OpenMOC `Geometry`. 

This works just fine if one simply runs a single eigenvalue / fixed source calculations in a script, such as is the case in all of our example inputs. However, I need to put in a little more sophisticated logic here to permit one to run multiple calculations in the same input (*i.e.*, a fixed source calculations for each SPH iteration).

openmoc/materialize.py
wbinventor: We discussed this very same topic in an OpenMC thread recently, where OpenCG, Pandas and h5py are optional dependencies needed only for certain features in the OpenMC Python API. We decided to forgo `try`/`except` blocks since it was decided that the `ImportError` that Python reports when a module is not found on a user's system is just as informative as any custom messages that we might craft to put in the `except` block. I stuck to the same style here, but am open to a `try`/`except` around this if you and/or @PrezNattyGibbs or @geogunow would like to see it.

sample-input/sph-factors/sph-factors.py
```diff
@@ -0,0 +1,78 @@
...
+solver = openmoc.CPUSolver(track_generator)
+solver.setConvergenceThreshold(tolerance)
+solver.setNumThreads(num_threads)
+
+
+###############################################################################
+#                 Eigenvalue Calculation w/o SPH Factors
+###############################################################################
+
+# Run an eigennvalue calulation with the MGXS from OpenMC
```
samuelshaner: eigennvalue -> eigenvalue

wbinventor: thanks, just fixed this mis-spelling

src/TrackGenerator.cpp
```diff
@@ -1047,6 +1049,58 @@ void TrackGenerator::initializeTrackUids() {
...
+               "have not yet been generated");
+
+  int num_FSRs = _geometry->getNumFSRs();
+  Cell* cell;
+  Material* material;
+
+  /* Compute volume and number of instances for each Cell and Material */
+  for (int i=0; i < num_FSRs; i++) {
+    cell = _geometry->findCellContainingFSR(i);
+    cell->setVolume(cell->getVolume() + getFSRVolume(i));
```
samuelshaner: This is going to be very slow since it has to look over all the segments for every fsr. How about caching the fsr volumes computed by integrating the volume of the segments?

wbinventor:
I just pushed a commit which removes the "#FIXME" comments in `openmoc.materialize`. The loading of MGXS for materials will now work for MGXS defined for either cells or materials. Furthermore, the same is true for the SPH factor scheme - it will now work for `openmc.mgxs.Library` object defined for either "material" or "cell" domain types. Finally, I made a minor change to the `TrackGenerator::initializeVolumes()`. The first implementation of this routine was O(NxK) operation with N tracks and K FSRs. Although this did not reveal itself for small problems like the pin cell sph factor example, it made larger problems like C5G7 intractable. It was a very minor code change to make this routine O(K) instead, so now all example inputs should run as fast as ever.

src/TrackGenerator.cpp
```diff
@@ -1047,6 +1049,61 @@ void TrackGenerator::initializeTrackUids() {
...
+  Cell* cell;
+  Material* material;
+  int num_FSRs = _geometry->getNumFSRs();
+  FP_PRECISION* fsr_volumes = getFSRVolumes();
+
+  /* Compute volume and number of instances for each Cell and Material */
+  for (int i=0; i < num_FSRs; i++) {
+    cell = _geometry->findCellContainingFSR(i);
+    cell->setVolume(cell->getVolume() + fsr_volumes[i]);
+    cell->setNumInstances(cell->getNumInstances() + 1);
```
samuelshaner: Currently, the only use of `Cell::setNumInstances()` and `Material::setNumInstances()` is to increment the number of instances. Could this just be `incrementNumInstances()` and there be no method argument?

wbinventor:
I added `incrementVolume(...)` and `incrementNumInstances()` routines to both `Cell` and `Material` classes. They are now used by the `TrackGenerator` instead of the `setVolume(...)` and `setNumInstances(...)` routines to improve code readability.

In addition, the volume stored by the `Cell` and `Material` are now both the aggregate volume of all instances of each throughout the geometry instead of the instance-averaged volume.

In addition, I fixed a bug in the `TrackGenerator::getFSRVolume(...)` routine. This routine is not used by the SPH factor calculator, but is simply a routine for the user to extract FSR volumes for post-processing.

wbinventor:
@PrezNattyGibbs any time to look over this in the coming week?

wbinventor:
This weekend I plan to extend the SPH factor calculator to accommodate a nested iteration scheme with an outer loop over regions (fissile and non-fissile) and an inner loop over sph residuals (for the region of itnerest). This won't dramatically change the overarching SPH scheme though, and will simply be a secondary mode one can use in the SPH calculator. It would be good to get some more feedback on this PR in the meantime so we can move it into the codebase soon.

wbinventor:
This is ready for final review. Any takers? 

My latest (and final) update introduces a doubly nested iteration scheme to the SPH factor calculator as requested by @bforget and @kord. The initial implementation only applied SPH factors to fissionable domains. The new implementation introduces a `domains` parameter to the calculator that can be set to "fissionable" (default) or "all". When `domains="all"`, an outer loop over all domains (fissionable or not) is applied atop an inner loop over fixed source calculations. Each inner loop fixes the MGXS in all domains aside from the one of interest specified in the outer loop. The inner loop applies SPH factors computed from fixed source calculations until the flux in the domain of interest converges to the OpenMC flux. Then the outer loop proceeds to the next domain, and so on...

Interestingly, the nested iteration scheme does not produce a more accurate eigenvalue for the example pin cell problem. For comparison, the following are the eigenvalues from OpenMC and OpenMOC without SPH, and SPH for "fissionable" or "all" domains:

```
OpenMC:                          1.02790 +/- 0.00094
OpenMOC w/o SPH:                 1.03212
OpenMOC w/ "fissionable" SPH:    1.02719
OpenMOC w/ "all" SPH: 	         1.02541
```

Perhaps there is still something amiss in the solution, but I've been unable to identify any glaring issues. Perhaps another set of eyes can confirm if everything looks correct or not.

Although these latest changes introduce new more complex capabilities to the SPH calculator, I think that overall the scheme has been simplified after my second run through the code.

samuelshaner:
I was waiting for @PrezNattyGibbs to jump in, but it seems like he's occupied so I'll do the review for this PR.

openmoc/materialize.py
```diff
@@ -355,3 +360,314 @@ def load_openmc_mgxs_lib(mgxs_lib, geometry=None):
...
+# @param mgxs_lib an openmc.mgxs.Library object
+# @param max_fix_src_iters the max number of fixed source iterations per domain
+# @parma max_domain_iters the max number of iterations across domains
+# @param domains the domain to apply SPH factors to ('fissionable' or 'all')
+# @param sph_tol the tolerance on the SPH factor convergence
+# @param fix_src_tol the tolerance on the MOC fixed source calculations
+# @param num_azim number of azimuthal angles to use in fixed source calculations
+# @param track_spacing the track spacing to use in fixed source calculations
+# @param num_threads the number of threads to use in fixed source calculations
+# @return a NumPy array of SPH factors and a new openmc.mgxs.Library object
```
samuelshaner: Needs a param for zcoord.

openmoc/materialize.py
```diff
@@ -355,3 +360,314 @@ def load_openmc_mgxs_lib(mgxs_lib, geometry=None):
...
+
+    # Map FSRs to domains to compute domain-averaged fluxes
+    num_domains = len(mgxs_lib.domains)
+    fsrs_to_domains = np.zeros(num_fsrs)
+    for fsr in range(num_fsrs):
+        cell = geometry.findCellContainingFSR(fsr)
+
+        if mgxs_lib.domain_type == 'material':
+            fsrs_to_domains[fsr] = cell.getFillMaterial().getId()
+        else:
```
samuelshaner: Could someone create a mgxs lib with a `universe` domain (and perhaps in the future mesh domain)? If so, I think it would be worthwhile to check for these other domain types and error out if they are not supported.

openmoc/materialize.py
```diff
@@ -355,3 +360,314 @@ def load_openmc_mgxs_lib(mgxs_lib, geometry=None):
...
+    for i in range(max_domain_iters):
+        for domain_id in sph_domains:
+            py_printf('NORMAL', 'SPH outer iteration %d: %s %d',
+                                i, mgxs_lib.domain_type.capitalize(), domain_id)
+
+            inner_sph = np.ones((num_domains, num_groups))
+
+            for j in range(max_fix_src_iters):
+
+                # Run fixed source calculation with throttled verbosity
```
samuelshaner: Just curious, why is the verbosity throttled?

openmoc/materialize.py
```diff
@@ -355,3 +360,314 @@ def load_openmc_mgxs_lib(mgxs_lib, geometry=None):
...
+                openmoc.set_log_level('WARNING')
+                solver.computeFlux()
+                openmoc.set_log_level('NORMAL')
+
+                # Extract the FSR scalar fluxes and total fission rate
```
samuelshaner: It looks like only the FSR scalar fluxes are being extracted

openmoc/materialize.py
```diff
@@ -355,3 +360,314 @@ def load_openmc_mgxs_lib(mgxs_lib, geometry=None):
...
+        else:
+            py_printf('ERROR', 'Unable to compute SPH factors for an OpenMC '
+                               'MGXS library without nu-fission cross sections')
+
+        # Extract an openmc.mgxs.MGXS object for the chi fission spectrum
+        if 'chi' in mgxs_lib.mgxs_types:
+            chi = mgxs_lib.get_mgxs(openmoc_domain.getId(), 'chi')
+        else:
+            py_printf('ERROR', 'Unable to compute SPH factors for an OpenMC '
+                               'MGXS library without nu-fission cross sections')
```
samuelshaner: without nu-fission -> without chi

openmoc/materialize.py
wbinventor: Good catch, got it

openmoc/materialize.py
wbinventor: Yes, the `openmc.mgxs` API supports universe domains, but the `openmoc.materialize.load_openmc_mgxs_lib()` does not (yet) support universe domains. I thought about supporting them at the time I refactored the `materialize` module in #210, but realized there would need to be a little more infrastructure in the code for this to work. In particular, I think a new `Universe::getAllMaterials()` and/or `Universe::getAllMaterialCells()` routine would need to be implemented. This wouldn't be too difficult, but I've never had a use case with MGXS on universe domains, so I left this feature for another day.

More specifically to your question, this code will error out if the MGXS library uses a universe (or mesh in the future) domain in line 451 below.

openmoc/materialize.py
wbinventor: The reason is that the output is very long and difficult to comprehend if it is not throttled. The `Solver::computeFlux()` fixed source calculation scheme will report its own residuals after each transport sweep. Combine this with the residuals from the inner / outer loops for the SPH calculation and there is a lot going on in the output.

To compromise on this, I added a new `throttle_output` parameter to this routine which is `True` by default. In this case, it will use verbosity throttling to suppress the output from the fixed source calculations, and restore the original verbosity level after the completion of the SPH calculation.

openmoc/materialize.py
wbinventor: yeah, this was a copy-n-paste issue from an early sph implementation

openmoc/materialize.py
wbinventor: this now reads "without chi fission spectrum"

wbinventor:
Great comments @samuelshaner, this has been updated to satisfy all of them.

openmoc/materialize.py
```diff
@@ -355,3 +360,328 @@ def load_openmc_mgxs_lib(mgxs_lib, geometry=None):
...
+                sph_domains[openmc_domain.id] = all_domains[openmc_domain.id]
+                sph_indices.append(i)
+    else:
+        py_printf('ERROR', 'SPH factors cannot be applied for domain type '
+                           '%s which is not supported', str(domains))
+
+    # Initialize array of domain-averaged fluxes and SPH factors
+    openmoc_fluxes = np.zeros((num_domains, num_groups))
+
+    # Sore starting verbosity log level
```
PrezNattyGibbs: Sore -> Store

openmoc/materialize.py
```diff
@@ -355,3 +360,328 @@ def load_openmc_mgxs_lib(mgxs_lib, geometry=None):
...
+                if throttle_output and log_level <= openmoc.NORMAL: 
+                    openmoc.set_log_level('NORMAL')
+
+                # Extract the FSR scalar fluxes
+                fsr_fluxes = get_scalar_fluxes(solver)
+
+                # Compute the domain-averaged flux in each energy group
+                for k, domain in enumerate(mgxs_lib.domains):
+                    domain_fluxes = fsr_fluxes[fsrs_to_domains == domain.id, :]
+                    openmoc_fluxes[k, :] = np.mean(domain_fluxes, axis=0)
```
PrezNattyGibbs: I think we want to volume average the fluxes, and this is only kosher if all the volumes are the same.

wbinventor: Sure, but they will all be the same since a "domain" here is a `Cell` or `Material`. All instances of a `Cell` or `Material` will have the same volume, hence this will always be kosher.

openmoc/materialize.py
```diff
@@ -355,3 +360,328 @@ def load_openmc_mgxs_lib(mgxs_lib, geometry=None):
...
+
+    # Compute the final SPH factors for all domains and energy groups
+    final_sph = openmc_fluxes / openmoc_fluxes
+    final_sph = np.nan_to_num(final_sph)
+    final_sph[final_sph == 0.0] = 1.0
+
+    # Restore verbosity log level to original status
+    openmoc.set_log_level(log_level)
+
+    return final_sph, mgxs_lib
```
PrezNattyGibbs: If you have multiple fissionable domains, this algorithm computes each one's SPH factor separately and iterates until they converge.  This is a different approach than Hebert's SPH factors, which use the flux from a fixed source calculation to update the SPH factors for all fissionable domains simultaneously.

openmoc/materialize.py
```diff
@@ -355,3 +360,328 @@ def load_openmc_mgxs_lib(mgxs_lib, geometry=None):
...
+
+                # Compute SPH factor residuals
+                inner_res = np.abs((inner_sph - old_inner_sph) / old_inner_sph)
+                inner_res = np.nan_to_num(inner_res)
+
+                # Report maximum SPH factor residual
+                py_printf('NORMAL', 'SPH inner iteration %d:\tres = '
+                                    '%1.3e', j, inner_res.max())
+
+                # Update multi-group cross sections with SPH factors
```
PrezNattyGibbs: Update this comment to reflect only one domain is being updated.

openmoc/materialize.py
```diff
@@ -355,3 +360,328 @@ def load_openmc_mgxs_lib(mgxs_lib, geometry=None):
...
+        # Check max SPH factor residuals for domains of interest for convergence
+        if inner_res[sph_indices].max() < sph_tol:
+            break
+
+    # Warn user if SPH factors did not converge
+    else:
+        py_printf('WARNING', 'SPH factors did not converge')
+
+    # Compute the final SPH factors for all domains and energy groups
+    final_sph = openmc_fluxes / openmoc_fluxes
```
PrezNattyGibbs: For your work, SPH factors are being computed for MGXS generated from OpenMC.  But it wouldn't be at all farfetched to imagine a day when OpenMOC has subgroup or ultrafine capabilities which could also make use of SPH capabilities.  Are you opposed to abstracting the OpenMC references to something like "reference" or "fine group"?

openmoc/materialize.py
wbinventor: Good point, it's a subtle issue. It just updated the comment, let me know if you have any others you'd like fleshed out.

openmoc/materialize.py
wbinventor: Very, very good catch! This is indeed a "bug" with the current implementation. I didn't (don't yet) have a test problem with more than one fissionable domains (but will soon), and hence hadn't run into this. I'll update you when I push a workaround scheme which applies SPH factors to *all* fissionable domains simultaneously when running in "fissionable" mode.

openmoc/materialize.py
wbinventor: You raise a very good point here - one which I considered when designing the SPH scheme. I agree that it wouldn't at all far-fetched to imagine subgroup and/or ultrafine capabilities which also make use of an SPH scheme. However, I settled upon the current design for the following three reasons:

* It is now so easy (knock on wood) to generate MGXS in OpenMC and plug them into OpenMC that I see it as becoming the de facto standard approach to simulation workflows with OpenMOC for the foreseeable future.

* There is no fundamental software issue precluding subgroup / ultrafine scheme implementations in OpenMOC. The fundamental issue is a people one - no one has stepped up to the plate yet (no offense intended), and I don't see anyone in the pipeline who will implement a "production" version in the future. *If* someone does need these capabilities, then we can discuss how best to better abstract the SPH calculator at that time.

* Related to my last point, it is very difficult to design code abstractions for an ambiguous or moving target. I think this point is related to Donald Knuths' truism that "premature optimization is the root of all evil." I know the OpenMC MGXS generation module very well, and can hence design an SPH customized for it, but don't have time to expend trying to design code for some unknown and non-existent type of MGXS library.

* Lastly, any future developer(s) of subgroup / ultrafine features will need to design scalable data structure(s) to store and retrieve MGXS (and scalar fluxes!) for arbitrary group structures on arbitrary unstructured meshes, yada, yada, yada. My guess is that such a data structure may look a lot like my `openmc.mgxs.MGXS` and `openmc.mgxs.Library` classes. As a result, I'll cross my fingers now and just say that the SPH factor calculator *should* be easily extensible to alternative MGXS library types if an when someone comes along with some time to burn to make it work.

That's my long-winded way of saying that I'm not going to bite on this idea at this point :-) But if you have any thoughts about how this might be better abstracted, write them down here now as a historical record. Who knows, they might be useful in a few years.

openmoc/materialize.py
wbinventor: With some nifty Pythonic tricks (or perhaps not so much), I've update this PR in f235f4f6f2215d5328243ae9c832ad8b9a7c3e9a to simultaneously update the MGXS for *all* fissionable domains in each inner iteration when using Hebert's original SPH scheme. I changed the parameter `domains` to `mode` which can take the values `"fissionable"` (default, Hebert's scheme) or `"all"` (kord's scheme). Take a look at this @PrezNattyGibbs and let me know what you think.

wbinventor:
@PrezNattyGibbs did you try running the example input file for this PR? I know it requires a little extra work to install OpenCG and the OpenMC Python API, but I think that you or @samuelshaner should try it as an independent test of this feature before merging this in.

wbinventor:
Well, this PR still isn't quite ready. I finally was able to try this out with an MGXS library that is "differentiated" with unique data for each pin in am assembly. For some reason the SPH scheme gives very different results from run-to-run and is not reproducible for this larger problem. I'll dig into this after the holiday.

openmoc/materialize.py
PrezNattyGibbs: Fair enough, I'm fine with it as is.  Who knows, maybe someday I'll make a subgroup library and implementation.  But it's reasonable to wait until then to worry about it.

PrezNattyGibbs:
Let me know when you have everything fixed up and ready to review again.  I'll run the sample inputs and take another pass through the code at that point.  But of course, make sure you poke me hard and repeatedly to make sure I don't flake out while I'm thesis writing.

openmoc/materialize.py
wbinventor: That would be cool! And if and when that happens we can figure out how to generalize the SPH calculator, and perhaps other features in the `materialize` module.

wbinventor:
Thanks @PrezNattyGibbs, I hope to have things working soon. I've been trying to debug the irreproducibility issue for the more complicated fuel assembly geometry. Once I get to the bottom of that bug, I'll probably add in another SPH sample input file for this geometry to ensure so that the test cases cover more of the code in the SPH calculator. I'll poke you shortly!

wbinventor:
I've pushed some fixes to this but still have some work left to do. One issue I'm seeing is that the `'all'` domains mode does not seem to be stable.  What I see is that once we cycle over all domains and come back to the first domain, the SPH factors do not converge nearly as quickly (they stall near 1E-1) as they do the first time through when they converge to 1E-6 in 2-5 iterations for every domain. @PrezNattyGibbs - when you reviewed this, did you see any issue with "updating the iterate" MGXS library as the outer loop over domains proceeds?

wbinventor:
After some thorough analysis of some complex geometries, I'm satisfied with the SPH scheme and think this PR is finally ready to go. It includes the changes from #239 and should be merged in after that.

wbinventor:
This PR is up-to-date with "develop" and I have no further (expected) changes to make. @PrezNattyGibbs fully reviewed this already, and @samuelshaner did a review earlier, so hopefully this can be merged in soon as well.

samuelshaner:
I'll review and test this out tomorrow.

On Tuesday, January 5, 2016, Will Boyd <notifications@github.com> wrote:

> This PR is up-to-date with "develop" and I have no further (expected)
> changes to make. @PrezNattyGibbs <https://github.com/PrezNattyGibbs>
> fully reviewed this already, and @samuelshaner
> <https://github.com/samuelshaner> did a review earlier, so hopefully this
> can be merged in soon as well.
>
> —
> <https://github.com/mit-crpg/OpenMOC/pull/219#issuecomment-169191184>.
>


-- 
Regards,
Sam Shaner
Graduate Student

sample-input/sph-factors/assembly/sph-cell-avg.py
```diff
@@ -0,0 +1,96 @@
...
+
+# Compute SPH factors
+sph, sph_mgxs_lib, sph_indices = \
+    compute_sph_factors(mgxs_lib, track_spacing=spacing,
+                        num_azim=num_azim, num_threads=num_threads)
+
+# Load the SPH-corrected MGXS library data
+openmoc_materials = load_openmc_mgxs_lib(sph_mgxs_lib, openmoc_geometry)
+
+# Run an eigenvalue calculation with the SPH-corrected modifed MGXS library
```
samuelshaner: modifed -> modified

openmoc/materialize.py
```diff
@@ -355,3 +370,347 @@ def load_openmc_mgxs_lib(mgxs_lib, geometry=None):
...
+                solver.setFixedSourceByFSR(fsr_id, group+1, source)
+
+    return openmc_fluxes
+
+
+##
+# @brief Apply SPH factors to an OpenMC MGXS library
+# @details This is a helper routine for openmoc.process.compute_sph_factors(...)
+# @param mgxs_lib an openmc.mgxs.Library object
+# @param sph a NumpPy array of SPH factors for each energy group for this domain
```
samuelshaner: Very small point, but you might consider changing the order of `sph` and `geometry` here to reflect the order of the arguments to the function.

sample-input/sph-factors/pin-cell/sph-factors.py
```diff
@@ -0,0 +1,207 @@
...
+sph, sph_mgxs_lib, sph_indices = \
+    openmoc.materialize.compute_sph_factors(mgxs_lib, track_spacing=spacing,
+                                            num_azim=num_azim,
+                                            num_threads=num_threads)
+
+# Load the SPH-corrected MGXS library data
+materials = \
+    openmoc.materialize.load_openmc_mgxs_lib(sph_mgxs_lib, openmoc_geometry)
+
+# Run an eigenvalue calculation with the SPH-corrected modifed MGXS library
```
samuelshaner: ditto

src/Material.cpp
```diff
@@ -414,9 +478,6 @@ void Material::setNumEnergyGroups(const int num_groups) {
 
     if (_chi != NULL)
       MM_FREE(_chi);
-
-    if (_fiss_matrix != NULL)
-      MM_FREE(_fiss_matrix);
```
samuelshaner: Is there a reason why the frees for `_fiss_matrix` have been removed?

sample-input/sph-factors/assembly/sph-cell-avg.py
wbinventor: Got it

openmoc/materialize.py
wbinventor: Great point, we certainly want the comments to be in the right order (esp. if we switch to docstring comments as I think we should).

src/Material.cpp
wbinventor: Simply to be consistent with the fact that `_fiss_matrix` is not deallocated below for non-aligned data. There is no need to deallocate `_fiss_matrix` here since the `Material::buildFissionMatrix()` will deal with deallocation / re-sizing of the array as needed.

sample-input/sph-factors/assembly/compute-mgxs.py
```diff
@@ -0,0 +1,58 @@
+import copy
+import numpy as np
+import openmc.mgxs
+from infermc.materialize import differentiate_mgxs_lib
```
samuelshaner: The materialize module seems to only be in the `sph` branch of your fork of InferMC. Could this branch be pushed to the master repo or merged into the master repo's develop branch? In case someone tries to pick this up down the road and sees the infermc dependency, it would be nice if this capability comes with the repo they are likely to clone to avoid confusion.

sample-input/sph-factors/assembly/sph-region-avg.py
```diff
@@ -0,0 +1,119 @@
+import matplotlib.pyplot as plt
```
samuelshaner: Have you run into any problems running this input file on clusters? I'm running it on AWS and getting the error shown below. I'm guessing it has to do with matplotlib no longer having a headless backend or something. 

  File "sph-region-avg.py", line 91, in <module>
    openmoc.plotter.plot_materials(openmoc_geometry)
  File "/home/shaner/.local/lib/python2.7/site-packages/openmoc/plotter.py", line 238, in plot_materials
    plot_spatial_data(materials, plot_params)
  File "/home/shaner/.local/lib/python2.7/site-packages/openmoc/plotter.py", line 919, in plot_spatial_data
    fig = plt.figure()
  File "/usr/lib/pymodules/python2.7/matplotlib/pyplot.py", line 423, in figure
    **kwargs)
  File "/usr/lib/pymodules/python2.7/matplotlib/backends/backend_tkagg.py", line 79, in new_figure_manager
    return new_figure_manager_given_figure(num, figure)
  File "/usr/lib/pymodules/python2.7/matplotlib/backends/backend_tkagg.py", line 87, in new_figure_manager_given_figure
    window = Tk.Tk()
  File "/usr/lib/python2.7/lib-tk/Tkinter.py", line 1767, in __init__
    self.tk = _tkinter.create(screenName, baseName, className, interactive, wantobjects, useTk, sync, use)
_tkinter.TclError: no display name and no $DISPLAY environment variable

sample-input/sph-factors/assembly/compute-mgxs.py
wbinventor: Yes, I forgot that I left this reference to `differentiate_mgxs_lib` in here (which is still very much bleeding edge code at this point). I really don't want there to be any reference to InferMC in this new code, but have no way of differentiate MGXS libraries otherwise. Perhaps I should just remove the assembly SPH problem entirely - do you have a preference @samuelshaner @geogunow?

sample-input/sph-factors/assembly/sph-region-avg.py
wbinventor: I actually haven't run this problem on a cluster, but I have seen this error before (when debugging an earlier implementation of the `openmoc.plotter` module for AWS in 2013). In particular, we avoid this issue with Matplotlib in `openmoc.plotter` with [`matplotlib.use('Agg')`](https://github.com/mit-crpg/OpenMOC/blob/develop/openmoc/plotter.py#L18) and [`plt.ioff()`](https://github.com/mit-crpg/OpenMOC/blob/develop/openmoc/plotter.py#L18) here.

Since this is one (the only?) example input file which uses Matplotlib directly rather than via `openmoc.plotter`, I think all I need to do is move my `matplotlib` import after the `openmoc.plotter` import. I'll try that and see if it works.

samuelshaner: Yep, that fixes it!

wbinventor: Great! Check to see if my latest commit does the trick for all of the input
files.

On Fri, Jan 8, 2016 at 9:54 AM, sam <notifications@github.com> wrote:

> In sample-input/sph-factors/assembly/sph-region-avg.py
> <https://github.com/mit-crpg/OpenMOC/pull/219#discussion_r49195111>:
>
> > @@ -0,0 +1,119 @@
> > +import matplotlib.pyplot as plt
>
> Yep, that fixes it!
>
> —
> <https://github.com/mit-crpg/OpenMOC/pull/219/files#r49195111>.
>



-- 
Will Boyd
Nuclear Science & Engineering
Cell: 423.413.8469
wbinventor@gmail.com

samuelshaner: Works great now - thanks!

samuelshaner:
Have you tried running any of the input files with Python 3? Python 3 always gives me problems whether I run it locally or on AWS (which could just be due to me only using Python 3 for testing purposes and only installing/updating packages when something fails) and I'm currently getting the following error when running the inputs with Python 3.

  File "sph-factors.py", line 54, in <module>
    fluxes_no_sph = openmoc.process.get_scalar_fluxes(solver)
  File "/Users/sam/Library/Python/3.4/lib/python/site-packages/openmoc/process.py", line 85, in get_scalar_fluxes
    fluxes[fsr, group-1] = solver.getFlux(fsr, group)
  File "/Users/sam/Library/Python/3.4/lib/python/site-packages/openmoc/openmoc.py", line 1445, in getFlux
    return _openmoc.Solver_getFlux(self, fsr_id, group)
TypeError: in method 'Solver_getFlux', argument 2 of type 'int'

sample-input/sph-factors/assembly/compute-mgxs.py
samuelshaner: I don't have any strong preference for keeping/removing the assembly SPH problem, so I'll just defer to you on whether you want it to be kept it in; if it is kept in, perhaps there could be a try/catch around the import of InferMC and an error message printed out informing the user of the need for InferMC to run the input file?

wbinventor:
This now works for Python 3. I also removed the assembly SPH sample input files due to the InferMC dependency.

PR #221, opened 2015-11-14
Fixed c5g7 materials and profile input file issues

samuelshaner:
This PR fixes a few issues in the input files, namely:

- Fixed issue in `c5g7-mgxs-hdf5.py` with Control Rod material not being properly inserted in hdf 5 file
- Updated loading of materials in `c5g7-3d-rodded-B.py`
- Updated profile input files to reflect changes to `Lattice:setUniverses()` method and change of `Circle` surface to `ZCylinder`

PR #222, opened 2015-11-17
Split segments hotfix

samuelshaner:
This PR builds on PR #221 and fixes a bug in `TrackGenerator::splitSegments()` in setting the CMFD surface and corner ids for segments created from splitting a segment with an optical length greater than the max optical length. I'm not an expert on how c++ data structures are treated under the hood, but here's what I believe is happening in the `TrackGenerator::splitSegments()` method. First, a pointer to current segment is retrieved from the track (see below). Remember, the pointer references the location in memory where the item is located and is not tied to the std::vector<segment> object, so if c++ decides the resize the Track's `_segments` vector at some later time, the pointer could point to garbage. If the segment needs to be split, new segments will be created and inserted into the `Track` object's `std::vector<struct> _segments` attribute. When this is done, c++ might have to resize the `_segments` vector, placing it in another location in memory (I believe). If the vector is resized, `curr_segment` will no longer point to the intended segment in memory and the next segment that is created will be given incorrect CMFD surface and corner ids. This PR follows the same procedure used in setting the other variables of the segment (e.g. `_material`, `_length`, `_region_id`) by saving the values of the CMFD surface and corner ids before the new segments are created.

**Old code with bug in setting CMFD surface and corner ids of cut segments**

    /* Get pointer to current segment */
    curr_segment = _tracks[i][j].getSegment(s);

    ...

    /* Split the segment into sub-segments */
    for (int k=0; k < min_num_cuts; k++) { 

        ....

        /* Assign CMFD surface boundaries */
        if (k == 0) {
          new_segment->_cmfd_surface_bwd = curr_segment->_cmfd_surface_bwd;
          new_segment->_cmfd_corner_bwd = curr_segment->_cmfd_corner_bwd;
        }

        if (k == min_num_cuts-1) {
          new_segment->_cmfd_surface_fwd = curr_segment->_cmfd_surface_fwd;
          new_segment->_cmfd_corner_fwd = curr_segment->_cmfd_corner_fwd;
        }

        /* Insert the new segment to the Track. */
        _tracks[i][j].insertSegment(s+k+1, new_segment);
    }

    /* Remove the original segment from the Track */
    _tracks[i][j].removeSegment(s);


geogunow:
Everything checks out

PR #223, opened 2015-11-18
Updated IPython Notebook example input file

wbinventor:
This PR fixes a few API calls in the "sample-input/ipython-notebook" example. Thanks to @bforget for noticing that the call to the `openmoc.materialize` module was out-of-date.


samuelshaner:
While not directly related to the problem this PR addresses, do you see two scale bars for the fission rates plot in the ipython notebook? I see two scale bars in ipython, but not when I open up the image in an image viewer.

wbinventor:
I see the same thing with double scale bars for the final fission rate plot. Interestingly, the double color bar does not seem to be a part of the image file that is created by the plotter module, so it must be something that is created when the image is displayed within ipython notebook. I'm not particularly concerned about it since it's likely some sort of matplotlib-specific quirk, though it would be nice to eliminate it.

PR #224, opened 2015-11-19
Added surface metadata to Cell->toString.

PrezNattyGibbs:
This is a simple PR that adds `Surface` metadata to the `Cell->toString()` string.  This fixes a bug I discovered that if you change `Surface` parameters between runs, OpenMOC will still read a now-incorrect track file.  With this change, OpenMOC can detect the change and will re-track the geometry.

### DIFF
--- a/src/Cell.cpp
+++ b/src/Cell.cpp
@@ -946,11 +946,11 @@ std::string Cell::toString() {
 
   string << ", # surfaces = " << getNumSurfaces();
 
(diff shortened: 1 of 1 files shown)


wbinventor:
Wahoo! We are wringing every last bit of productivity out of you that we can before you leave!

wbinventor:
Everything checks out for me. Thanks for the fix @PrezNattyGibbs!

PR #225, opened 2015-11-23
Cell Rotations and Translations

wbinventor:
This PR introduces cell rotations and translations. These features enable us to model the fully-detailed (3D) BEAVRS benchmark generated by the OpenCG BEAVRS builder for the first time! 

Unfortunately, adding in rotations/translations was a little more difficult than I had expected (but not overly cumbersome). Fortunately, the process of putting in these features required some refactoring of the ray tracing code - but it keeps getting better and better and is far more methodical than a few years ago.

In particular, this PR introduces length 3 `_rotation` and `_translation` arrays for the `Cell` class which may be optionally set by the user. The ray tracing code in the `Geometry` and `Universe` classes has been refactored such that all updates to `LocalCoords` positions take place directly within the `LocalCoords` object (which is a cleaner abstraction than before). In addition, the 2D angle `phi` wrt the x-axis is now an attribute of `LocalCoords`, where it is defined based on a track's starting angle and the rotations applied to any `Cells` found during CSG tree traversal.

With the changes made here, it is quite simple to build a full-core OpenMOC model of the 3D BEAVRS core. In particular, with an installation of the BEAVRS model builder from the mit-crpg/PWR_benchmarks repo, along with OpenCG (a pre-requisite for the BEAVRS builder), one can instantiate an OpenMOC model of BEAVRS as follows:

```python
import opencg
from beavrs.builder import BEAVRS
from openmoc.compatible import get_openmoc_geometry

# Instantiate a BEAVRS OpencG model from mit-crpg/PWR_benchmarks
beavrs = BEAVRS(nndc_xs=True)

# Construct an OpenCG Geometry for the BEAVRS reactor model
opencg_geometry = opencg.Geometry()
opencg_geometry.root_universe = beavrs.main_universe

# Export the 3D OpenCG Geometry to OpenMOC
openmoc_geometry = get_openmoc_geometry(opencg_geometry)

# Voila!
```

Previously, if one attempted to export the model as above, the baffles were not placed in the correct positions. This is because the baffles have rotations applied. Now that the baffles can be properly treated, the model above can be plotted as shown below and the baffles look as they should. 

**Full core**

**Upper right core**

Note that the OpenMOC `Geometry` created above will *not* have a bounding box applied, and hence is not yet ready for ray tracing. In order to make the `Geometry` "ray tracing-ready", one would must create a `Cell` outside the core vessel filled by a dummy `Material` and bounded by the core vessel's `ZCylinder` and some bounding `XPlanes` and `YPlanes` with `VACUUM` boundary conditions.

### DIFF
--- a/openmoc/compatible/opencg_compatible.py
+++ b/openmoc/compatible/opencg_compatible.py
@@ -385,10 +385,11 @@ def get_opencg_cell(openmoc_cell):
       opencg_cell.fill = get_opencg_universe(fill)
 
   if openmoc_cell.isRotated():
-    rotation = np.array([openmoc_cell.getPhi(),
-                         openmoc_cell.getTheta(),
-                         openmoc_cell.getPsi()])
+    rotation = openmoc_cell.getRotation(3)
     opencg_cell.rotation = rotation * -1
+  if openmoc_cell.isTranslated():
+    rotation = openmoc_cell.getTranslation(3)
+    opencg_cell.translation = translation
 
   surfaces = openmoc_cell.getSurfaces()
 
@@ -597,6 +598,9 @@ def get_openmoc_cell(opencg_cell):
   if opencg_cell.rotation is not None:
     rotation = np.asarray(opencg_cell.rotation, dtype=np.float64) * -1
     openmoc_cell.setRotation(rotation)
+  if opencg_cell.translation is not None:
+    translation = np.asarray(opencg_cell.translation, dtype=np.float64)
+    openmoc_cell.setTranslation(translation)
 
   surfaces = opencg_cell.surfaces
 
--- a/openmoc/compatible/opencg_compatible.py
+++ b/openmoc/compatible/opencg_compatible.py
@@ -768,7 +768,6 @@ def get_openmoc_lattice(opencg_lattice):
   universes = opencg_lattice.universes
 
   # Initialize an empty array for the OpenMOC nested Universes in this Lattice
-  # The Lattice universe array is stored x-y-z in OpenCG, but z-y-x in OpenMOC
   universe_array = np.ndarray(tuple(dimension[::-1]), dtype=openmoc.Universe)
 
   # Create OpenMOC Universes for each unique nested Universe in this Lattice
--- a/openmoc/compatible/opencg_compatible.py
+++ b/openmoc/compatible/opencg_compatible.py
@@ -587,6 +587,10 @@ def get_openmoc_cell(opencg_cell):
   else:
     openmoc_cell.setFill(get_openmoc_material(fill))
 
+  if opencg_cell.rotation is not None:
+    rotation = np.asarray(opencg_cell.rotation, dtype=np.float64) * -1
+    openmoc_cell.setRotation(rotation)
+
   surfaces = opencg_cell.surfaces
 
   for surface_id in surfaces:
@@ -753,8 +757,8 @@ def get_openmoc_lattice(opencg_lattice):
   universes = opencg_lattice.universes
 
   # Initialize an empty array for the OpenMOC nested Universes in this Lattice
-  universe_array = np.ndarray(tuple(np.array(dimension[0:2])), \
-                              dtype=openmoc.Universe)
+  # The Lattice universe array is stored x-y-z in OpenCG, but z-y-x in OpenMOC
+  universe_array = np.ndarray(tuple(dimension[::-1]), dtype=openmoc.Universe)
 
   # Create OpenMOC Universes for each unique nested Universe in this Lattice
   unique_universes = opencg_lattice.get_unique_universes()
@@ -763,14 +767,15 @@ def get_openmoc_lattice(opencg_lattice):
     unique_universes[universe_id] = get_openmoc_universe(universe)
 
   # Build the nested Universe array
-  for y in range(dimension[1]):
-    for x in range(dimension[0]):
-      universe_id = universes[0][y][x].id
-      universe_array[x][y] = unique_universes[universe_id]
+  for z in range(dimension[2]):
+    for y in range(dimension[1]):
+      for x in range(dimension[0]):
+        universe_id = universes[z][y][x].id
+        universe_array[z][dimension[1]-y-1][x] = unique_universes[universe_id]
 
   openmoc_lattice = openmoc.Lattice(lattice_id, name)
   openmoc_lattice.setWidth(width[0], width[1], width[2])
-  openmoc_lattice.setUniverses([universe_array.tolist()])
+  openmoc_lattice.setUniverses(universe_array.tolist())
   openmoc_lattice.setOffset(offset[0], offset[1], offset[2])
 
   # Add the OpenMOC Lattice to the global collection of all OpenMOC Lattices
--- a/openmoc/compatible/opencg_compatible.py
+++ b/openmoc/compatible/opencg_compatible.py
@@ -386,7 +386,7 @@ def get_opencg_cell(openmoc_cell):
 
   if openmoc_cell.isRotated():
     rotation = openmoc_cell.getRotation(3)
-    opencg_cell.rotation = rotation * -1
+    opencg_cell.rotation = rotation
   if openmoc_cell.isTranslated():
     rotation = openmoc_cell.getTranslation(3)
     opencg_cell.translation = translation
@@ -596,7 +596,7 @@ def get_openmoc_cell(opencg_cell):
     openmoc_cell.setFill(get_openmoc_material(fill))
 
   if opencg_cell.rotation is not None:
-    rotation = np.asarray(opencg_cell.rotation, dtype=np.float64) * -1
+    rotation = np.asarray(opencg_cell.rotation, dtype=np.float64)
     openmoc_cell.setRotation(rotation)
   if opencg_cell.translation is not None:
     translation = np.asarray(opencg_cell.translation, dtype=np.float64)
--- a/openmoc/compatible/opencg_compatible.py
+++ b/openmoc/compatible/opencg_compatible.py
@@ -374,21 +374,28 @@ def get_opencg_cell(openmoc_cell):
   name = openmoc_cell.getName()
   opencg_cell = opencg.Cell(cell_id, name)
 
-  fill = openmoc_cell.getFill()
-  if (openmoc_cell.getType == openmoc.MATERIAL):
+  if (openmoc_cell.getType() == openmoc.MATERIAL):
+    fill = openmoc_cell.getFillMaterial()
     opencg_cell.fill = get_opencg_material(fill)
   elif (openmoc_cell.getType() == openmoc.FILL):
+    fill = openmoc_cell.getFillUniverse()
     if isinstance(fill, openmoc.Lattice):
       opencg_cell.fill = get_opencg_lattice(fill)
     else:
       opencg_cell.fill = get_opencg_universe(fill)
+
+  if openmoc_cell.isRotated():
+    rotation = np.array([openmoc_cell.getPhi(),
+                         openmoc_cell.getTheta(),
+                         openmoc_cell.getPsi()])
+    opencg_cell.rotation = rotation * -1
 
   surfaces = openmoc_cell.getSurfaces()
 
   for surf_id, surface_halfspace in surfaces.items():
     halfspace = surface_halfspace._halfspace
     surface = surface_halfspace._surface
-    opencg_cell.addSurface(get_opencg_surface(surface), halfspace)
+    opencg_cell.add_surface(get_opencg_surface(surface), halfspace)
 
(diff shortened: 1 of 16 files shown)


samuelshaner:
@wbinventor - this is great! I'll give this a full review later this week.

openmoc/compatible/opencg_compatible.py
```diff
@@ -374,21 +374,29 @@ def get_opencg_cell(openmoc_cell):
...
   elif (openmoc_cell.getType() == openmoc.FILL):
+    fill = openmoc_cell.getFillUniverse()
     if isinstance(fill, openmoc.Lattice):
       opencg_cell.fill = get_opencg_lattice(fill)
     else:
       opencg_cell.fill = get_opencg_universe(fill)
 
+  if openmoc_cell.isRotated():
+    rotation = openmoc_cell.getRotation(3)
+    opencg_cell.rotation = rotation * -1
```
samuelshaner: What's the reason for the -1 here and on line 599?

src/Cell.cpp
```diff
@@ -149,6 +156,117 @@ Universe* Cell::getFillUniverse() {
...
+ * @return the rotation angle about the z-axis
+ */
+double Cell::getPsi() {
+  return _rotation[2];
+}
+
+
+/**
+ * @brief Return pointer to array for the rotation matrix.
+ * @returns a pointer to an array of rotation angles
```
samuelshaner: returns -> return

openmoc/compatible/opencg_compatible.py
```diff
@@ -753,8 +768,8 @@ def get_openmoc_lattice(opencg_lattice):
   universes = opencg_lattice.universes
 
   # Initialize an empty array for the OpenMOC nested Universes in this Lattice
-  universe_array = np.ndarray(tuple(np.array(dimension[0:2])), \
-                              dtype=openmoc.Universe)
+  # The Lattice universe array is stored x-y-z in OpenCG, but z-y-x in OpenMOC
```
samuelshaner: Looking at the master branch of OpenCG, it appears OpenCG stores the Lattice universe array in z-y-x, just like in OpenMOC. For instance, line 1045 of `opencg/universe.py` indexes the universes array with the z-component first:

    return self._universes[lat_z][lat_y][lat_x]

Is OpenCG indeed storing the universes array as x-y-z?

src/Cell.cpp
```diff
@@ -388,6 +506,88 @@ void Cell::setFill(Universe* fill) {
...
+ * @code
+ *          rotation = numpy.array([0., 0., 90.])
+ *          cell = openmoc.Cell()
+ *          cell.setRotation(rotation)
+ * @endcode
+ *
+ * @param rotation the array of rotation angles
+ * @param num_axes the number of axes (this must always be 3)
+ */
+void Cell::setRotation(double* rotation, int num_axes) {
```
samuelshaner: Googling 3D rotations, it looks like there are a few different notations in common use [[LINK]](http://mathworld.wolfram.com/EulerAngles.html). Could you add some commentary in the details on which notation is being used and perhaps a reference for where to find more info on this notation?

src/Cell.cpp
```diff
@@ -388,6 +506,88 @@ void Cell::setFill(Universe* fill) {
...
+  for (int i=0; i < 3; i++)
+    _rotation[i] = rotation[i];
+
+  /* Compute rotation angles in x,y,z directions */
+  double phi = -_rotation[0] * M_PI / 180.;
+  double theta = -_rotation[1] * M_PI / 180.;
+  double psi = -_rotation[2] * M_PI / 180.;
+
+  /* Calculate rotation matrix based on angles given */
+  /* Indexed by (y,x) since the universe array is indexed by (z,y,z) */
```
samuelshaner: I'm having a hard time understanding this comment. Why only two coordinates in the first parentheses and three in the second parentheses? Also, what do you think about making the matrix a 2D array and stating whether it is row-major or column-major?

src/Cell.cpp
```diff
@@ -944,6 +1148,15 @@ std::string Cell::toString() {
   else
     string << ", type = UNFILLED";
 
+  if (_rotated) {
```
samuelshaner: What do you think about having parentheses around the rotation and translation elements in the string? Also, what about line breaks after the rotation and translation substrings to break them up? I feel like this would make it more readable.

samuelshaner:
The rotations are input and stored in degrees, but angles are for the most part handled internally in radians. The `Cell` and `LocalCoords` objects both have `getPhi()` methods, but `Cell` returns phi in degrees while `LocalCoords` returns phi in radians. At the surface, this can be a bit confusing. I think handling angles internally in one common set of units (I think radians would be most appropriate) would make the code easier to understand and avoid any conversion errors down the road, but allowing the user to input angles for rotations in degrees or radians (which a function parameter indicating what units are being input) would be a reasonable compromise from a user-input standpoint. What are your thoughts on the radians/degrees mixing in the code?

src/Universe.cpp
```diff
@@ -535,27 +534,45 @@ Cell* Universe::findCell(LocalCoords* coords) {
...
+	  double y = coords->getY();
+	  double z = coords->getZ();
+	  double* matrix = cell->getRotationMatrix();
+	  double new_x = matrix[0] * x + matrix[1] * y + matrix[2] * z;
+	  double new_y = matrix[3] * x + matrix[4] * y + matrix[5] * z;
+	  double new_z = matrix[6] * x + matrix[7] * y + matrix[8] * z;
+	  next_coords->setX(new_x);
+	  next_coords->setY(new_y);
+	  next_coords->setZ(new_z);
+          next_coords->incrementPhi(cell->getPsi() * M_PI / 180.);
```
samuelshaner: indented one too many spaces

openmoc/numpy_typemaps.i
```diff
@@ -24,6 +24,23 @@
  * using NumPy arrays */
 %apply (double* IN_ARRAY1, int DIM1) {(double* xs, int num_groups)}
 
+/* The typemap used to match the method signature for the Cell rotation
+ * angle setter method. This allows users to set the rotation angles
+ * using NumPy arrays */
+%apply (double* IN_ARRAY1, int DIM1) {(double* rotation, int num_axes)}
```
geogunow: Is there a good reason for using arrays rather than vectors everywhere? If this was a vector, no typemap would be needed.

src/Cell.cpp
```diff
@@ -388,6 +506,88 @@ void Cell::setFill(Universe* fill) {
...
+  /* Indexed by (y,x) since the universe array is indexed by (z,y,z) */
+  _rotation_matrix[0] = cos(theta) * cos(psi);
+  _rotation_matrix[1] = cos(theta) * sin(psi);
+  _rotation_matrix[2] = -sin(theta);
+  _rotation_matrix[3] = -cos(phi) * sin(psi) + 
+                        sin(phi) * sin(theta) * cos(psi);
+  _rotation_matrix[4] = cos(phi) * cos(psi) +
+                        sin(phi) * sin(theta) * sin(psi);
+  _rotation_matrix[5] = sin(phi) * cos(theta);
+  _rotation_matrix[6] = sin(phi) * sin(psi) + \
```
geogunow: Is the slash needed? The other liens suggest it is not.

src/Universe.cpp
```diff
@@ -1194,13 +1211,15 @@ Cell* Lattice::findCell(LocalCoords* coords) {
   /* Create a new LocalCoords object for the next level Universe */
   LocalCoords* next_coords;
 
-  if (coords->getNext() == NULL)
+  if (coords->getNext() == NULL) {
     next_coords = new LocalCoords(next_x, next_y, next_z);
+  }
```
geogunow: Are these curly braces really needed? Let's be consistent on the if/else style.

openmoc/numpy_typemaps.i
wbinventor: If this is indeed the case, then I'm sold. I'm not going to make vectors the standard in this PR, but certainly think it would be nice to start using them in place of arrays in some if not all cases, esp. if it would simplify our swig typemaps.

src/Cell.cpp
wbinventor: fixed

src/Cell.cpp
wbinventor: Oops, I'm just going to remove this comment altogether in place of a more detailed comment.

src/Cell.cpp
wbinventor: Sure, that's a good idea. This implementation uses the "x-y-z pitch-roll-yaw" convention based on eqns 51-59 in the link you sent. I'll include this in a more detailed comment now.

src/Cell.cpp
wbinventor: Nope, it was leftover from the Python implementation in OpenCG so I'll remove it now.

openmoc/compatible/opencg_compatible.py
wbinventor: Haha, it was a little hack while I was trying to figure things out. I've removed it now as I've straightened out things in the roll-pitch-yaw convention of the rotation matrix below.

src/Cell.cpp
wbinventor: I've clarified the comment as well as revised the code here to match up with the equations on Wolfram.

src/Cell.cpp
wbinventor: Agreed, all good ideas which I will add in now.

wbinventor: Actually, I'm leaving out the line breaks since we haven't used line breaks in any of our `toString()` methods. I do think it would be nice to clean up the strings output from these methods with line breaks in a consistent manner at some point in the future.

src/Universe.cpp
wbinventor: Weird, it doesn't show up that way in emacs. I removed all whitespace and added it back in so perhaps it will show up properly now.

src/Universe.cpp
wbinventor: Nope, not needed now though they were for an early implementation. I'll remove them now.

openmoc/compatible/opencg_compatible.py
wbinventor: You are right that it is indexed by z-y-x in both codes (or at least should be). I made this comment to myself while I was debugging since I had convinced myself otherwise at one point. I'll revert back to the original comment now.

wbinventor: On a related note though, OpenCG's `dimension` attribute is stored as x-y-z since this (may) make more intuitive sense to the user. Hence the need to reverse this array here when initializing OpenMOC's universes array.

wbinventor:
I've addressed all of the comments made thus far and think this is ready for a final review.

wbinventor:
I've addressed all of the comments made thus far and think this is ready for a final review.

samuelshaner:
Thanks for updating the PR @wbinventor. I had a comment on mixing of angle units (radians vs degrees) in the code that it looks like you might have missed.

src/Cell.cpp
wbinventor: To your comment about making the matrix a 2D rather than a 1D array - I considered this, but because we do not have a general purpose matrix class, I decided against it. And I didn't want to implement one or introduce an additional dependency to support a matrix class for such a small matrix. If we want general purpose matrices, there are other places in the code which might stand to benefit much more than this (i.e., CMFD), or at least that was my thinking.

wbinventor:
@samuelshaner - the rotation angles are now stored internally in radians. The setter and getter routines for the angles now accept a `std::string units` parameter which defaults to `"degrees"` but which may alternatively accept `"radians"`.

openmoc/numpy_typemaps.i
```diff
@@ -24,6 +24,23 @@
...
+
+/* The typemap used to match the method signature for the Cell translation
+ * setter method. This allows users to set translations using NumPy arrays */
+%apply (double* IN_ARRAY1, int DIM1) {(double* translation, int num_axes)}
+
+/* The typemap used to match the method signature for the Cell's
+ * getter method for rotations used by the OpenCG compatibility module. */
+%apply (double* ARGOUT_ARRAY1, int DIM1) {(double* rotations, int num_axes)}
```
samuelshaner: does this signature need another attribute for the units?

wbinventor: I wondered that too, but it doesn't seem too since it works with my test case.

samuelshaner: Hmmm...maybe swig just ignores all arguments after those included in the signature.

wbinventor: That's what I think. But this is related to the way NumPy typemaps work with SWIG - and that isn't particularly well documented IMHO.

samuelshaner:
Just tested this out and everything looks good to merge!

PR #228, opened 2015-12-06
Exaggerated Fission Rate Color Scheme

wbinventor:
This PR adds a new boolean `transparent_zeros` parameter to the `openmoc.plotter.plot_fission_rates(...)` routine. This can be used to make all regions with a zero fission rate (*e.g.*, gap, clad, moderator) transparent in the plot (and excluded from the colorbar). This can make it easier to compare the relative fission rates between different fuel pins in an assembly.

For example, let's say we have the following fuel assembly of cells:

The current code would plot the fission rates with dark blue for those regions without fission as follows:

The above plot may still be generated with the new code introduced here when `transparent_zeros=False` (the default). If instead the user sets `transparent_zeros=True` one will obtain the following more useful plot:

@geogunow - you want to review this easy PR and take the load off of @samuelshaner?

### DIFF
--- a/openmoc/plotter.py
+++ b/openmoc/plotter.py
@@ -1,4 +1,4 @@
-#
+##
 # @file plotter.py
 # @package openmoc.plotter
 # @brief The plotter module provides utility functions to plot data from
@@ -941,10 +941,12 @@ def plot_energy_fluxes(solver, fsrs, group_bounds=None, norm=True, loglog=True):
 # @endcode
 #
 # @param solver a Solver object that has converged the source for the Geometry
+# @param transparent_zeros make regions without fission transparent
 # @param gridsize an optional number of grid cells for the plot
 # @param xlim optional list/tuple of the minimim/maximum x-coordinates
 # @param ylim optional list/tuple of the minimim/maximum y-coordinates
-def plot_fission_rates(solver, gridsize=250, xlim=None, ylim=None):
+def plot_fission_rates(solver, transparent_zeros=False, gridsize=250,
+                       xlim=None, ylim=None):
 
   global subdirectory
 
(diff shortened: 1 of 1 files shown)


samuelshaner:
The fission rate plots with transparent zeros look so much better!!!

wbinventor:
I agree! It's almost like there are clusters or something...

samuelshaner:
Haha yeah!

geogunow:
This is such a huge PR, will definitely help take the load off Sam.

wbinventor:
Every little bit helps.

geogunow:
Everything looks good as far as the changes go. I'm going to merge as soon as I look at all the example files to make sure they are using the correct syntax.

wbinventor:
Ok, I'm not sure that we plot the fission rates in any of the example input files. The feature is documented on the [website](https://mit-crpg.github.io/OpenMOC/usersguide/processing.html#fission-rate-visualization), but that will need to be extended with a separate PR to the "gh-pages" branch. Plus, I'm inclined to wait on the documentation update till @samuelshaner you and I can discuss a concerted effort to update our badly out-of-date documentation on the website.

samuelshaner:
`c5g7-cmfd.py` plots the fission rates.

fission-rates-z-0 0

openmoc/plotter.py
```diff
@@ -941,10 +941,12 @@ def plot_energy_fluxes(solver, fsrs, group_bounds=None, norm=True, loglog=True):
 # @endcode
 #
 # @param solver a Solver object that has converged the source for the Geometry
+# @param transparent_zeros make regions without fission transparent
 # @param gridsize an optional number of grid cells for the plot
 # @param xlim optional list/tuple of the minimim/maximum x-coordinates
 # @param ylim optional list/tuple of the minimim/maximum y-coordinates
-def plot_fission_rates(solver, gridsize=250, xlim=None, ylim=None):
+def plot_fission_rates(solver, transparent_zeros=False, gridsize=250,
```
geogunow: I would like to actually see transparent_zeros be True by default

wbinventor:
Good call @geogunow, I made it `True` by default.

geogunow:
All input files look good, all plotting transparent zeros by default

PR #231, opened 2015-12-10
Hotfix for CMFD Re-initialization

wbinventor:
This very minor PR fixes two issues in CMFD related to re-initialization. In particular, a user will currently observe a segmentation when using same `CMFD`, `Geometry`, `TrackGenerator` and `Solver` objects for two simulations. This is due to the way a few of `CMFD's` attributes are deleted during re-initialization. This PR fixes these issues and makes it possible to run as many simulations as one wishes to with a single `CMFD` object (*e.g.*, with different MGXS libraries).

### DIFF
--- a/src/Cmfd.cpp
+++ b/src/Cmfd.cpp
@@ -905,11 +905,11 @@ void Cmfd::initializeCurrents() {
 
   /* Delete old Cmfd surface currents array it it exists */
   if (_surface_currents != NULL)
-    delete [] _surface_currents;
+    delete _surface_currents;
 
   /* Delete old Cmfd corner currents array it it exists */
   if (_corner_currents != NULL)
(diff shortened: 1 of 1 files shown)


src/Cmfd.cpp
```diff
@@ -905,11 +905,11 @@ void Cmfd::initializeCurrents() {
 
   /* Delete old Cmfd surface currents array it it exists */
```
samuelshaner: While you are fixing this part of the code, can you also correct the comment here to fix this typo and the one for the corner currents?

it it -> if it

wbinventor: Sure, done

samuelshaner:
Tested this out and it works great. Thanks for this fix @wbinventor!

PR #236, opened 2015-12-10
changed eigenvalueSolve to normalize new source instead of old source

samuelshaner:
This PR makes a small, but significant change to the `eigenvalueSolve()` function in `linalg.cpp`. While, this would seemingly not have an effect on the residual computed during the eigenvalue solve, in practice the residual is affected by this change. Additionally, this slightly reduces the amount of computation in the eigenvalue solve.


wbinventor:
This seems reasonable to me, and works fine.

PR #238, opened 2015-12-11
Optimized Eigenvalue Calculation

wbinventor:
This PR removes the calculation of the old (prior iteration's) fission source when computing the eigenvalue. This is unnecessary since the (old) flux is scaled after each power iteration such that the total (old) fission source is unity. @samuelshaner do you want to look at this since we discovered this together?

### DIFF
--- a/src/CPUSolver.cpp
+++ b/src/CPUSolver.cpp
@@ -605,31 +605,11 @@ void CPUSolver::computeKeff() {
   FP_PRECISION* sigma;
   FP_PRECISION volume;
 
-  FP_PRECISION old_fission, new_fission;
+  FP_PRECISION fission;
   FP_PRECISION* FSR_rates = new FP_PRECISION[_num_FSRs];
   FP_PRECISION* group_rates = new FP_PRECISION[_num_threads * _num_groups];
 
   /* Compute the old nu-fission rates in each FSR */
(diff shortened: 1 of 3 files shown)


samuelshaner:
Sure, I'll review it tonight.

samuelshaner:
@wbinventor - have you been able to compile this with the intel compiler? I tried compiling with icpc on my mac and on nsecluster and was unsuccessful both places. Do you know if there are any environment variables or other things that need to be done to compile with icpc on nsecluster?

wbinventor:
Hmm, what issue do you see with icpc? I just compiled on nsecluster with the following command:

```bash
python setup.py install --user --cc=icpc
```

If you are seeing a runtime error, note that you have to delete any track files generated with a different compiled version of the code since the FSRs are hashed differently between compilers.

samuelshaner:
I was getting an error on NSECluser when I used ccache. Removing the `--with-ccache` option allowed me to compile. 

Everything looks good to me!

wbinventor:
Oh yes, ccache doesn't play nice with the intel compiler. From what I recall when I put support in for ccache, it is really built around gcc (in particular, the command line args for gcc). Anyway, thanks for testing things out with our different supported compilers!

Issue #230, opened 2015-12-07
Automate Pin Cell Coolant Radial Discretization


wbinventor:
We really need to make it simple and easy to discretize the coolant around a fuel pin. Although a user can radially discretize fuel pins and the clad with the `Cell.setNumRings(...)` routine, this may not be used to discretize an "unbounded" `Cell` such as the coolant. We should come up with a systematic and agreed upon way to extend our automated radial discretization to the unbounded `Cells`, since it is particularly important for the moderator.

I'm not yet sure how best to go about this yet. But I really need a method implemented to do this to simplify the workflow for the simulations I'm currently running. If anyone has any ideas on a nice robust way to go about this, post them here!

PrezNattyGibbs:
Equal volume doesn't make a whole lot of sense for unbounded cells. But 
you could definitely do fixed delta radius rings.  I'd recommend putting 
in "setNumRings" and "setRingWidth" for these cells.

The other option would be to make some stencils for common 
discretizations and build those into the code.  So you could request a 
"pin-cell, fine" or "pin-cell, coarse" rather than setting the surfaces 
all yourself.  If we want some weird stencils, this is the way to go, 
but I haven't heard any of us wanting to do that.  Hebert talked about a 
pretty weird stencil he uses in the last talk I saw from him, so it's 
not unreasonable to think about this, but I don't think we need it now.

wbinventor:
@PrezNattyGibbs - I like your idea of a `Cell:setRingWidth(...)` and/or `Cell::setRingWidths(...)` method. One of the downsides of this type of API is that it is very hard-coded to the cylindrical coordinate system (albeit, so are the current `Cell::setNumRings(...)` and `Cell::setNumSectors(...)` methods). We might also add optional `max_radius` and `conserve` parameters to the `Cell::setNumRings(...)` method. These might have defaults like `max_radius=-1` and `preserve="volume"` to preserve the current behavior of equi-volume rings with automated max radius detection. But the parameters could also accept positive floating point values for the maximum radius and the string `"delta radius"`, respectively.

An alternative could be a separate "meshing" module that abstracts the discretization away from the general and arbitrary `Cell` class. I have such an implementation in OpenCG, but it does require more code in the input to discretize a pin with this type of abstraction. Because of this, and since OpenMOC is fairly well focused on cylindrical geometries like those in LWRs, I'm inclined to think that the abstraction - while nice from an OO standpoint - is less preferable to a user than what you propose.

[...]

PR #239, opened 2015-12-12
Unbounded Cell Ring Subdivisions

wbinventor:
This PR addresses #230 and makes it easy to automatically generate radial discretizations within an unbounded `Cell`, such as what one might use to model the moderator in an LWR pin cell. The following shows the new simplified code used to create a moderator cell in the updated example IPython Notebook:

```python
# Moderator rings
moderator = openmoc.Cell()
moderator.setFill(materials['Water'])
moderator.addSurface(+1, fuel_radius)
moderator.setNumRings(3)
moderator.setNumSectors(8)
```


src/Universe.cpp
```diff
@@ -589,23 +589,34 @@ Cell* Universe::findCell(LocalCoords* coords) {
...
-  log_printf(DEBUG, "Subdividing Cells for Universe %d", _id);
+  log_printf(DEBUG, "Subdividing Cells for Universe ID=%d "
+	     "with max radius %f", _id, max_radius);
 
-  std::map<int, Cell*>::iterator iter1;
+  std::map<int, Cell*>::iterator iter;
 
-  while (iter1 != _cells.end()) {
-    for (iter1 = _cells.begin(); iter1 != _cells.end(); ++iter1) {
+  while (iter != _cells.end()) {
```
samuelshaner: Is this while necessary?

src/Universe.cpp
```diff
@@ -1112,6 +1123,34 @@ void Lattice::removeUniverse(Universe* universe) {
...
+  /* Compute minimum radius needed to surround a lattice cell */
+  /* This is used as the maximum radius for all ringified Cells */
+  double radius = sqrt(_width_x * _width_x / 2. + _width_y * _width_y / 2.);
+
+  /* If the lattice pitch is smaller than max_radius parameter, over-ride it */
+  if (radius < max_radius)
+    max_radius = radius;
+
+  /* Subdivide all Cells */
+  while (iter != universes.end()) {
```
samuelshaner: ditto

src/Universe.cpp
```diff
@@ -589,23 +589,34 @@ Cell* Universe::findCell(LocalCoords* coords) {
...
 
-  while (iter1 != _cells.end()) {
-    for (iter1 = _cells.begin(); iter1 != _cells.end(); ++iter1) {
+  while (iter != _cells.end()) {
+    for (iter = _cells.begin(); iter != _cells.end(); ++iter) {
 
-      if (((*iter1).second)->getType() == MATERIAL) {
-        Cell* cell = (*iter1).second;
+      /* Cells filled with Materials */
+      if (((*iter).second)->getType() == MATERIAL) {
```
samuelshaner: I think this can be:

    if (iter->second->getType() == MATERIAL) {

src/Universe.cpp
```diff
@@ -589,23 +589,34 @@ Cell* Universe::findCell(LocalCoords* coords) {
...
+        Cell* cell = (*iter).second;
```
samuelshaner: ditto

    Cell* cell = iter->second;

src/Universe.cpp
```diff
@@ -589,23 +589,34 @@ Cell* Universe::findCell(LocalCoords* coords) {
...
 
         if (cell->getNumRings() > 0 || cell->getNumSectors() > 0)
-          cell->subdivideCell();
+          cell->subdivideCell(max_radius);
+      }
+      /* Cells filled with Universes */
+      else {
+	Universe* fill = ((*iter).second)->getFillUniverse();
```
samuelshaner: ditto

src/Universe.cpp
```diff
@@ -1112,6 +1123,34 @@ void Lattice::removeUniverse(Universe* universe) {
...
+  double radius = sqrt(_width_x * _width_x / 2. + _width_y * _width_y / 2.);
+
+  /* If the lattice pitch is smaller than max_radius parameter, over-ride it */
+  if (radius < max_radius)
+    max_radius = radius;
+
+  /* Subdivide all Cells */
+  while (iter != universes.end()) {
+    for (iter = universes.begin(); iter != universes.end(); ++iter)
+      ((*iter).second)->subdivideCells(max_radius);
```
samuelshaner: ditto

src/Universe.cpp
wbinventor: Not anymore!

src/Universe.cpp
wbinventor: Great catch, it's nice how the discretization algorithms have evolved and simplified to this point.

src/Universe.cpp
wbinventor: Yes to all of these instances

wbinventor:
This has been updated per your comments thus far @samuelshaner

sample-input/benchmarks/c5g7/geometry.py
```diff
@@ -38,21 +38,12 @@
...
-moderator_ring3.setNumSectors(8)
-moderator_ring1.setFill(materials['Water'])
-moderator_ring2.setFill(materials['Water'])
-moderator_ring3.setFill(materials['Water'])
-moderator_ring1.addSurface(+1, fuel_radius)
-moderator_ring1.addSurface(-1, moderator_inner_radius)
-moderator_ring2.addSurface(+1, moderator_inner_radius)
-moderator_ring2.addSurface(-1, moderator_outer_radius)
-moderator_ring3.addSurface(+1, moderator_outer_radius)
+# Moderator
```
samuelshaner: Beautiful!

src/Geometry.cpp
```diff
@@ -750,14 +751,14 @@ std::string Geometry::getFSRKey(LocalCoords* coords) {
  */
 void Geometry::subdivideCells() {
 
-  std::map<int, Universe*> all_universes = _root_universe->getAllUniverses();
-  std::map<int, Universe*>::iterator iter;
+  /* Compute minimum radius needed to surround a lattice cell */
+  /* This is used as the maximum radius for all ringified Cells */
+  double width_x = getWidthX() / 2.;
+  double width_y = getWidthY() / 2.;
+  double max_radius = sqrt(width_x * width_x + width_y * width_y);
```
samuelshaner: divide the squared widths by 2

samuelshaner:
@wbinventor - this new feature makes setting the rings for the moderator so much easier! My one question is: Is the discretization scheme for the moderator the discretization scheme we desire? I believe the max radius we desire should be the max radius that *conserves* the volume of the lattice cell. This way, the rings in the moderator are discretized with the outer ring being of equal volume to the inner rings. Note that the last ring would have to be unbounded since the last ring would actually lie outside the max radius instead of inside it.

src/Geometry.cpp
wbinventor: Why? I divide the widths before squaring them here.

wbinventor:
Great point about the discretization scheme, I'm not really sure *what* we want. Although imperfect, this initial implementation seemed the easiest to 1) leverage the existing `Cell::ringify()` method and 2) ensure the radial discretization extends into the corners of a pin cell lattice. I'm not convinced that we need to conserve the area of the lattice cell, why do you believe that to be important? The one benefit that I can see to this is that the radial discretization would be "squeezed" towards the fuel pin with rings of larger delta radius in the corners of the pin cell (where the flux gradient is smaller).

src/Geometry.cpp
samuelshaner: Whoops, you're right!

wbinventor:
@samuelshaner I pushed a revision to the code which uses the "equivalent" cylinder for a lattice cell as the maximum radius used in the discretization. At the moment this has added ~20 lines of new and somewhat redundant looking code to `Cell::ringify()` but it seems to do achieve what you requested. I don't see any easy way to generalize the code and am not too motivated to try harder, so will leave it as is unless someone else sees a cleaner solution.

src/Cell.cpp
```diff
@@ -1041,10 +1036,47 @@ void Cell::ringify(std::vector<Cell*>* subcells) {
                "the 2 halfspaces for each Surface.", _id, halfspace1,
                zcylinder1->getId(), halfspace2, zcylinder2->getId());
 
+  /* Loop over ZCylinders and create a new Cell clone for each ring */
+  std::vector<ZCylinder*>::iterator iter2;
+  std::vector<Cell*>::iterator iter3;
+
+  /* If this is an unbounded Cell - i.e., moderator in a fuel pin cell - then
+   * make the first ring span the volume outside the outermost cylinder */
+  if (radius1 > radius2) {
```
samuelshaner: This must be true given the check on line 1032

src/Cell.cpp
```diff
@@ -1041,10 +1036,47 @@ void Cell::ringify(std::vector<Cell*>* subcells) {
...
+    ZCylinder* outer = new ZCylinder(x1, y1, radius1);
+
+    /* Create outermost ZCylinder for each of the sectorized Cells */
+    if (subcells->size() != 0) {
```
geogunow: Why is ``subcells`` a pointer to std::vector rather than just std::vector?

wbinventor: Because it is passed in as an argument to this method. If it weren't a pointer, we'd need copy constructors in all of our methods (which we once had) *and* it will slow this method down a lot.

src/Cell.cpp
geogunow: Agreed, also there is no matching else for this conditional. To hold the equal volume invariant you could make this conditional reflect whether the circle of radius equal to ``radius1`` exceeds the bounding box.

src/Cell.cpp
```diff
@@ -1056,24 +1088,19 @@ void Cell::ringify(std::vector<Cell*>* subcells) {
...
+        log_printf(DEBUG, "Creating a new ring in sector Cell ID=%d",
                    (*iter3)->getId());
 
         /* Create a new Cell clone */
         Cell* ring = (*iter3)->clone();
         ring->setNumSectors(0);
         ring->setNumRings(0);
-
-        /* Add new bounding ZCylinder surfaces to the clone */
         ring->addSurface(-1, (*iter2));
```
geogunow: If you implement the change above to hold the equal volume invariant you should make a conditional on adding this surface match the conditional on adding the outer cell above.

wbinventor: Can you propose some sample code to do what you think it should do?

src/Cell.cpp
geogunow: What about passing by reference?

src/Cell.cpp
wbinventor: Alright @samuelshaner @geogunow I tried to refactor this to achieve the equal volume invariance - let me know what you think. Note that this fix was quick - I simply merge what were previously the outermost two "rings" in the moderator. If you want something else, propose some code.

src/Cell.cpp
wbinventor: Alright, everything is passed by reference now

samuelshaner:
Just tested this out and it works great! Thanks for the addition @wbinventor. I'll let @geogunow do the honors once he has approved it.

PR #240, opened 2015-12-13
General Spatial Plotting

wbinventor:
This PR adds a new routine to `openmoc.plotter` for general plotting of FSR-based data on a 2D spatial mesh. This routine is a generalization of the routines used to plot materials, cells, fluxes, etc. The new `plot_spatial_data(...)` method is now used to greatly simplify many of the other routines in `openmoc.plotter`. I also started the PEP8-ifying process by making all tabs 4 spaces rather than just 2 in the `plotter` module.

@geogunow - you had mentioned to me before that you thought the `plotter` module had a lot of redundant code and could be simplified. This PR starts that process (to be honest, I just wanted the general routine for myself) but if you have more ideas let me know.


geogunow:
I like this a lot! The diff for this is quite large though - I can't view the changes to ``plotter.py`` on github.

wbinventor:
Yeah GitHub has something like a 1000 line limit for diffs in PRs. Since I tabbed every line over in "plotter.py" it went over the limit. We've run into this limitation several times recently with PRs to OpenMC. You can either 1) look at "plotter.py" in its entirety on my fork/branch and compile your comments in a message here or 2) make comments directly on the diffs for one or more of my commits for this PR [link](https://github.com/wbinventor/OpenMOC/commits/generalized-spatial-plotter).

wbinventor:
Alright, so I think this is ready for a second *thorough* review. I went overhauled the plotting methods even further with my latest commits to do the following:

* Added a new `PlotParams` class to containerize default and user-defined plotting parameters
* Made `plot_spatial_data(...)` accept either a NumPy `ndarray` or a Pandas `DataFrame` 

The second item is particularly cool since it allows users to quickly generate bunch of plots on the spatial mesh. This can be nice if one has a bunch of different data (e.g., fission rate evolution in time, fission rate errors for different MGXS libraries, etc.) This is **fast** since `plot_spatial_data(...)` will *only* build the 2D FSR surface map *once* and will then fill it with columns from the `DataFrame` for each plot. The following snippet of code shows how one can initialize and plot a Pandas `DataFrame`:

```python
import numpy as np
import pandas as pd

# Initialize a Pandas DataFrame with normally distributed random data
num_fsrs = geometry.getNumFSRs()
df = pd.DataFrame(np.random.randn(num_fsrs,3), columns=list('ABC'))

# Initialize a PlotParams object
plot_params = openmoc.plotter.PlotParams()
plot_params.geometry = geometry
plot_params.suptitle = 'Pandas DataFrame'
plot_params.filename = 'pandas-df'
plot_params.colorbar = True

# Enforce consistent color scheme across figures
plot_params.vmin = df.values.min()
plot_params.vmax = df.values.max()

openmoc.plotter.plot_spatial_data(df, plot_params)
```

This will generate the following three plots for the three columns in the `DataFrame`:

**Column 'A'**


**Column 'B'**


**Column 'C'**

wbinventor:
@geogunow @samuelshaner no rush on the review, but this PR is finished from my standpoint as it does everything I need it to do.

wbinventor:
Just to quantify how nice it is to support plotting of Pandas DataFrames, I've observed that it takes roughly than 1/70 the time to plot a DataFrame of 70-group SPH factors than it did the 70-group FSR scalar fluxes for a fuel assembly problem. We might consider moving the flux plotting to use DataFrames, though this would introduce a new dependency. Anyway, just a thought...

samuelshaner:
Why is `loglog` set to `False` on line 588 in plotter.py?

samuelshaner:
Can you add units to the Energy label on line 638 of plotter.py?

samuelshaner:
If we want to be strict to the PEP 8 style guide for the files, plotter.py exceeds the max line limit of 79 characters on lines 143, 717, 780, 783, 824, 889, 896, 1080, 1094, and 1123.

samuelshaner:
That's super impressive that Pandas speeds up plotting by that much!

wbinventor:
Well Pandas doesn't speed it up per se. The speedup is due to the fact that
we only have to query the geometry to build the 2D grid of FSRs *once* for
*all* columns in the DataFrame.

wbinventor:
Thanks for the review @samuelshaner!

wbinventor:
@samuelshaner here are my replies to your comments:

* `loglog` is set to `False` if a user does not define energy group bounds. In this case, the fluxes are plot according to their integer energy group indices instead of energy bounds. It wouldn't make sense to use a loglog plot here (though perhaps semilogy would be appropriate).
* Units on "Energy" label for plots of flux in energy - a user can define energy group bounds with arbitrary units. In order to put units in the label, we'd need a new input parameter for the units. If you think this is needed I can add it in.
* All lines are now 80 characters or less for PEP8 conformance.

samuelshaner:
Thanks @wbinventor! This all sounds great. I'll test it out later this morning and then hopefully merge in.

samuelshaner:
Can you remove the blank line 828 in plotter.py. It awkwardly sits in the middle of a statement.

wbinventor:
After our convo yesterday, let's hold off on this for me to revert back to the old algorithm for plotting cells and materials. I think we need to retain the ability to plot cells and materials without first generating FSRs from track segmentation. I also want to root out the segmentation fault that @samuelshaner identified - no segmentation fault should go unresolved!

wbinventor:
I've confirmed that `openmoc.plotter` still works for the eigenmode fluxes. I'll post a fix for the cells / materials plotting soon.

samuelshaner:
@wbinventor - what do you think about generalizing the map to be by `domains` instead of strictly by `FSRs`? If the number of FSRs is 0, then Cells can be used as the domains and the Cell and Material plotting should work before segments/FSRs are created. I've worked a bit on implementing this just to see if it is possible (and I think it will be) and can PR into your branch if we think this is a worthwhile approach. It adds a few lines to the `plot_cells`, `plot_materials`, and `plot_spatial_data` functions, but I don't think it overly bloats them.

wbinventor:
@samuelshaner - your idea sounds interesting, I'd be open to something
along those lines. I'm out for the rest of today, but could look at a PR
later if you have a starting point for the change.

wbinventor:
I've updated this PR to resolve the issues you raised @samuelshaner. The `plot_spatial_data(...)` routine now can take in a map of cell, material or FSR IDs mapped to the data to project on the 2D image. Overall, this reduced the amount of code in `plot_cells(...)` and `plot_materials(...)` even further, at the expense of ~10 extra lines of code in `plot_spatial_data(...)`.

This should be ready for another round of reviews.

geogunow:
@wbinventor I finished my review of this PR. Everything looks great, with a few minor aesthetic remarks. Additionally, the spacing (particularly the lack thereof) on lines 887, 890, 922, and 926 seems a bit weird to me but I'm not sure if it goes against the style guide in any way. Other than that, this seems ready to merge but I'll let @samuelshaner weigh in before merging.

wbinventor:
This PR has been updated per @geogunow's comments.

wbinventor:
@geogunow is this ready to merge?

samuelshaner:
This looks good to me. I'll let @geogunow give the final approval and merge it in.

wbinventor:
Just thought I'd ping you on this @geogunow. I'd like to move on this PR and my SPH PR which have been up here for months.

PR #242, opened 2016-01-17
Improved Memory De-Allocation for CMFD

wbinventor:
This PR aims to improve the memory allocation and deallocation in `Cmfd`. In particular, I ran into some issues when looping over many simulations which use the same `Geometry` and `Cmfd` objects (initialized outside of some loop) but with different MGXS libraries. In particular, the MGXS libraries had different energy group structures, which has to be properly communicated to `Cmfd` by the `Solver` rather than by the `Geometry`. For example, here is a simplified version of the simulation loop:

```python
# Create OpenMOC Geometry for 17 x 17 fuel assembly
...

# Initialize CMFD
cmfd = openmoc.Cmfd()
cmfd.setSORRelaxationFactor(1.5)
cmfd.setLatticeStructure(17,17)
cmfd.setKNearest(3)
openmoc_geometry.setCmfd(cmfd)

# Loop over OpenMC MGXS Libraries with different energy group structures
for mgxs_lib in mgxs_libs:
    openmoc_materials = \
        openmoc.materialize.load_openmc_mgxs_lib(mgxs_lib, openmoc_geometry)

    # Initialize a TrackGenerator
    track_generator = openmoc.TrackGenerator(openmoc_geometry, num_azim, spacing)
    track_generator.setNumThreads(num_threads)
    track_generator.generateTracks()

    # Initialize a Solver
    solver = openmoc.CPUSolver(track_generator)
    solver.setConvergenceThreshold(tolerance)

    # Run an eigenvalue calculation
    solver.computeEigenvalue()
    solver.printTimerReport()
```

Without this PR, one had to initialize `Cmfd` inside of the loop on each iteration. Depending on how one set up the loop, there were also memory leaks from `Cmfd` when initialized multiple times. I believe these leaks have been fixed. Although I would like to also move `TrackGenerator` initialization outside of the loop, it does not seem to be easily possible without wrestling with `Cmfd` further (since `Cmfd` gives NaN eigenvalues on all but the first iteratoin for some reason).

In addition, this PR fixes a memory leak in the `openmoc.materialize` module's methods to load MGXS libraries from HDF5 or OpenMC. The leak was only present if one loaded an MGXS library into the same `Geometry` many times, over-riding the materials in that `Geometry` each time.


wbinventor:
I should note that I recognize that the revisions to `Cmfd` here do not permit user-defined coarse group structures to be specified - I'm not sure how best to add this back in but am open to ideas. Perhaps some sort of new boolean `_user_groups` class attribute for the `Cmfd` class would be helpful.

wbinventor:
Hold off on reviewing this - I've been running into some issues "thinning" the parametric study in the code above and re-using tracks between simulations with different MGXS libraries.

openmoc/materialize.py
```diff
@@ -131,6 +131,9 @@ def load_from_hdf5(filename='mgxs.h5', directory='mgxs',
                 # the Material must be cloned for each unique Cell
                 if material != None:
                     if len(domains) > geometry.getNumMaterials():
+                        # Return memory ownership of material to Python
+                        # so it will be deallocated after material is cloned
+                        material.thisown = 1
```
samuelshaner: If I understand correctly, `thisown` can set to 0 or 1 signifying False and True. I feel like a boolean would make it a bit more obvious what is going on. What are your thoughts on setting `thisown` here (and perhaps other places) with booleans.

wbinventor:
WHEW! Alright, this PR definitely exploded well beyond what I had planned for it, but it introduces some much needed revisions and bug fixes. Both @samuelshaner and @geogunow should look this over, esp. since our goal is to sync up both closed and open versions of the code soon. The over-arching objective behind this PR is to make it possible to more cleanly re-use objects and inter-change them between simulations. In particular, I need to run many simulations with different MGXS libraries - with different `Material` objects and/or different group structures - which is not currently possible without running into (irreproducible) memory (de-)allocation errors. This PR resolves these issues.

My original plan for this PR was to meet the following two objectives:

* Run multiple simulations with different `Material` objects within a single set of `Geometry`, `TrackGenerator`, `Cmfd` and `Solver` objects
* Run multiple simulations with different energy group structures with a single set of `Geometry`, `TrackGenerator`, `Cmfd` and `Solver` objects

Previously, I gave some example code for the types of parametric studies I want to be able to run with the same set of objects (basically the same set of `Tracks` to reduce runtime overhead). The following is a updated version of pseudocode showing the sort of "low overhead" parametric study possible with this PR:

```python
# Create OpenMOC Geometry for 17 x 17 fuel assembly
...

# Initialize CMFD
cmfd = openmoc.Cmfd()
cmfd.setSORRelaxationFactor(1.5)
cmfd.setLatticeStructure(17,17)
cmfd.setKNearest(3)
openmoc_geometry.setCmfd(cmfd)

# Initialize a TrackGenerator
track_generator = openmoc.TrackGenerator(openmoc_geometry, num_azim, spacing)
track_generator.setNumThreads(num_threads)
track_generator.generateTracks()

# Initialize a Solver
solver = openmoc.CPUSolver(track_generator)
solver.setConvergenceThreshold(tolerance)

# Loop over OpenMC MGXS Libraries with different energy group structures
for mgxs_lib in mgxs_libs:
    openmoc_materials = \
        openmoc.materialize.load_openmc_mgxs_lib(mgxs_lib, openmoc_geometry)
    solver.computeEigenvalue()
```

The diff for this PR will appear as a variety of unrelated changes to the codebase, but they all fall under the same umbrella of a philosophy for memory (de-)allocation and simulation setup. In order to permit flexibility with parametric studies, it is my opinion that we should **defer as much memory allocation and simulation setup to simulation runtime.** My goal would be to permit users to modify the material / geometric model in between simulations while safely assuming that those changes will be picked up by the `Solver` when an eigenvalue or fixed source calculation is performed. Hopefully the diff will make more sense when viewed through a lens that recognizes this was my approach here. 

On a related note, the `Geometry::initializeFlatSourceRegions()` method was refactored - and should no longer be included in inputs since the new version is called directly by the `Solver` at simulation runtime. In addition, in the process of doing my own personal tests for this PR, I updated the `GPUSolver` and `VectorizedSolver` to produce the same results for both fixed source and eigenvalue calculations as the most commonly used `CPUSolver`. The main issue (but by no means the only one) was that the formulations used to compute different types of residuals in `CPUSolver` changed sometime in the last year and the corresponding code for the `VectorizedSolver` and `GPUSolver` was left unchanged.

In conclusion, if you see a change you don't understand, just ask and I'll try to remember which of the many bugs that particular change was intended to fix ;-)

samuelshaner:
Wow - this is fantastic! Have the ability to run parameter sweeps in the same input file will be awesome! Thanks for putting in the effort to get this working @wbinventor!

openmoc/materialize.py
```diff
@@ -138,7 +141,11 @@ def load_from_hdf5(filename='mgxs.h5', directory='mgxs',
                     if isinstance(domain_spec, int):
                         material = openmoc.Material(id=domain_spec)
                     else:
-                        material = openmoc.Material(name=domain_spec)
+                        # Reproducibly hash the domain name into an integer ID
+                        domain_id = \
+                            int(hashlib.md5(domain_spec).hexdigest()[:4], 16)
```
samuelshaner: I'm curious, what do the 4 and 16 represent here?

openmoc/materialize.py
```diff
@@ -211,6 +220,10 @@ def load_from_hdf5(filename='mgxs.h5', directory='mgxs',
             py_printf('INFO', 'Loaded "fission" MGXS for "%s %s"',
                       domain_type, str(domain_spec))
 
+    # Inform SWIG to garbage collect any old Materials from the Geometry
+    for material_id in old_materials:
+        old_materials[material_id].thisown = False
```
samuelshaner: Thanks for changing the `thisown` to be set to `True` and `False`!

src/CPUSolver.cpp
```diff
@@ -222,16 +191,37 @@ void CPUSolver::initializeSourceArrays() {
...
+    int fsr_id, group;
+    for (iter = _fixed_sources_map.begin();
+	 iter != _fixed_sources_map.end(); ++iter) {
+      fsr_group_key = iter->first;
+      fsr_id = fsr_group_key.first;
+      group = fsr_group_key.second;
+
+      if (group <= 0 || group > _num_groups)
+        log_printf(ERROR,"Unable to use fixed source for group %d in "
+                   "in a %d energy group problem", group, _num_groups);
```
samuelshaner: remove one of the `in`s here.

src/CPUSolver.cpp
```diff
@@ -550,11 +540,13 @@ double CPUSolver::computeResidual(residualType res_type) {
     FP_PRECISION inverse_k_eff = 1.0 / _k_eff;
     FP_PRECISION* nu_sigma_f;
     Material* material;
+    FP_PRECISION* sigma_s;
 
     for (int r=0; r < _num_FSRs; r++) {
```
samuelshaner: I remember Matt Everson running into some performance issues with the calculation for the scattering source taking a long time when many energy groups were used. While not directly related to this PR, seeing this for loop reminded me of that issue. Could you parallelize this loop to avoid any performance bottlenecks?

src/Cmfd.cpp
```diff
@@ -880,6 +886,10 @@ void Cmfd::initializeMaterials() {
 
   Material* material;
 
+  /* Delete old Cmfd surface currents vector if it exists */
+  if (_materials != NULL)
+    delete _materials;
```
samuelshaner: I think this should be `delete [] _materials`, otherwise I would assume just the pointer to the array is deleted and not the actual array.

src/Cmfd.cpp
```diff
@@ -948,11 +958,19 @@ void Cmfd::initializeCellMap() {
...
-    _group_indices_map = new int[_num_moc_groups];
+  /* Setup one-to-one fine-to-coarse group map if not specified by user */
+  if (!_user_group_indices) {
+    setGroupStructure(NULL, _num_moc_groups+1);
+    _user_group_indices = false;
   }
 
+  /* Delete old group indices map if it exists */
+  if (_group_indices_map != NULL)
+    delete _group_indices_map;
```
samuelshaner: ditto

src/VectorizedSolver.cpp
```diff
@@ -246,6 +240,32 @@ void VectorizedSolver::initializeSourceArrays() {
...
+  /* Compute the number of SIMD vector widths needed to fit energy groups */
+  _num_vector_lengths = (_num_groups / VEC_LENGTH) + 1;
+
+  /* Reset the number of energy groups by rounding up for the number
+   * of vector widths needed to accomodate the energy groups */
+  _num_groups = _num_vector_lengths * VEC_LENGTH;
+  _polar_times_groups = _num_groups * _num_polar;
+
+  /* Allocate array of mutex locks for each FSR */
+  _FSR_locks = new omp_lock_t[_num_FSRs];
```
samuelshaner: `_FSR_locks` is already allocated and initialized within `CPUSolver::initializeFSRs();` so this can be removed.

openmoc/materialize.py
wbinventor: This simply specifies the how many decimal places to include in the hashed ID. I'll be honest that the method *may* allow for hash collisions and redundant material IDs...

src/CPUSolver.cpp
wbinventor: got it

src/CPUSolver.cpp
wbinventor: Sure, I just parallelized all of the residual calculations by FSR. It's worth noting that the only *real* fix for lots of energy groups would be some sort of sparse data structure for scattering matrices, but that can be left for another day...

src/VectorizedSolver.cpp
wbinventor: Yes indeed!

src/Cmfd.cpp
wbinventor: good catch

src/Cmfd.cpp
wbinventor: agreed

src/CPUSolver.cpp
```diff
@@ -501,6 +491,7 @@ double CPUSolver::computeResidual(residualType res_type) {
 
     norm = _num_FSRs;
 
+    #pragma omp for schedule(guided)
```
samuelshaner: I think you mean `#pragma omp parallel for`

src/CPUSolver.cpp
```diff
@@ -522,6 +513,8 @@ double CPUSolver::computeResidual(residualType res_type) {
     FP_PRECISION* nu_sigma_f;
     Material* material;
 
+    #pragma omp for private(new_fission_source, old_fission_source, \
```
samuelshaner: ditto

src/CPUSolver.cpp
```diff
@@ -550,11 +543,15 @@ double CPUSolver::computeResidual(residualType res_type) {
     FP_PRECISION inverse_k_eff = 1.0 / _k_eff;
     FP_PRECISION* nu_sigma_f;
     Material* material;
+    FP_PRECISION* sigma_s;
 
+    #pragma omp for private(new_total_source, old_total_source, \
```
samuelshaner: ditto

src/accel/cuda/GPUSolver.cu
```diff
@@ -82,6 +82,101 @@ struct isnan_test {
 
 
 /**
+ * @brief A functor to multiply	all elements in	a Thrust vector by a constant.
```
samuelshaner: odd gap here

wbinventor: oh weird, doesn't even show up in my emacs

src/CPUSolver.cpp
wbinventor: indeed

wbinventor:
This is up-to-date.

samuelshaner:
Tested out the CPUSolver, GPUSolver, and VectorizedSolver and everything looks good. It will be nice to start running some parameter sweeps without having to use multiple input files!

PR #248, opened 2016-01-22
Travis Continuous Integration

wbinventor:
OpenMOC is about to grow up! This PR is the first to add in support for continuous integration [Travis CI](https://travis-ci.org/). It also adds in a basic regression test suite infrastructure which uses [CTest](https://cmake.org/Wiki/CMake/Testing_With_CTest). The test suite infrastructure includes an `InputSet` and `TestingHarness` classes which are loosely based on those in [OpenMC](https://github.com/mit-crpg/openmc).


tests/input_set.py
```diff
@@ -0,0 +1,69 @@
...
+                                               directory='../../sample-input/')
+
+    def create_geometry(self):
+        """Instantiate a pin cell Geometry."""
+
+        zcylinder = openmoc.ZCylinder(x=0.0, y=0.0, radius=1.0, name='pin')
+        left = openmoc.XPlane(x=-2.0, name='left')
+        right = openmoc.XPlane(x=2.0, name='right')
+        top = openmoc.YPlane(y=2.0, name='top')
+        bottom = openmoc.YPlane(y=-2.0, name='bottom')
```
samuelshaner: Can we change these surface names to `xmin`, `xmax`, etc? With the transition to 3D, it can be unclear whether top and bottom refer to y or z.

wbinventor: Sure

wbinventor:
It's worth noting that I haven't seen any indication that Travis supports GPU instances, so we may be limited to only testing the `CPUSolver`.

wbinventor:
Woohoo! I like this green I'm seeing :-)

tests/testing_harness.py
```diff
@@ -0,0 +1,176 @@
...
+        assert compare, 'Results do not agree.'
+
+    def _cleanup(self):
+        """Delete track, plot, etc. directories and test files."""
+
+        # Create list of directories and/or files to remove
+        outputs = [os.path.join(os.getcwd(), 'tracks')]
+        outputs.append(os.path.join(os.getcwd(), 'log'))
+        outputs.append(os.path.join(os.getcwd(), 'plots'))
+        outputs.append(os.path.join(os.getcwd(), 'simulation-state'))
```
samuelshaner: Should also include `fission-rates`

PR #252, opened 2016-01-23
Updated documentation for polar quadrature sets

samuelshaner:


docs/source/methods/track_generation.rst
```diff
@@ -9,7 +9,7 @@ The method of characteristics uses tracks across the geometry to discretize the
 
 .. _cyclic-tracks:
 
-Cylic Tracks
+Cyclic Tracks
 ============
```
wbinventor: Does the underlying "=" need to be extended to the full length of this subtitle?

docs/source/methods/track_generation.rst
```diff
@@ -80,15 +80,58 @@ The azimuthal angle quadrature set is computed based on the fraction of azimutha
...
 
    w_{m} = \frac{1}{2\pi}\left[2\pi - \phi_{m} + \frac{\phi_{m} - \phi_{m-1}}{2}\right] & \;\; \text{if} \;\;\; m = M
 
+
+.. _polar-angle-quadrature:
+
+Polar Angle Quadrature
+======================
+
+In OpenMOC, there are five polar quadrature sets that couple with the standard constant-angle azimuthal quadrature set. These include equal angles, equal weights, Gauss Legendre, Leonard,and Tabuchi-Yamamoto polar quadrature sets.  The formulas for computing the angles and weights of the equal angles and equal weights quadrature sets, given below, are described in the Handbook of Nuclear Engineering [Cacuci]_. The angles and weights for the Gauss Legendre, Leonard, and Tabuchi Yamamoto quadrature sets were taken from reference without modification [Yamamoto]_, [Cacuci]_.
```
wbinventor: space between "Leonard," and "and"

wbinventor:
This looks great! Ready to merge once a few typos are addressed.

wbinventor:
The only other thing I can think of for this PR would be to include the
code you used to generate the polar quad plots as downloadable attachments.
This could be like the files included in the developers guide for SWIG.
What do you think @samuelshaner?

wbinventor:
One more thing - should we provide a few code examples showing how one can interchange the different polar quadratures or define a custom one? Or do you want to wait on that if you'll be refactoring the quadratures anyway?

docs/source/methods/track_generation.rst
samuelshaner: good catch - yes it does.

samuelshaner:
I've written up the `plot_quadrature(...)` method that we have in ClosedMOC, so I'll just include that in this PR and then include the code for plotting the quadrature.

samuelshaner:
Including custom quadratures sounds like a good idea. I'll include that with this PR. It should just require minor modifications when I refactor the quadrature.

samuelshaner:
I have updated this PR by adding a `plot_quadrature(...)` function to the plotter and added sample code in the documentation on how to create PolarQuad objects.

wbinventor:
This looks good to go to me! One more thought which can be addressed here or in a future PR is that the `thisown` attribute for the `PolarQuad` should be appropriately toggled when an object is handed off to a `Solver`.

samuelshaner:
Isn't that already done? The `Solver::setPolarQuadrature(...)` method is in `thisown.i` and it looks like thisown is being toggled.

wbinventor:
Oops, I missed that when I looked earlier! Ok, this is good to go then.

PR #255, opened 2016-02-01
Auto ID Maximization

wbinventor:
This PR adds new routines to maximize auto-generated IDs for `Surfaces`, `Materials`, `Cells` and `Universes`. These new routines include the following:

* `maximize_surface_id()`
* `maximize_material_id()`
* `maximize_cell_id()`
* `maximize_universe_id()`

Our auto-generated ID streams start at 10000 by default. In the event that a user creates an OpenMOC `Geometry` from an OpenCG `Geometry` which has its own IDs, the auto-generated ID streams are not updated (or "maximized"). This PR changes that so that after an `opencg.Geometry` is transformed into an equivalent `openmoc.Geometry` the auto ID streams are updated to the max IDs of the `opencg.Surface`, `opencg.Material`, `opencg.Cell` and `opencg.Universe` objects. This prevents IDs from overlapping if a user decides to 1) create an OpenMOC `Geometry` from an OpenCG `Geometry` and 2) interchange some `Materials`, `Cells`, etc. with new OpenMOC ones.

### DIFF
--- a/src/Cell.cpp
+++ b/src/Cell.cpp
@@ -3,7 +3,7 @@
 
 int Cell::_n = 0;
 
-static int auto_id = 10000;
+static int auto_id = DEFAULT_INIT_ID;
 
 
 /**
@@ -26,7 +26,7 @@ int cell_id() {
  * @brief Resets the auto-generated unique Cell ID counter to 10000.
  */
 void reset_cell_id() {
-  auto_id = 10000;
+  auto_id = DEFAULT_INIT_ID;
 }
 
 
--- a/src/Cell.cpp
+++ b/src/Cell.cpp
@@ -31,6 +31,21 @@ void reset_cell_id() {
 
 
 /**
+ * @brief Maximize the auto-generated unique Cell ID counter.
+ * @details This method updates the auto-generated unique Cell ID
+ *          counter if the input parameter is greater than the present
+ *          value. This is useful for the OpenCG compatibility module
+ *          to ensure that the auto-generated Cell IDs do not
+ *          collide with those created in OpenCG.
+ * @param material_id the id assigned to the auto-generated counter
+ */
+void maximize_cell_id(int cell_id) {
+  if (cell_id > auto_id)
+    auto_id = cell_id;
+}
+
+
+/**
  * @brief Constructor sets the unique and user-specifed IDs for this Cell.
  * @param id the user-specified optional Cell ID
  * @param name the user-specified optional Cell name
--- a/src/Cell.cpp
+++ b/src/Cell.cpp
@@ -37,7 +37,7 @@ void reset_cell_id() {
  *          value. This is useful for the OpenCG compatibility module
  *          to ensure that the auto-generated Cell IDs do not
  *          collide with those created in OpenCG.
- * @param material_id the id assigned to the auto-generated counter
+ * @param cell_id the id assigned to the auto-generated counter
  */
 void maximize_cell_id(int cell_id) {
   if (cell_id > auto_id)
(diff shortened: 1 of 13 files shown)


src/Cell.cpp
```diff
@@ -31,6 +31,21 @@ void reset_cell_id() {
 
 
 /**
+ * @brief Maximize the auto-generated unique Cell ID counter.
+ * @details This method updates the auto-generated unique Cell ID
+ *          counter if the input parameter is greater than the present
+ *          value. This is useful for the OpenCG compatibility module
+ *          to ensure that the auto-generated Cell IDs do not
+ *          collide with those created in OpenCG.
+ * @param material_id the id assigned to the auto-generated counter
```
geogunow: This parameter does not match the input parameter

src/Surface.cpp
```diff
@@ -23,12 +23,27 @@ int surf_id() {
 /**
  * @brief Resets the auto-generated unique Surface ID counter to 10000.
  */
-void reset_surf_id() {
+void reset_surface_id() {
   auto_id = 10000;
```
geogunow: Could we either:
a) Have an argument detailing the reset value of ``auto_id``
OR
b) Make 10000 a constant such as ``DEFAULT_INIT_ID`` or some other name

geogunow:
I just finished my first review of this PR. I did not run any tests yet verifying it works (hopefully this will eventually be done by Travis). I actually didn't realize we assigned ID's starting at 10000. I see the reason for doing this but ....
1) It was confusing for me at first to see ``auto_id`` in functions where it was not declared. Could we do something to make it obvious this is a global constant (all caps / make it a class variable) ?
2) The 10000 number seems rather arbitrary. Could we define a constant for 10000 instead of just using the numeric value everywhere?

src/Cell.cpp
wbinventor: good catch

src/Surface.cpp
wbinventor: The reset value of 10000 actually is detailed in the doxygen comment for this method.

I do understand your proposal to define a macro constant for this value. My hesitation to do so is that 1) a `DEFAULT_CELL_INIT_ID` would only be used in one place, and 2) I'm not sure where to define a more global `DEFAULT_INIT_ID` for `Materials`, `Surfaces`, `Cells` and `Lattices`.

wbinventor: Sorry, I forgot we had "constants.h" :-) I'll put a `DEFAULT_INIT_D` there.

wbinventor:
Thanks for reviewing this @geogunow. Sorry to surprise you with the auto IDs! I added a `DEFAULT_INIT_ID` to "constants.h" as you suggested. I didn't (yet) capitalize `auto_id()` since it is not actually a global - each method is local to "Cell.cpp," "Material.cpp," etc. If you have some ideas for how to refactor this whole auto-ID scheme I'm open to it, but perhaps in a separate PR (plus, we should chat since I've spent quite a lot of time developing auto-ID schemes which sync between OpenMC-OpenCG-OpenMOC...it's not as trivial as it sounds).

Also, your comment made it sound like you weren't sure if this had passed Travis - it did indeed as noted by the green check marks above. Now that we have Travis setup, it will test *every* commit made to a pull request and either give a green check box (all test passed) or a red x (at least one test failed).

geogunow:
Right, but I believe Travis has very little code coverage at the moment and
we still rely on running sample input manually.

wbinventor:
Oh yes, I think that is a good idea. Though I think that most of the code
coverage encapsulate by sample inputs is accounted for save for 1)
GPUSolver (CUDA/GPUs not available on Travis), 2) Krylov solver (too slow),
3) SPH factors (requires installation of OpenMC and OpenCG), and 3)
alternative compilers such as icpc and clang. So if you run your own tests
you might focus on those aspects.

PR #257, opened 2016-02-03
Fixed bug in handling horizontal tracks

geogunow:
This PR fixes a bug in our tracking. Currently, ``ZCylinder::intersection`` does not work for tracks that travel horizontally. This is because the change in the ``y`` coordinate is used to determine whether a point is inside or outside of a ``ZCylinder``. This PR changes the routine to check the ``x`` coordinate for non-vertical tracks. In addition, a simple constructor for ``fsr_data`` was added to ensure that pointers are initialized as ``NULL`` (this is not guaranteed otherwise).

### DIFF
--- a/src/Surface.cpp
+++ b/src/Surface.cpp
@@ -842,7 +842,8 @@ int ZCylinder::intersection(Point* point, double angle, Point* points) {
     /* Solve for where the line x = x0 and the Surface F(x,y) intersect
      * Find the y where F(x0, y) = 0
      * Substitute x0 into F(x,y) and rearrange to put in
-     * the form of the quadratic formula: ay^2 + by + c = 0 */
+     * the form of the quadratic formula: ay^2 + by + c = 0 
+     * This is simplified for a z-cylinder with a vertical axis */
     a = 1.;
     b = _D;
     c = _A * x0 * x0 + _C * x0 + _E;
--- a/src/Surface.cpp
+++ b/src/Surface.cpp
@@ -843,11 +843,11 @@ int ZCylinder::intersection(Point* point, double angle, Point* points) {
      * Find the y where F(x0, y) = 0
      * Substitute x0 into F(x,y) and rearrange to put in
      * the form of the quadratic formula: ay^2 + by + c = 0 */
-    a = _B * _B;
+    a = 1.;
     b = _D;
     c = _A * x0 * x0 + _C * x0 + _E;
 
-    discr = b*b - 4*a*c;
+    discr = b*b - 4*c;
 
     /* There are no intersections */
     if (discr < 0)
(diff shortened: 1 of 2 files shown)


src/Surface.cpp
```diff
@@ -843,11 +843,11 @@ int ZCylinder::intersection(Point* point, double angle, Point* points) {
      * Find the y where F(x0, y) = 0
      * Substitute x0 into F(x,y) and rearrange to put in
      * the form of the quadratic formula: ay^2 + by + c = 0 */
-    a = _B * _B;
+    a = 1.;
```
wbinventor: Can you add a comment above that the equations below are simplified for the case of a cylinder aligned with the z-axis? this will be helpful if anyone ever looks at this and tries to map these equations back to the generic equations for a quadratic surface.

wbinventor:
Looks good @geogunow, I'm curious if this will have any impact on ray tracing time. Thanks for the bug fix, I'll merge it in once Travis finishes.

geogunow:
@wbinventor it seems Travis has stalled. Do you have any idea why?

wbinventor:
I'm unsure, it is very peculiar. I think the best course of action is to fix the issue such that the test suite successfully runs on your machine and then push so that Travis will start a fresh run. If that doesn't solve things then I can start looking around in search of the problem.

wbinventor:
Oh wait it looks like tlit passed with your last commit!

src/Surface.cpp
```diff
@@ -901,11 +902,12 @@ int ZCylinder::intersection(Point* point, double angle, Point* points) {
      * rearrange to put in the form of the quadratic formula:
      * ax^2 + bx + c = 0
      */
-    double m = sin(angle) / cos(angle);
+    bool right = angle < M_PI / 2. || angle > 3. * M_PI / 2.;
```
wbinventor: Can you give a comment describing the meaning of this new boolean variable?

src/Surface.cpp
```diff
@@ -901,11 +902,12 @@ int ZCylinder::intersection(Point* point, double angle, Point* points) {
...
+    double m = tan(angle);
     q = y0 - m * x0;
-    a = _A + _B * _B * m * m;
-    b = 2 * _B * m * q + _C + _D * m;
-    c = _B * q * q + _D * q + _E;
+    a = 1. + m * m;
+    b = 2 * m * q + _C + _D * m;
```
wbinventor: looks weird to multiply by `1.` and `2` in successive lines - either remove the decimal from the first line or add it to the second line (understandably it doesn't matter in C/C++, but for the sake of consistency).

src/Surface.cpp
```diff
@@ -919,9 +921,9 @@ int ZCylinder::intersection(Point* point, double angle, Point* points) {
       ycurr = y0 + m * (points[num].getX() - x0);
       zcurr = z0;
       points[num].setCoords(xcurr, ycurr, zcurr);
-      if (angle < M_PI && ycurr > y0)
+      if (right && xcurr > x0)
```
wbinventor: Perhaps we need some comments explaining each of these conditionals since it took you some time to dust them off and re-learn them to debug this problem in the first place.

wbinventor:
I like this PR once a few more explanatory comments are added.

PR #260, opened 2016-02-05
Default Matplotlib Settings in Plotter Module

wbinventor:
This short PR updates the `openmoc.plotter` module to always use the default matplotlib settings. This is important because `ipython` and IPython Notebook each override the default matplotlib settings with their own settings to produce plots which render well within their interactive environments. The result is that any figures created from `openmoc.plotter` when called from `ipython` will appear much smaller (~1/4x) than when called from a normal Python script.

This PR now applies the default matplotlib settings before generating a plot. In addition, it restores whatever settings were in use prior to the call to `openmoc.plotter` before returning from each plotting routine. This ensures that the settings from `ipython` will continue to be used once control is returned back to the user.


geogunow:
Good idea, but if we're customizing the settings how about we chose some
non-default settings that work best for us. I often need to increase the
text size on figures when including them in papers/presentations/anything.

wbinventor:
Good point @geogunow. I just pushed new functionality to return the `matplotlib.pyplot.figure` objects created by each routine through optional `get_figure` parameters. While I think we should support "helper parameters" to set things up for our figures, we can't support *everything* that matplotlib allows. Perhaps a good compromise for now is to return the figure for the user to customize as they like. Also, the IPython Notebook example has been updated to showcase the figure return feature.

geogunow:
I'm confused. Why can't we support matplotlib parameters? Maybe we're not understanding each other. For example, whenever I make a plot for a paper or presentation from Python I always copy over:
```
font = {'family' : 'sans-serif',
        'weight' : 'normal',
        'size' : 15}
plt.rc('font', **font)
```
This increases the font size on all plots. Having plots that are readable by default would be very useful for users that want to present the results in papers (ie. all of us). But I still do like the idea of allowing the user to change the plot parameters as well.

wbinventor:
I think we are in agreement about wanting / no needing / to change plot
parameters for effective visuals. I'm just saying that matplotlib allows
you to make an awful lot of adjustments to your plots. We can't possibly
"wrap" all of these various options up or else it will bloat the size of
our function signatures. But of course we can consider adding some new
options to the `PlotParams` class. But for now most of these tweaks can be
made by the user directly to the `figure` that can now be returned from
each plotting routine, without requiring the routines to customize plots
themselves.

Also, I don't want to go implementing a bunch of options for matplotlib
figures in this PR. I address a very specific set of problems with a series
of new features here. If one of us wants to submit a PR with support for
fontsizes or anything else then that come in a subsequent PR.

geogunow:
I'm just saying we add a few lines like the ones I posted at the top of ``plotter.py``. We choose some parameters we like and it should carry over through ``curr_rc``. I'm basically saying the default ``rc`` in matplotlib is not very good because the font is very small for most purposes. This should be at most 10 lines of code extra.

wbinventor:
I added the font parameters that you suggested as a global variable to the top of `openmoc.plotter`, and set the fontsize after setting matplotlib's default fonts.

openmoc/plotter.py
```diff
@@ -37,6 +37,11 @@
 # in your ~/.matplotlib/matplotlibrc
 plt.ioff()
 
+## Font parameters to use in all plots
```
geogunow: Remove the double #

openmoc/plotter.py
```diff
@@ -54,12 +59,17 @@
...
+# @param get_figure whether or not to return the Matplotlib figure
+def plot_tracks(track_generator, get_figure=False):
 
+    global subdirectory, font
     directory = openmoc.get_output_directory() + subdirectory
 
+    # Ensure that normal settings are used even if called from ipython
+    curr_rc = dict(matplotlib.rcParams)
+    matplotlib.rcParams.update(matplotlib.rcParamsDefault)
+    plt.rc('font', **font)
```
geogunow: I think we only need this once. Not every time we plot. When you set the font, it overwrites the ``rcParams`` in the library I believe.

openmoc/plotter.py
wbinventor: That is there since that is doxygen syntax for global python variables. I'd prefer to keep consistent with doxygen syntax for our comments until (unless) we make a systematic effort to transition towards only docstrings only in our Python code.

geogunow: Ok

openmoc/plotter.py
wbinventor: We need to call it each time now. The code now sets the default matplotlib settings (and your fonts) at the *beginning* of each method call, and resets the settings to whatever they were prior at the *end* of each function call.

geogunow: Ok. How about we save the ``rcParams`` we want to a class or global variable? That way we would just need to set the parameters using that variable.

wbinventor: You mean store a deepcopy of the rcParams from matplotlib as a global
variable? What would this accomplish? All I can see it doing is saving us
2-3 limes of code since we could inject our font styles into it in one
place. Perhaps that's worth it, just wondered what value you see in doing
this.

wbinventor:
@geogunow I've added a new `matplotlib_rcparams` global variable to the top of `openmoc.plotter`. It is customized with the font specs you suggested, and is available to the user to further customize if needed.

geogunow:
Everything looks good to me. Unless @samuelshaner has any additional comments, I think this is ready to merge.

Issue #258, opened 2016-02-03
Error in tracking for tracks nearly tangent to ZCylinders


geogunow:
Developing my tests for tracking, I have come across some bugs in our current tracking. Specifically, if a create a track that *barely* intersects a circle, it calculates a segment length of infinity. Whereas the segment lengths should be roughly:

[1.82467365265, 0.00752119300887, 1.82467365265]

OpenMOC calculates 
[1.82467365265, 0.00752119300887, inf]

The code to produce this result is:

```
import openmoc
import math

###############################################################################
#                          Main Simulation Parameters
###############################################################################

options = openmoc.options.Options()

num_threads = options.getNumThreads()
track_spacing = options.getTrackSpacing()
num_azim = options.getNumAzimAngles()
tolerance = options.getTolerance()
max_iters = options.getMaxIterations()

openmoc.log.set_log_level('NORMAL')


###############################################################################
#                            Creating Materials
###############################################################################

openmoc.log.py_printf('NORMAL', 'Importing materials data from HDF5...')

materials = openmoc.materialize.load_from_hdf5('c5g7-mgxs.h5', '../')

###############################################################################
#                            Creating Surfaces
###############################################################################

openmoc.log.py_printf('NORMAL', 'Creating surfaces...')

zcylinder = openmoc.ZCylinder(x=0.0, y=0.0, radius=1.0, name='pin')
left = openmoc.XPlane(x=-2.0, name='left')
right = openmoc.XPlane(x=2.0, name='right')
top = openmoc.YPlane(y=2.0, name='top')
bottom = openmoc.YPlane(y=-2.0, name='bottom')

left.setBoundaryType(openmoc.REFLECTIVE)
right.setBoundaryType(openmoc.REFLECTIVE)
top.setBoundaryType(openmoc.REFLECTIVE)
bottom.setBoundaryType(openmoc.REFLECTIVE)

###############################################################################
#                             Creating Cells
###############################################################################

openmoc.log.py_printf('NORMAL', 'Creating cells...')

fuel = openmoc.Cell(name='fuel')
fuel.setFill(materials['UO2'])
fuel.addSurface(halfspace=-1, surface=zcylinder)

moderator = openmoc.Cell(name='moderator')
moderator.setFill(materials['Water'])
moderator.addSurface(halfspace=+1, surface=zcylinder)
moderator.addSurface(halfspace=+1, surface=left)
moderator.addSurface(halfspace=-1, surface=right)
[...]

PR #261, opened 2016-02-06
Adding tracking tests to test suite

geogunow:
This PR adds tracking-specific cases to the test suite that test edge and corner cases. These tests create specific tracks indicative of both common tracks and corner cases and check their segmentation. Six cases are added for three setups. For each setup there is a test with and a test without an overlaid CMFD mesh. The three setups are:

* A simple 3x3 gird created with a lattice
* A simple 3x3 grid created with ``XPlanes`` and ``YPlanes`` (identical to the lattice case)
* A  simple pin cell

 NOTE: The nearly tangent track case right now calculates an incorrect value of infinity for a segment in the pin cell geometry. This is due to issue #258 which has yet to be resolved. Once resolved the reference test data should be updated to contain correct tracking results.


tests/input_set.py
```diff
@@ -135,6 +135,108 @@ def create_geometry(self):
...
+    def create_materials(self):
+        """Instantiate C5G7 Materials."""
+        self.materials = \
+            openmoc.materialize.load_from_hdf5(filename='c5g7-mgxs.h5',
+                                               directory='../../sample-input/')
+
+    def create_geometry(self):
+        """Instantiate a 3x3 grid Geometry by making a lattice"""
+
+        left = openmoc.XPlane(x=-3.0, name='left')
```
wbinventor: Use `xmin`, `xmax`, etc. in place of `left`, `right`, etc.

tests/input_set.py
```diff
@@ -135,6 +135,108 @@ def create_geometry(self):
...
+    def create_materials(self):
+        """Instantiate C5G7 Materials."""
+        self.materials = \
+            openmoc.materialize.load_from_hdf5(filename='c5g7-mgxs.h5',
+                                               directory='../../sample-input/')
+
+    def create_geometry(self):
+        """Instantiate a 3x3 grid Geometry."""
+
+        xplanes = [None] * 4
```
wbinventor: Add a comment above to describe what this code block does (i.e., sets up planes to form a grid)

tests/input_set.py
```diff
@@ -135,6 +135,108 @@ def create_geometry(self):
...
+            val = -3.0 + 6.0/3 * i
+            xplanes[i] = openmoc.XPlane(x=val, name='xplane' + str(i))
+            yplanes[i] = openmoc.YPlane(y=val, name='yplane' + str(i))
+
+        xplanes[0].setBoundaryType(openmoc.REFLECTIVE)
+        xplanes[-1].setBoundaryType(openmoc.REFLECTIVE)
+        yplanes[0].setBoundaryType(openmoc.REFLECTIVE)
+        yplanes[-1].setBoundaryType(openmoc.REFLECTIVE)
+
+        cells = [[None]*3 for i in range(3)]
```
wbinventor: same for here - i think a few more comments throughout would be helpful

tests/test_tracking_grid/results_true.dat
```diff
@@ -0,0 +1,27 @@
+Diagonal track...
+Number of segments = 3
```
wbinventor: Do we need the text metadata like "Segment", "length," "CMFD FWD=", etc. or can we do away with it and simply print numbers? For example, in place of:

```
Segment 0: length=2.82842707634, FSR ID=0, CMFD FWD=-1, CMFD BWD=-1, Material Name=Water, Material ID=10083
```

can we instead have:

```
2.82842707634, 0, 1, -1, Water, 10083
```

Of course this is a tradeoff between verbosity and bloating the size of our test suite and repo.

tests/test_tracking_grid/results_true.dat
```diff
@@ -0,0 +1,27 @@
...
+Segment 1: length=2.82842707634, FSR ID=1, CMFD FWD=-1, CMFD BWD=-1, Material Name=UO2, Material ID=49930
+Segment 2: length=2.82842707634, FSR ID=2, CMFD FWD=-1, CMFD BWD=-1, Material Name=Water, Material ID=10083
+Nudged Diagonal track...
+Number of segments = 5
+Segment 0: length=2.82841300964, FSR ID=0, CMFD FWD=-1, CMFD BWD=-1, Material Name=Water, Material ID=10083
+Segment 1: length=1.4142136024e-05, FSR ID=3, CMFD FWD=-1, CMFD BWD=-1, Material Name=Water, Material ID=10083
+Segment 2: length=2.82841300964, FSR ID=1, CMFD FWD=-1, CMFD BWD=-1, Material Name=UO2, Material ID=49930
+Segment 3: length=1.4142136024e-05, FSR ID=4, CMFD FWD=-1, CMFD BWD=-1, Material Name=Water, Material ID=10083
+Segment 4: length=2.82841300964, FSR ID=2, CMFD FWD=-1, CMFD BWD=-1, Material Name=Water, Material ID=10083
+Horizontal track...
```
wbinventor: The ellipsis "..." isn't needed

wbinventor: The ellipsis "..." isn't needed

wbinventor: Sorry to be a stickler, but I know from the openmc experience, the test suite will quickly grow to dominate the source tree in this repo, without careful management of our test results database.

tests/test_tracking_grid/test_tracking_grid.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python
+
+import os
+import sys
+sys.path.insert(0, os.pardir)
+sys.path.insert(0, os.path.join(os.pardir, 'openmoc'))
+from testing_harness import TestHarness
+from input_set import GridInput
+
+import openmoc
```
wbinventor: Per PEP8, put the `math` import at the top along with the other standard Python modules `os` and `sys`, and before third party modules like `openmoc`

tests/test_tracking_grid/test_tracking_grid.py
```diff
@@ -0,0 +1,85 @@
...
+import math
+
+class TrackingGridTestHarness(TestHarness):
+    """Tests cell radial discretization."""
+
+    def __init__(self):
+        super(TrackingGridTestHarness, self).__init__()
+        self.input_set = GridInput()
+        self._result = ''
+
```
wbinventor: Per PEP8, only one blank line between class methods.

tests/test_tracking_grid/test_tracking_grid.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def _setup(self):
+        """Initialize the materials and geometry in the InputSet."""
+        super(TrackingGridTestHarness, self)._create_geometry()
+
+
+    def _segment_track(self, track, geometry):
+
```
wbinventor: Add a one line docstring to this method signature

tests/test_tracking_grid/test_tracking_grid.py
```diff
@@ -0,0 +1,85 @@
...
+        # segmentize a track in a geometry, recording the segments in a string
+        geometry.segmentize(track)
+        num_segments = track.getNumSegments()
+        info = 'Number of segments = ' + str(num_segments) + '\n'
+        for i in range(num_segments):
+            info += 'Segment ' + str(i) + ': '
+            segment = track.getSegment(i)
+            info += 'length=' + str(round(segment._length, 8)) + ', '
```
wbinventor: Related to my comment above, I would opt to remove a lot of this extra textual metadata

tests/test_tracking_grid/test_tracking_grid.py
```diff
@@ -0,0 +1,85 @@
...
+            info += 'FSR ID=' + str(segment._region_id) + ', '
+            info += 'CMFD FWD=' + str(segment._cmfd_surface_fwd) + ', '
+            info += 'CMFD BWD=' + str(segment._cmfd_surface_bwd) + ', '
+            info += 'Material Name=' + str(segment._material.getName()) + ', '
+            info += 'Material ID=' + str(segment._material.getId()) + '\n'
+        track.clearSegments()
+        return info
+
+    def _run_openmoc(self):
+
```
wbinventor: Add a one line docstring to this method signature - specifically describe how this is different from the parent's `_run_openmoc()`

tests/test_tracking_grid/test_tracking_grid.py
```diff
@@ -0,0 +1,85 @@
...
+        # set track trajectories and locations
+        diag_track.setValues(-3, -3, 0, 3, 3, 0, math.atan(1))
+        nudge = 1e-5
+        nudge_diag_track.setValues(-3+nudge, -3, 0, 3, 3-nudge, 0, math.atan(1))
+        hor_track.setValues(-3, 0, 0, 3, 0, 0, 0)
+        ver_track.setValues(0, -3, 0, 0, 3, 0, math.pi/2)
+        rev_diag_track.setValues(3, 3, 0, -3, -3, 0, math.pi + math.atan(1))
+
+        # segmentize over the geometry
+        self._result += 'Diagonal track...\n'
```
wbinventor: Minimize textual metadata in the result string

tests/test_tracking_grid_cmfd/results_true.dat
```diff
@@ -0,0 +1,344 @@
+Pin cell with an overlaid 3 x 3 CMFD mesh
+Diagonal track...
+Number of segments = 3
+Segment 0: length=2.82842708, FSR ID=0, CMFD FWD=6, CMFD BWD=4, Material Name=Water, Material ID=10083
```
wbinventor: Ditto, lots of unnecessary metadata here

tests/test_tracking_grid/test_tracking_grid.py
```diff
@@ -0,0 +1,85 @@
...
+    def __init__(self):
+        super(TrackingGridTestHarness, self).__init__()
+        self.input_set = GridInput()
+        self._result = ''
+
+
+    def _setup(self):
+        """Initialize the materials and geometry in the InputSet."""
+        super(TrackingGridTestHarness, self)._create_geometry()
+
```
wbinventor: ditto

tests/test_tracking_grid/test_tracking_grid.py
```diff
@@ -0,0 +1,85 @@
...
+        # initialize track objects
+        diag_track = openmoc.Track()
+        nudge_diag_track = openmoc.Track()
+        hor_track = openmoc.Track()
+        ver_track = openmoc.Track()
+        rev_diag_track = openmoc.Track()
+        geometry = self.input_set.geometry
+
+        # set track trajectories and locations
```
wbinventor: For consistency with the rest of the code in the suite, capitalize the first word in comments

tests/test_tracking_grid_cmfd/test_tracking_grid_cmfd.py
```diff
@@ -0,0 +1,95 @@
...
+
+import os
+import sys
+sys.path.insert(0, os.pardir)
+sys.path.insert(0, os.path.join(os.pardir, 'openmoc'))
+from testing_harness import TestHarness
+from input_set import GridInput
+
+import openmoc
+import math
```
wbinventor: ditto, put `math` along with `os` and `sys

tests/test_tracking_grid_cmfd/test_tracking_grid_cmfd.py
```diff
@@ -0,0 +1,95 @@
...
+        self._result = ''
+
+
+    def _setup(self):
+        """Initialize the materials and geometry in the InputSet."""
+        super(TrackingGridCMFDTestHarness, self)._create_geometry()
+
+
+    def _segment_track(self, track, geometry):
+
```
wbinventor: add one line docstring to this method signature.

tests/test_tracking_grid_cmfd/test_tracking_grid_cmfd.py
```diff
@@ -0,0 +1,95 @@
...
+            info += 'FSR ID=' + str(segment._region_id) + ', '
+            info += 'CMFD FWD=' + str(segment._cmfd_surface_fwd) + ', '
+            info += 'CMFD BWD=' + str(segment._cmfd_surface_bwd) + ', '
+            info += 'Material Name=' + str(segment._material.getName()) + ', '
+            info += 'Material ID=' + str(segment._material.getId()) + '\n'
+        track.clearSegments()
+        return info
+
+    def _run_openmoc(self):
+
```
wbinventor: ditto, one line docstring

tests/test_tracking_grid_cmfd/test_tracking_grid_cmfd.py
```diff
@@ -0,0 +1,95 @@
...
+            self._result += self._segment_track(ver_track, geometry)
+            self._result += 'Reverse Diagonal track...\n'
+            self._result += self._segment_track(rev_diag_track, geometry)
+
+
+    def _get_results(self, num_iters=False, keff=False, fluxes=False,
+                     num_fsrs=True, num_segments=True, num_tracks=True,
+                     hash_output=False):
+        """Return the result string"""
+
```
wbinventor: Remove this blank line (per PEP8 don't leave a blank line between docstring and  single line method bodies)

tests/test_tracking_lattice/test_tracking_lattice.py
```diff
@@ -0,0 +1,85 @@
...
+
+import os
+import sys
+sys.path.insert(0, os.pardir)
+sys.path.insert(0, os.path.join(os.pardir, 'openmoc'))
+from testing_harness import TestHarness
+from input_set import LatticeGridInput
+
+import openmoc
+import math
```
wbinventor: ditto, move `math` up

tests/test_tracking_lattice/test_tracking_lattice.py
```diff
@@ -0,0 +1,85 @@
...
+
+class TrackingLatticeGridTestHarness(TestHarness):
+    """Tests cell radial discretization."""
+
+    def __init__(self):
+        super(TrackingLatticeGridTestHarness, self).__init__()
+        self.input_set = LatticeGridInput()
+        self._result = ''
+
```
wbinventor: ditto, remove extra blank lines

tests/test_tracking_lattice_cmfd/test_tracking_lattice_cmfd.py
```diff
@@ -0,0 +1,95 @@
...
+sys.path.insert(0, os.pardir)
+sys.path.insert(0, os.path.join(os.pardir, 'openmoc'))
+from testing_harness import TestHarness
+from input_set import LatticeGridInput
+
+import openmoc
+import math
+
+class TrackingLatticeGridCMFDTestHarness(TestHarness):
+    """Tests cell radial discretization."""
```
wbinventor: I don't think this class docstring applies, looks like a copy-paste issue

tests/test_tracking_lattice/test_tracking_lattice.py
```diff
@@ -0,0 +1,85 @@
...
+sys.path.insert(0, os.pardir)
+sys.path.insert(0, os.path.join(os.pardir, 'openmoc'))
+from testing_harness import TestHarness
+from input_set import LatticeGridInput
+
+import openmoc
+import math
+
+class TrackingLatticeGridTestHarness(TestHarness):
+    """Tests cell radial discretization."""
```
wbinventor: ditto, copy-pasted class dostring

tests/test_tracking_pin_cell_cmfd/test_tracking_pin_cell_cmfd.py
```diff
@@ -0,0 +1,100 @@
...
+sys.path.insert(0, os.pardir)
+sys.path.insert(0, os.path.join(os.pardir, 'openmoc'))
+from testing_harness import TestHarness
+from input_set import PinCellInput
+
+import openmoc
+import math
+
+class TrackingPinCellCMFDTestHarness(TestHarness):
+    """Tests cell radial discretization."""
```
wbinventor: ditto, class docstring

wbinventor:
Your new tests all look structurally sound, just look over my comments pertaining to PEP8-related issues. According to Travis, all of your test are currently failing. Do they pass on your machine?

geogunow:
@wbinventor They did pass on my computer individually but when I ran the entire test suite, they failed, likely because OpenMOC recompiled. The results showed that the computed segment lengths for some cases were ever-so-slightly different. I don't know what caused this but the difference was very small (on the order of 1e-7). It is possibly from the change in ``Circle::intersect`` that I submitted recently. Anyway, since I don't want trivial changes to cause the tests to fail, I increased the tolerance on segment length from rounding to 1e-8 to 1e-7 and I updated the ``results_true.dat``. Now all the tests pass.

tests/test_tracking_grid/results_true.dat
geogunow: I mean it depends. I can definitely output that but then a developer looking through the data might not understand the output. We can probably reduce the text in half by doing what you suggest but the output would be far harder to comprehend. We could reduce the output far more by hashing the result, making it even more hard to comprehend and understand though.

wbinventor: "It is certainly a tradeoff between transparency and data footprint. I would just argue that in order to debug one of these tracking tests, a developer would have to understand how the test works by looking at "test_tracking_*.py" anyway, which would lessen the need to have all of this redundant metadata.

wbinventor:
If the reason that the tests failed is that the "results_true.dat" were created using the version of OpenMOC installed on your machine rather than the "true" version, then the fact that the tests failed due to precision is exactly as we would want. I think the segment lengths, etc. should be returned to their original precision so that we catch discrepancies like this in the future.

To be clear, if you use the following to update the "results_true.dat" it will use the *installed* version of OpenMOC on your system which may or may not be consistent with the version on the current branch:

```python
cd tests/test_track*
python test_track*.py --update
```

Instead, it is better to update the "results_true.dat" with the version on the current branch (*i.e.*, the "true" version) with the following command:

```python
python update_results.py
```

This first installs the current version of OpenMOC in a local temporary folder to be imported for use in the tests, rather than using the globally available installed version of OpenMOC. In addition, it will build OpenMOC with the same compile time options as those used when "run_tests.py" is called, either locally or on Travis.

My recommendation is to 1) restore the original precision to the output strings and 2) call `python update_results.py` to create the correct "results_true.dat" files.

geogunow:
I have removed most of the meta-data from the tests. I think this should be ready for review/merge.

tests/input_set.py
```diff
@@ -135,6 +135,116 @@ def create_geometry(self):
         super(PinCellInput, self).create_geometry()
 
 
+class GridInput(InputSet):
+    """Sets up a simple grid problem using explicit plane"""
```
wbinventor: typo: "plane" -> "planes"

tests/input_set.py
```diff
@@ -135,6 +135,116 @@ def create_geometry(self):
...
+        """Instantiate C5G7 Materials."""
+        self.materials = \
+            openmoc.materialize.load_from_hdf5(filename='c5g7-mgxs.h5',
+                                               directory='../../sample-input/')
+
+    def create_geometry(self):
+        """Instantiate a 3x3 grid Geometry."""
+
+        # Create the planes that bound the cell and geometry
+        xplanes = [None] * 4
```
wbinventor: Is it really better to pre-allocate these lists? The code would be simpler by just appending to the empty lists on each iteration.

tests/input_set.py
```diff
@@ -135,6 +135,116 @@ def create_geometry(self):
...
+        # Add the cells to the universe
+        root_universe = openmoc.Universe(name='root universe')
+        for i in range(3):
+            for j in range(3):
+                root_universe.addCell(cells[i][j])
+
+        self.geometry = openmoc.Geometry()
+        self.geometry.setRootUniverse(root_universe)
+
+        super(GridInput, self).create_geometry()
```
wbinventor: What is the value of calling the parent class's `create_geometry()` method here?

tests/test_tracking_grid/test_tracking_grid.py
```diff
@@ -0,0 +1,85 @@
...
+        for i in range(num_segments):
+            info += str(i) + ': '
+            segment = track.getSegment(i)
+            info += str(round(segment._length, 8)) + ', '
+            info += str(segment._region_id) + ', '
+            info += str(segment._cmfd_surface_fwd) + ', '
+            info += str(segment._cmfd_surface_bwd) + ', '
+            info += str(segment._material.getName()) + ', '
+            info += str(segment._material.getId()) + '\n'
+        track.clearSegments()
```
wbinventor: Is it necessary to call `clearSegments()`? Won't this be called by the `Track` destructor anyway?

tests/test_tracking_grid/test_tracking_grid.py
```diff
@@ -0,0 +1,85 @@
...
+        num_segments = track.getNumSegments()
+        info = ' ' + str(num_segments) + '\n'
+        for i in range(num_segments):
+            info += str(i) + ': '
+            segment = track.getSegment(i)
+            info += str(round(segment._length, 8)) + ', '
+            info += str(segment._region_id) + ', '
+            info += str(segment._cmfd_surface_fwd) + ', '
+            info += str(segment._cmfd_surface_bwd) + ', '
+            info += str(segment._material.getName()) + ', '
```
wbinventor: Are both material name and ID necessary? Perhaps just one of them is sufficient?

wbinventor:
This PR is looking really good @geogunow. It is actually quite cool to see the ray tracing controlled from within Python - I've never tried this before myself and didn't realize it was "just" so simple!

One thought I have about this PR is that there is a lot of redundant code in your five `TestTracking*TestHarness` classes. What do you think about creating a new generic `TrackingTestHarness` class in "testing_harness.py" that contains all of the shared code? In particular, the `_segment_track(...)` method would only need to be implemented once in this parent class. The subclasses could each implement their own `run_openmoc(...)` method which would *only* initialize all `Tracks` of interest within a `dict`. The collection of `Tracks` would be passed to the parent class' `run_openmoc(...)` which would iterate over each and append to the `_results` string (using the `dict` keys as the headers for each `Track`). Finally, the `_get_results(...)` would only need to be implemented once in the parent class.

I personally like the of migrating simulation work flows which are common across many tests into generic `*TestingHarness` classes. This makes it easy to add new tests which inherit the majority of this functionality but test different features, parameters, initializations, corner cases, etc. with as little repeated code as possible.

What do you think?

tests/input_set.py
geogunow: Ellaborate on how that would work. My inclination is to tell you no because doing something like ```xplanes = [openMOC.Xplane(...)] * 4``` would just create 4 pointers to the same data.

tests/input_set.py
geogunow: I have no idea. You did it for the other inputs though. So I did the same.

tests/test_tracking_grid/test_tracking_grid.py
geogunow: It actually is necessary for cases in which I re-use the ``Track`` on multiple cases. If I don't clear the segments then re-using the ``Track`` would just append the segments to the current ``Track``.

tests/test_tracking_grid/test_tracking_grid.py
geogunow: Material ID is sufficient for checking accuracy, material name helps in debugging. The name is not sufficient because multiple materials could potentially have the same name.

tests/test_tracking_grid/test_tracking_grid.py
wbinventor: Ok I see.

tests/input_set.py
wbinventor: That doesn't mean it is applicable here.

geogunow:
@wbinventor I really like the idea of removing repeated code. I'll take a crack at this later today.

tests/input_set.py
wbinventor: If you like the pre-allocation then keep it. Alternatively, consider something more functional:

```python
xplanes = map(lambda x: openmoc.XPlane(x=x), [-3, -1, +1, +3])
yplanes = map(lambda x: openmoc.YPlane(y=y), [-3, -1, +1, +3])
```

tests/test_tracking_grid/test_tracking_grid.py
wbinventor: Yeah I agree that the ID is really all that is necessary here.

geogunow: Right, but if the test fails and you need to debug, the material name is
extremely useful.

wbinventor: Alright I'm fine with leaving it in

geogunow:
@wbinventor I believe this PR is ready for further review. However, Travis failed on my tests here on Github but passes locally on my machine. I have no idea why. I have no local changes not tracked and all commits have been pushed. It should work.

wbinventor:
Hmm, have you looked at the log files on Travis? There should be one each
for Python 2 and 3 which you can inspect online - I can't remember where,
but if you click around you'll find it.

geogunow:
@wbinventor It's failing for python3 but the output is not very useful here. Is there a way to run with python3 locally?

wbinventor:
I bet I know why. The hash function changed between Python 2 and 3 so the
iterator over dictionaries orders keys differently. Try storing your tracks
in an OrderedDict rather than a dict.

geogunow:
It seems that OrderedDict is not a built in Python data structure.

wbinventor:
Just import it from the "collections" module. This is something we've faced
before in OpenMC many times.
It seems that OrderedDict is not a built in Python data structure.

On Mon, Feb 8, 2016 at 8:17 PM Will Boyd <notifications@github.com> wrote:

> I bet I know why. The hash function changed between Python 2 and 3 so the
> iterator over dictionaries orders keys differently. Try storing your
tracks
> in an OrderedDict rather than a dict.
> On Feb 8, 2016 8:15 PM, "Geoffrey Gunow" <notifications@github.com> wrote:
>
> > @wbinventor <https://github.com/wbinventor> It's failing for python3 but
> > the output is not very useful here. Is there a way to run with python3
> > locally?
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/mit-crpg/OpenMOC/pull/261#issuecomment-181655606>.
> >
>
> —
> <https://github.com/mit-crpg/OpenMOC/pull/261#issuecomment-181655869>.
>

—
Reply to this email directly or view it on GitHub
<https://github.com/mit-crpg/OpenMOC/pull/261#issuecomment-181656248>.

tests/test_tracking_lattice/test_tracking_lattice.py
```diff
@@ -0,0 +1,48 @@
...
+import math
+from collections import OrderedDict
+sys.path.insert(0, os.pardir)
+sys.path.insert(0, os.path.join(os.pardir, 'openmoc'))
+from testing_harness import TrackingTestHarness
+from input_set import LatticeGridInput
+
+import openmoc
+
+class TrackingLatticeGridTestHarness(TrackingTestHarness):
```
wbinventor: What does `LatticeGrid` mean beyond just `Lattice`?

tests/test_tracking_lattice_cmfd/test_tracking_lattice_cmfd.py
```diff
@@ -0,0 +1,65 @@
...
+import math
+from collections import OrderedDict
+sys.path.insert(0, os.pardir)
+sys.path.insert(0, os.path.join(os.pardir, 'openmoc'))
+from testing_harness import TrackingTestHarness
+from input_set import LatticeGridInput
+
+import openmoc
+
+class TrackingLatticeGridCMFDTestHarness(TrackingTestHarness):
```
wbinventor: ditto, what does `LatticeGrid` mean beyond just `Lattice`? Unless "Grid" actually means something, I think it should be removed to eliminate any source of confusion.

tests/test_tracking_pin_cell/test_tracking_pin_cell.py
```diff
@@ -0,0 +1,52 @@
...
+    def __init__(self):
+        super(TrackingPinCellTestHarness, self).__init__()
+        self.input_set = PinCellInput()
+
+    def _setup(self):
+        """Initialize the materials, geometry, and tracks."""
+        super(TrackingPinCellTestHarness, self)._create_geometry()
+
+        # Initialize track objects
+        tracks = OrderedDict()
```
wbinventor: I would put this initialization in the parent class' `__init__()` so that `OrderedDict` only has to be imported into one file. Then each child class will simply populate `self.tracks` rather than `tracks`.

wbinventor:
I'm very happy with the progress on this PR! It's ready to merge pending my few most recent relatively minor comments.

tests/test_tracking_lattice/test_tracking_lattice.py
geogunow: We already have ``SimpleLattice`` so ``LatticeGrid`` implies that it is a ``Lattice`` but it implicitly forms a grid geometry (the same geometry as ``Grid``)

tests/test_tracking_lattice_cmfd/test_tracking_lattice_cmfd.py
geogunow: See above

geogunow:
I think we have a problem with Travis. It runs fine on my machine but not on Github.

wbinventor:
After working with it for some time now, I think that's doubtful but
possible. There is almost always a reason for it to fail.

wbinventor:
You might be right - it's not clear what exactly happened from the Travis
logs, but it definitely didn't fail on your tests. It looks like the
semi-official monkey patch that I use to install gcc 4.8 on Travis' AMI
with Ubuntu 12.04 - which still uses gcc 4.4 without fill C++11 support -
failed this time around. Try pushing the changes for my last comment and
see if that affects anything, perhaps the AWS instance just couldn't
download it by chance...

geogunow:
@wbinventor It passes Travis now

wbinventor:
Looks good - great job with this PR @geogunow!

PR #262, opened 2016-02-12
Implemented Sam's suggested fix for tracking

geogunow:
This implement's Sam's suggested change to ``Cell::containsPoint`` to resolve issue #258. This is only a one line change in the source code. Almost the entire diff comes from updating the test results to reflect the new tracking.


src/Cell.cpp
```diff
@@ -884,7 +884,7 @@ bool Cell::containsPoint(Point* point) {
 
     /* Return false if the Point is not in the correct Surface halfspace */
     if (iter->second->_surface->evaluate(point) * iter->second->_halfspace
-        < -ON_SURFACE_THRESH)
+        < 0.0)
```
wbinventor: Can this conditional be expressed in one line in under 80 characters?

geogunow: Not trivially

wbinventor:
This looks good to go to me once the test failures are corrected. This case doesn't look like a Travis failure, though one can never say never. Any idea what is wrong?

samuelshaner:
Thanks for making this PR @geogunow. Looks good to me.

wbinventor:
Kudos to this entry to a whole new era of track generation!

PR #263, opened 2016-02-12
Dots-Per-Inch (DPI) in Plotter Module

wbinventor:
This PR adds a new `dpi` parameter to all of the routines in the `openmoc.plotter` module. The default matplotlib dpi is 100 and this is now over-ridden to 500 as an OpenMOC default, but this can be specified by the user as well. In addition, the OpenMC MGXS `Libraries` for the SPH factor sample inputs (slab and pin cell) have been updated to conform to the latest version of OpenMC.


wbinventor:
My latest commit to this PR goes one step further and adds in support for the Python Imaging Library (PIL) in addition to matplotlib. We can now save our figures as pure bitmaps using PIL which look *much* better than the images from matplotlib. This is particularly true for high-res figures in which we may set the `gridsize` parameter to 1000 or more. In these cases, matplotlib's `imshow` method will provide diminishing returns since the resolution is only improved without first setting the figure size. Although this is easy to do, I think using PIL directly is a better option since it automatically scales with `gridsize` (the number of pixels) and does not require a new input parameter for `figsize`.

One can choose between `'matplotlib'` and `'pil'` using the new optional `library` parameter to most of the routines in `openmoc.plotter`.  Of course matplotlib has a lot of features that PIL does not, including colorbars, scales for axes, labels, etc., so `'matplotlib'` is kept as the default library.

wbinventor:
Just as a reference point, using PIL allows for crisp plots of complicated geometries such as the following spectrally clustered PWR geometries. These plots look like pixelated crap when generated with `matplotlib.plt.imshow`.

**2x2 BEAVRS Assemblies w/ Grid Spacers**


**2x2 BEAVRS Assemblies w/ Grid Spacer and Reflector**


Plots such as these are easily generated and displayed inline in Jupyter Notebooks as follows:

```python
openmoc.plotter.plot_materials(geometry, gridsize=1000, library='pil')
IPython.display.Image(filename='plots/materials.png', width=400, height=400)
```

samuelshaner:
Wow - those plots are beautiful!

openmoc/plotter.py
```diff
@@ -414,7 +430,7 @@ def plot_flat_source_regions(geometry, gridsize=250, xlim=None, ylim=None,
     fig = figures[0]
 
     # Plot centroids on top of 2D flat source region color map
-    if centroids:
+    if library == 'matplotlib' and centroids:
```
samuelshaner: Can the centroids not be plotted with `pil`?

wbinventor:
Now you can :-) I'm going to add some tests to the test suite to cover a lot of our plotting functionality. Once that is in, this PR will address everything I set out to do. But the new plotting features - DPI and PIL - can be reviewed independently of the new tests.

wbinventor:
Alright, I think this PR is ready for a full review. My last series of commits do the following:

* Added 9 new tests to suite for near full coverage of `openmoc.plotter`
* Fixed `update_results.py` to first install OpenMOC locally just as is done in `run_tests.py`
* Fixed test suite issues with Python 3
* Removed `dpi` param from `openmoc.plotter` routines since PIL is more effective way to produce high-res images. Plus, one can still adjust the DPI in `matplotlib_rcparams`.

Note that the new plotting tests only work with Matplotlib >=v1.5. Older versions generate plots with slightly different pixel counts, which fail the tests when those images are hashed and compared to the true hashed results generated with >=v1.5.

openmoc/plotter.py
```diff
@@ -1028,48 +1074,66 @@ def plot_spatial_data(domains_to_data, plot_params, get_figure=False):
...
+        # If input was Pandas DataFrame, append column name to filename
+        if pandas_df:
+            plot_filename += '-{0}'.format(domains_to_data.columns[i])
 
-        # If the user requested the Matplotlib figure handles for further
-        # specialization, append the handle to this figure to a list
-        if get_figure:
-            figures.append(fig)
+        # Append file extension (e.g., '.png', '.ppm') to filename
+        plot_filename +=  plot_params.extension
```
samuelshaner: extra spacing after `+=`

openmoc/plotter.py
```diff
@@ -540,10 +581,11 @@ def plot_spatial_fluxes(solver, energy_groups=[1], norm=False,
     plot_params.geometry = geometry
     plot_params.zcoord = zcoord
     plot_params.gridsize = gridsize
+    plot_params.library = library
     plot_params.xlim = xlim
     plot_params.ylim = ylim
     plot_params.colorbar = True
-    plot_params.cmap = None
+    plot_params.cmap = plt.get_cmap('jet')
```
samuelshaner: Is there any reason the default should not just be set to `jet`?

tests/testing_harness.py
```diff
@@ -258,3 +263,57 @@ def _get_results(self, num_iters=False, keff=False, fluxes=False,
...
         return self._result
+
+
+class PlottingTestHarness(TestHarness):
+    """Specialized TestHarness for testing plotting."""
+
+    def __init__(self):
+        super(PlottingTestHarness, self).__init__()
+        self.figures = []
+
```
samuelshaner: I think PEP 8 suggests only one blank line between class methods.

tests/testing_harness.py
```diff
@@ -258,3 +263,57 @@ def _get_results(self, num_iters=False, keff=False, fluxes=False,
...
+            else:
+                fig.save(plot_filename)
+            
+            # Open the image file in PIL
+            img = Image.open(plot_filename)
+
+            # Hash the image and append to output string
+#            sha512 = hashlib.sha512()
+#            sha512.update(outstr.encode('utf-8'))
+#            outstr = sha512.hexdigest()
```
samuelshaner: I'm assuming these commented out lines should be removed?

samuelshaner:
This PR looks great! I really like the new crisper plots. I'm happy with it once my comments are addressed.

openmoc/plotter.py
wbinventor: We use matplotlib's default colormap for all plots except the flux and fission rates (or if one specifies one to `PlotParams` for `plot_spatial_data(...)`). Because of this I decided to keep the default as `None` and only specify `jet` for the two plot types where we want it.

tests/testing_harness.py
wbinventor: Yes

wbinventor:
Just addressed all of your comments @samuelshaner.

geogunow:
Looks good to me!

PR #264, opened 2016-02-14
Parallelized segmentation loops in the TrackGenerator

samuelshaner:
This pull requests speeds up some of the set up methods by parallelizing all loops over segments in the TrackGenerator. In doing so, the ownership of the FSR locks has been given to the TrackGenerator and the Solver gets the FSR locks from the TrackGenerator to avoid unnecessary duplication.

In running a very fine track laydown on the LRA problem, I found that the `TrackGenerator::initializeSegments()` method was very slow because it called `Geometry::findFSRMaterial()` for every segment. This method is quite slow because it requires a search through the whole CSG hierarchy. The `TrackGenerator::initializeSegments()` method has been restructured so that the materials for each FSR are precomputed by looping over all the FSRs.

The LRA input files have also been updated by separating the creation of the surfaces, cells, universes, and lattices into different files instead of forming them all in the `geometry.py` file.


sample-input/benchmarks/LRA/LRA.py
```diff
@@ -1,10 +1,13 @@
 import openmoc
+import openmoc.plotter as plotter
+from openmoc.options import Options
+from lattices import lattices, universes, cells
```
wbinventor: it doesn't look like you use `lattices` or `cells` in this script, can they be removed from this import?

sample-input/benchmarks/LRA/LRA-cmfd.py
```diff
@@ -1,10 +1,13 @@
 import openmoc
+import openmoc.plotter as plotter
+from openmoc.options import Options
+from lattices import lattices, universes, cells
```
wbinventor: And here?

sample-input/benchmarks/LRA/LRA.py
samuelshaner: Good catch, yes they can. I just kind of accumulated the list of items imported as I worked up from surfaces to cells to universes to lattices to the main input file.

sample-input/benchmarks/LRA/LRA-cmfd.py
samuelshaner: yes

src/Solver.cpp
```diff
@@ -624,11 +625,14 @@ void Solver::countFissionableFSRs() {
   log_printf(INFO, "Counting fissionable FSRs...");
 
   /* Count the number of fissionable FSRs */
-  _num_fissionable_FSRs = 0;
+  int num_fissionable_FSRs;
+#pragma omp parallel for reduction(+:num_fissionable_FSRs)
```
wbinventor: Wow, so is this "smart" enough to only reduce when the conditional evaluates as `true`? I never realized this was possible. The skeptic in me just wants a confirmation that the value is the same with or without OpenMP - though I assume this is reflected in the test suite results when run with or without threading.

src/TrackGenerator.cpp
```diff
@@ -261,21 +282,34 @@ FP_PRECISION* TrackGenerator::getFSRVolumes() {
...
+    FP_PRECISION volume;
 
-      for (int s=0; s < _tracks[i][j].getNumSegments(); s++) {
-        curr_segment = _tracks[i][j].getSegment(s);
-        volume = curr_segment->_length * _azim_weights[azim_index];
-        FSR_volumes[curr_segment->_region_id] += volume;
+    for (int i=0; i < _num_azim; i++) {
+#pragma omp for
+      for (int j=0; j < _num_tracks[i]; j++) {
+
```
wbinventor: I'd opt to remove this extra line

src/Solver.cpp
samuelshaner: Each thread gets is own copy of the local variable `num_fissionable_FSRs` (notice I had to create a local variable; OpenMP cannot do a reduction over a private variable) and each thread accumulates its own counter of fissionable FSRs. After leaving the for loop, the accumulators for each thread are added together.

src/TrackGenerator.cpp
```diff
@@ -261,21 +282,34 @@ FP_PRECISION* TrackGenerator::getFSRVolumes() {
...
+    int azim_index, fsr_id;
+    segment* curr_segment;
+    FP_PRECISION volume;
 
-      for (int s=0; s < _tracks[i][j].getNumSegments(); s++) {
-        curr_segment = _tracks[i][j].getSegment(s);
-        volume = curr_segment->_length * _azim_weights[azim_index];
-        FSR_volumes[curr_segment->_region_id] += volume;
+    for (int i=0; i < _num_azim; i++) {
+#pragma omp for
```
wbinventor: All of the OpenMP pragmas show up without any indentation. Is this intentional? They are indented inline within rest of the codebase. I've seen this style elsewhere and honestly do not strongly prefer one style over another, but would like to keep things consistent.

src/TrackGenerator.cpp
```diff
@@ -644,6 +705,24 @@ void TrackGenerator::generateTracks(bool neighbor_cells) {
...
+ * @brief Create an array of OpenMP mutual exclusion locks for each FSR.
+ * @details This method allocates and initializes an array of OpenMP
+ *          mutual exclusion locks for each FSR for use in loops that
+ *          set or update values by FSR.
+ */
+void TrackGenerator::initializeFSRLocks() {
+
+  /* Allocate array of mutex locks for each FSR */
+  int num_FSRs = _geometry->getNumFSRs();
+  _FSR_locks = new omp_lock_t[num_FSRs];
```
wbinventor: Any old locks should be deleted here if necessary

src/TrackGenerator.cpp
samuelshaner: According to the google c++ style guide, preprocessor directives should not be indented. I sneaked in a PR a week ago removing the indents on the pragmas so have kept that consistency here. I also don't have a strong preference either way, but lean towards removing them because the google c++ style guide emacs plugin likes to remove the indents when it auto-indents. That said, they do stand out quite a bit.

src/TrackGenerator.cpp
samuelshaner: ok

src/TrackGenerator.cpp
```diff
@@ -1702,51 +1781,67 @@ void TrackGenerator::correctFSRVolume(int fsr_id, FP_PRECISION fsr_volume) {
   log_printf(INFO, "Correcting FSR %d volume from %f to %f",
              fsr_id, curr_volume, fsr_volume);
 
-  int num_segments, azim_index;
-  double dx_eff, d_eff;
-  double volume, corr_factor;
-  segment* curr_segment;
-  segment* segments;
+#pragma omp parallel
```
wbinventor: Are you using FSR volume correction? Have you tested out this new parallel implementation? This code isn't used in any of our tests or sample inputs, so I'm wary to accept these changes without either a) confirmation that it works as expected from a test problem and/or b) a new test in the test suite to demonstrate that it works.

src/TrackGenerator.cpp
samuelshaner: Good point, I'll add that.

src/TrackGenerator.cpp
samuelshaner: I never use and have not tested it out in parallel. I'll add a test for it.

src/TrackGenerator.cpp
```diff
@@ -1825,66 +1929,71 @@ void TrackGenerator::splitSegments(FP_PRECISION max_optical_length) {
     log_printf(ERROR, "Unable to split segments since "
 	       "tracks have not yet been generated");
 
-  int num_cuts, min_num_cuts;
-  segment* curr_segment;
+#pragma omp parallel
```
wbinventor: Ditto with this new code - I'm not sure that any of our tests actually require segment splitting. I think we need a test problem that meets the segment splitting criterion in order to verify this.

samuelshaner: will do.

wbinventor: And this segment splitting should be both with and without CMFD to ensure the CMFD `segment` attributes are properly specified.

src/TrackGenerator.cpp
```diff
@@ -1918,14 +2024,29 @@ void TrackGenerator::initializeSegments() {
...
   /* Iterate over all Track segments and assign them each a Material */
-  for (int i=0; i < _num_azim; i++) {
-    for (int j=0; j < _num_tracks[i]; j++) {
-      for (int s=0; s < _tracks[i][j].getNumSegments(); s++) {
-        curr_segment = _tracks[i][j].getSegment(s);
-	region_id = curr_segment->_region_id;
-        Material* mat = _geometry->findFSRMaterial(region_id);
-	curr_segment->_material = _geometry->findFSRMaterial(region_id);
-	FSR_keys_map->at(FSRs_to_keys->at(region_id))->_mat_id = mat->getId();
+#pragma omp parallel
```
wbinventor: I like the performance improvements to this routine a lot. As above, I'm wary of this simply because I added this functionality in to support parametric studies of multiple simulations with different MGXS libraries. We probably need a few "multi-sim" tests w/ and w/o CMFD with different MGXS to demonstrate that initialization and memory deallocation still work properly with parallelization for these scenarios as they do now. I don't think the onus needs to be on you to put in these tests for this PR, but perhaps we should hold off on merging this until I can patch together 3-5 "multi-sim" tests in a separate PR.

wbinventor:
Very nice PR, I'm a big fan since I've noticed similar issues with very slow simulation initialization for highly discretized problems. I'm looking forward to trying this out with the hope that it might improve things a bit! I do think that a few new tests should be added to the suite to complement this PR to address segment splitting, FSR volume correction, and multi-simulation. I can work on a PR for the latter piece since it is particularly important to me and somewhat separate from the impetus for this PR.

wbinventor:
Also, when you debug the test suite failures, do note my latest comments about and changes to the test suite in #263. It might even be worthwhile to merge that in to here since it may help eliminate issues with updating test results, if needed.

src/TrackGenerator.cpp
samuelshaner: Letting this PR simmer until the multi-sim tests are merged is fine with me.

wbinventor: I'll submit a PR with multi-sim tests this evening since it's very important to me that we maintain that capability. Hopefully we can merge that and then this in within a few days. Do you plan to add tests for fsr volume correction and segment splitting? We should also have tests for track loading from a file (w/ and w/o a CMFD mesh) to ensure that tracks are always loaded when possible. Thought those tests aren't necessary here.

samuelshaner: I'll add tests for fsr volume correction and segment splitting here. Totally agree - track loading should be addressed in some tests, especially as we consider the considerable changes to the track objects for 3D MOC that we'll want to merge in. Best to have the tests spot potential sync issues before they get integrated.

Thanks for getting on the multi-sim tests so quickly!

tests/test_correct_fsr_volume/test_correct_fsr_volume.py
```diff
@@ -0,0 +1,52 @@
...
+        self.track_generator.correctFSRVolume(1, 2.7)
+        one_thread_volume = self.track_generator.getFSRVolume(1)
+
+        # Save results for one thread
+        self._result += str(round(old_volume,6)) + '\n'
+        self._result += str(round(one_thread_volume, 1)) + '\n'
+
+        # get results with multiple threads
+        self.track_generator.setNumThreads(2)
+        self.track_generator.correctFSRVolume(1, 2.5)
```
wbinventor: Is there a reason to use a different volume for 2 threads than for 1? If not, then I suggest that we simplify this method to the following:

```python
    # Always use 1 thread for FSR reproducibility
    self.track_generator.setNumThreads(1)
    self.track_generator.generateTracks()
 
    # Correct FSR volumes with the appropriate number of threads
    self.track_generator.setNumThreads(self.num_threads)
    old_volume = self.track_generator.getFSRVolume(1)
    self.track_generator.correctFSRVolume(1, 2.7)
    new_volume = self.track_generator.getFSRVolume(1)

    # Save results for one thread
    self._result += '{0: 1.6f}'.format(old_volume)
    self._result += '{0: 1.1f}'.format(new_volume)
```

wbinventor:
I like the new tests a lot! They do raise a question for me though - as you probably noticed, all of the tests set the `TrackGenerator's` thread count to 1 to ensure reproducible FSR assignments. In order to cover your new parallel code, perhaps we should add a new line to the `TestingHarness._generate_tracks()` method to set the number of threads to `TestingHarness.num_threads` *after* track generation?

friedmud:
I just ran some timing tests on this branch and the results are straight up shocking!  This is for a VERY fine pin-cell with 1.1 Million segments on my laptop, running OpenMC with 4 threads.  I just timed the entire run (because it was easy :-)...

Develop branch:
```
real	19m52.010s
user	22m10.573s
sys	0m0.436s
```

THIS branch:
```
real	1m45.949s
user	4m10.169s
sys	0m1.291s
```

This branch is essentially 10x faster!

samuelshaner:
Glad to hear! It is probably due almost entirely to a rearrangement of the `TrackGenerator::initializeSegments()` method and the speed up should increase as the ratio of segments to FSRs goes up. The develop branch called the `Geometry::findFSRMaterial()` method on *every segment*, but (as the method name implies) the method only needs to be called on *every FSR*. This PR made the switch and thus greatly sped up the process of initializing segments.

Also, all loops that required iterations over segments are now parallelized. I was running cases with 40 threads so literally one serial loop over segments during setup would be equivalent to the entire amount of time spent solving for the eigenvalue.

tests/test_correct_fsr_volume/test_correct_fsr_volume.py
samuelshaner: I used different volumes just to ensure there isn't a cases where multithreading computes the old volume. I like your solution better, so I'm make the change.

samuelshaner:
@wbinventor - I like the idea of resetting the number of threads for the `TrackGenerator` after track generation! I'm going to make that change in the test suite.

wbinventor:
Wow, I'm glad to see that the performance really benefits from this! I agree with @samuelshaner that this isn't really due to the parallelism introduced here as much as it is to his fixing a performance issue that I introduced in a recent PR. Sounds like the fix works quite well!

wbinventor:
I'll wait to thoroughly review until after #266 is merged since the diff for this is quite large

samuelshaner:
Alright, this PR is now ready for a full review.

sample-input/benchmarks/LRA/LRA-cmfd.py
```diff
@@ -1,10 +1,13 @@
 import openmoc
+import openmoc.plotter as plotter
+from openmoc.options import Options
+from lattices import lattices, universes
```
wbinventor: Is `lattices` needed here?

samuelshaner: nope

sample-input/benchmarks/LRA/LRA-mgxs-hdf5.py
```diff
@@ -25,11 +25,11 @@
...
 # Create a subgroup for region 1 materials data
-region_1 = material_group.create_group('region_1')
+region_1 = material_group.create_group('Region 1')
 
-sigma_t = numpy.array([0.2656, 1.5798])
-sigma_s = numpy.array([0.2318925, 0.02533, 0.00, 1.4794789])
-sigma_f = numpy.array([0.002, 0.05])
+sigma_t = numpy.array([0.265604, 1.579779])
+sigma_s = numpy.array([0.232022, 0.02533, 0.00, 1.479479])
+sigma_f = numpy.array([0.004602, 0.1091]) / 2.43
```
wbinventor: I presume the 2.43 is nu, correct? Perhaps explicitly define it as a `nu` variable for transparency.

sample-input/benchmarks/LRA/LRA.py
```diff
@@ -1,10 +1,13 @@
 import openmoc
+import openmoc.plotter as plotter
+from openmoc.options import Options
+from lattices import lattices, universes
```
wbinventor: ditto remove `lattices`

sample-input/benchmarks/LRA/lattices.py
```diff
@@ -0,0 +1,62 @@
+import openmoc
+from universes import universes, cells
+
+###############################################################################
+###########################   Creating Lattices   #############################
+###############################################################################
+
+refines = 1
```
wbinventor: Define this variable and what it does.

sample-input/benchmarks/LRA/surfaces.py
```diff
@@ -0,0 +1,18 @@
+import openmoc
+
+###############################################################################
+#                 Create dictionary of common surfaces
+###############################################################################
+
+surfaces = {}
+
+# Instantiate surfaces
+surfaces['Root x-min'] = openmoc.XPlane(x=-82.5, name='Root x-min')
```
wbinventor: If the only surfaces correspond to the "Root" then I would prefer removing the redundant `"Root"` from all of these keys and `Surface` names.

sample-input/benchmarks/LRA/LRA-mgxs-hdf5.py
samuelshaner: Correct, I'll make a `nu` variable.

sample-input/benchmarks/LRA/LRA.py
samuelshaner: will do

sample-input/benchmarks/LRA/lattices.py
samuelshaner: ok

sample-input/benchmarks/LRA/surfaces.py
samuelshaner: Ok

src/Cell.cpp
```diff
@@ -794,81 +940,6 @@ void Cell::addNeighborCell(Cell* cell) {
...
-      _min_y_bc = surface->getBoundaryType();
-    }
-    if (min_z != -std::numeric_limits<double>::infinity() && min_z > _min_z) {
-      _min_z = min_z;
-    }
-  }
-
-  /* If we could not find a bounds for any dimension, readjust
-   * it to +/- infinity */
-  if (_max_x == -std::numeric_limits<double>::infinity())
```
wbinventor: Is this readjustment accounted for in your new implementation?

src/ParallelHashMap.h
```diff
@@ -244,7 +244,7 @@ void FixedHashMap<K,V>::insert(K key, V value)
   /* check to see if key already exists in map */
   if (contains(key))
     return;
-
+
```
wbinventor: Hmm, I'm not sure where this whitespace is coming from in this file. Are these changes from @geogunow's PR #270?

src/ParallelHashMap.h
```diff
@@ -255,9 +255,9 @@ void FixedHashMap<K,V>::insert(K key, V value)
 
   /* place element in linked list */
   *iter_node = new_node;
-
+  
   /* increment counter */
-#pragma omp atomic
+  #pragma omp atomic
```
wbinventor: I thought we were no longer tabbing OpenMP pragmas?

src/ParallelHashMap.h
```diff
@@ -829,12 +829,12 @@ void ParallelHashMap<K,V>::print_buckets()
...
 
   /* get pointer to table, announce it will be searched */
-  FixedHashMap<K,V> *table_ptr;
+  FixedHashMap<K,V> *table_ptr; 
   do{
     table_ptr = _table;
     _announce[tid].value = table_ptr;
   } while (table_ptr != _table);
-
+
```
wbinventor: yeah, the diff for this file *looks* like junk but perhaps I'm wrong

src/Solver.cpp
```diff
@@ -624,11 +625,14 @@ void Solver::countFissionableFSRs() {
...
-  _num_fissionable_FSRs = 0;
+  int num_fissionable_FSRs = 0;
+#pragma omp parallel for reduction(+:num_fissionable_FSRs)
   for (int r=0; r < _num_FSRs; r++) {
     if (_FSR_materials[r]->isFissionable())
-      _num_fissionable_FSRs++;
+      num_fissionable_FSRs++;
   }
+
+  _num_fissionable_FSRs = num_fissionable_FSRs;
```
wbinventor: Do we have a test for the fissionable FSR counter?

src/Cell.cpp
samuelshaner: It looks to me like these readjustments are unnecessary. `_max_x` is initialized to positive infinity and I don't see how a surface could have a maximum x value of negative infinity.

src/TrackGenerator.cpp
```diff
@@ -324,23 +355,30 @@ FP_PRECISION TrackGenerator::getFSRVolume(int fsr_id) {
...
-  for (int i=0; i < _num_azim; i++) {
-    for (int j=0; j < _num_tracks[i]; j++) {
-      for (int s=0; s < _tracks[i][j].getNumSegments(); s++) {
-        curr_segment = _tracks[i][j].getSegment(s);
-        length = curr_segment->_length;
-        material = curr_segment->_material;
-        sigma_t = material->getSigmaT();
+#pragma omp parallel
+  {
+
```
wbinventor: Remove blank line

src/TrackGenerator.cpp
```diff
@@ -324,23 +355,30 @@ FP_PRECISION TrackGenerator::getFSRVolume(int fsr_id) {
...
+    segment* curr_segment;
+    FP_PRECISION length;
+    Material* material;
+    FP_PRECISION* sigma_t;
 
-        for (int e=0; e < material->getNumEnergyGroups(); e++)
-          max_optical_length = std::max(max_optical_length, length*sigma_t[e]);
+    /* Iterate over all tracks, segments, groups to find max optical length */
+    for (int i=0; i < _num_azim; i++) {
+#pragma omp for reduction(max:max_optical_length)
```
wbinventor: Is this the parallel region which necessitated the introduction of `TAU_NUDGE`?

src/ParallelHashMap.h
samuelshaner: I think they are due to @geogunow not having something in his vi settings file to remove all extra whitespace when the file is saved like I do for emacs. Github doesn't do show whitespace added on new lines so it is kind of hard to catch this on new edits, unless the new edit is just adding whitespace to a blank line.

src/ParallelHashMap.h
samuelshaner: I was just fixing the whitespace in this PR. @geogunow fixes the indents in his current PR #265

src/TrackGenerator.cpp
```diff
@@ -1269,7 +1330,7 @@ void TrackGenerator::segmentize() {
 
     /* Loop over all Tracks */
     for (int i=0; i < _num_azim; i++) {
-#pragma omp parallel for firstprivate(track)
```
wbinventor: Whoa, this looks like it may have been a bug before

src/ParallelHashMap.h
samuelshaner: The only change is whitespace being removed. Blame VI!!!

src/Solver.cpp
samuelshaner: Nope, I'll add one since. I did mess this up in one of my commits and was not computing the number of fissionable FSRs correctly so this will be a good test to have.

src/TrackGenerator.cpp
samuelshaner: Nope, just unnecessary firstprivate. No need to initialize track to an uninitialized pointer.

src/TrackGenerator.cpp
samuelshaner: The max optical length is first computed here, but the parallel region in `TrackGenerator::splitSegments()` where segments are cut up based on their length w.r.t. the max optical path length is where the `TAU_NUDGE` manifested itself. This parallel region *might* bear some responsibility as well though.

src/TrackGenerator.cpp
```diff
@@ -1917,15 +1990,30 @@ void TrackGenerator::initializeSegments() {
...
-    for (int j=0; j < _num_tracks[i]; j++) {
-      for (int s=0; s < _tracks[i][j].getNumSegments(); s++) {
-        curr_segment = _tracks[i][j].getSegment(s);
-	region_id = curr_segment->_region_id;
-        Material* mat = _geometry->findFSRMaterial(region_id);
-	curr_segment->_material = _geometry->findFSRMaterial(region_id);
-	FSR_keys_map->at(FSRs_to_keys->at(region_id))->_mat_id = mat->getId();
+#pragma omp parallel
+  {
+
```
wbinventor: remove blank line

src/Universe.cpp
```diff
@@ -53,7 +53,7 @@ Universe::Universe(const int id, const char* name) {
 
   /* If the user did not define an optional ID, create one */
   if (id == -1)
-    _id = cell_id();
+    _id = universe_id();
```
wbinventor: well this was a bug!

samuelshaner: indeed!

tests/test_1d_gradient/results_true.dat
```diff
@@ -1,203 +1,203 @@
 # Iterations: 47
 keff:  1.68873E-01
 fluxes:
-1.970213E+00
-4.070833E-01
-2.320673E+00
-5.841409E-01
-1.937412E+00
-4.023670E-01
+1.970212E+00
```
wbinventor: remind me why all of the results changed?

tests/test_2d_gradient/results_true.dat
```diff
@@ -1,203 +1,203 @@
...
+7.983337E-01
 1.830133E+00
 7.384444E-01
-1.958458E+00
-7.875505E-01
-2.003656E+00
-8.037618E-01
-1.958458E+00
-7.875505E-01
+1.958457E+00
```
wbinventor: I'm curious if these results were updated in single precision rather than double?

src/Universe.cpp
samuelshaner: I also see the default for cell id in the constructor is 0 while the default universe id in the constructor is -1. I feel like they should both be -1 (or at least the same value). Agreed?

tests/test_correct_fsr_volume/test_correct_fsr_volume.py
```diff
@@ -0,0 +1,46 @@
...
+
+        # Correct FSR volumes with the appropriate number of threads
+        self.track_generator.setNumThreads(self.num_threads)
+        old_volume = self.track_generator.getFSRVolume(1)
+        self.track_generator.correctFSRVolume(1, 2.7)
+        new_volume = self.track_generator.getFSRVolume(1)
+
+        # Save results for one thread
+        self._result += '{0: 1.6f}'.format(old_volume)
+        self._result += '{0: 1.1f}'.format(new_volume)
```
wbinventor: Should we report more decimal places for the new volume? I feel like we should since we want to ensure that the new volume - computed from the adjusted segment lengths - accurately matches the value assigned by the user to more than a single decimal place.

tests/test_correct_fsr_volume/test_correct_fsr_volume.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def _get_results(self, num_iters=False, keff=False, fluxes=False,
+                     num_fsrs=True, num_segments=True, num_tracks=True,
```
wbinventor: Set the defaults for `num_fsrs`, `num_segments` and `num_tracks` to `False` here. I don't like the way we have these keyword parameters defined but left unused for subclassed methods throughout our test suite, but will leave that for another day.

tests/test_multisim_num_azim/results_true.dat
```diff
@@ -1,4 +1,4 @@
-# tracks: 116	# segments: 198
+# tracks: 116	# segments: 196
```
wbinventor: So 196 is the "correct" value here?

tests/test_1d_gradient/results_true.dat
samuelshaner: due to the split segments change. They split the segment that had the max optical path length. As you can see, not all the tests are corrected suggesting the decision of whether to split the segment with the max optical path length was not consistent with this version of the code.

tests/test_split_segments/test_split_segments.py
```diff
@@ -0,0 +1,34 @@
...
+import os
+import sys
+sys.path.insert(0, os.pardir)
+sys.path.insert(0, os.path.join(os.pardir, 'openmoc'))
+from testing_harness import TestHarness
+from input_set import PinCellInput
+
+
+class SplitSegmentsTestHarness(TestHarness):
+    """An eigenvalue calculation in a pin cell with 7-group C5G7 data."""
```
wbinventor: Update this docstring to reflect the purpose of the test.

tests/test_2d_gradient/results_true.dat
samuelshaner: They were recomputed with:

    python update_results.py -R test_2d_gradient

so the code was rebuilt and the test was run. It looks like `update_results.py` uses double precision.

tests/test_1d_gradient/results_true.dat
wbinventor: Really, wow. So in the past we *always* split at least one track segment? The one with the max tau if it was less than the overall max tau allowed?

tests/test_multisim_num_azim/results_true.dat
samuelshaner: Correct, 196 segments is the number of segments when no segments are split, which is the desired behavior.

tests/test_split_segments/test_split_segments.py
samuelshaner: ok

wbinventor:
I'd like to see a test for the number of fissionable FSRs to ensure that the parallelism around that code region is working as expected. Otherwise I think this is ready to go!

tests/test_1d_gradient/results_true.dat
samuelshaner: I believe in the past we *sometimes* split at least one track segment. Now we *never* split a segment (as long as the max optical length used to split the segments is based on the max optical path of the existing segments). Thus, not all tests need to be updated.

tests/test_correct_fsr_volume/test_correct_fsr_volume.py
samuelshaner: Sure, I'll make it 10. I will note that Python 2 and Python 3 were reporting different values for the last decimal place (I think it was beyond 10 places).

wbinventor:
I made extensive use of the code in this PR today and it does indeed perform very well!

samuelshaner:
@wbinventor - I just updated the PR based on your comments. Thanks for the extensive review!

In my updates, I added test for segment splitting with CMFD and checking the procedure in the solver for computing the number of fissionable FSRs.

tests/test_split_segments/test_split_segments.py
```diff
@@ -0,0 +1,34 @@
...
+
+        # Set a small max optical path length so segments are split
+        self.solver.setMaxOpticalLength(0.5)
+
+        super(SplitSegmentsTestHarness, self)._run_openmoc()
+
+    def _get_results(self):
+        """Digest info in the results and return as a string."""
+        return super(SplitSegmentsTestHarness, self).\
+            _get_results(num_segments=True)
```
wbinventor: Do we really need the eigenvalue and flux here? If not, set the `fluxes=False` and `keff=False` keyword parameters in the call to the super class `TestHarness._get_results()` method.

tests/test_split_segments_cmfd/results_true.dat
```diff
@@ -0,0 +1,60 @@
+# Iterations: 25
+keff:  1.04701E+00
```
wbinventor: Ditto, doesn't seem like the fluxes and eigenvalue are needed in the results for this test.

wbinventor:
Everything works out...except for the SPH factors input files which now segfault. From a very preliminary look at this, it appears that the "multi-sim" pattern used by the SPH factor calculator is no longer valid, leading to the segfault.

wbinventor:
I'll add that the SPH factor cases *do* still work for the "develop" branch. Also, I would like to add the SPH cases to the test suite, but this appears to be quite challenging since OpenCG is a private repo, which can't (easily) be included as a pre-requisite install for Travis.

wbinventor:
When compiled with `--debug-mode` and then run with GDB, I see the following error trace for the "pin-cell" SPH factor case:

```bash
#9  Geometry::findCellContainingFSR (this=0x19501c0, fsr_id=fsr_id@entry=0)
    at src/Geometry.cpp:1196
#10 0x00007ffff5cac349 in Geometry::findFSRMaterial (this=<optimized out>, 
    fsr_id=fsr_id@entry=0) at src/Geometry.cpp:471
#11 0x00007ffff5cd1582 in TrackGenerator::initializeSegments ()
    at src/TrackGenerator.cpp:2003
#12 0x00007ffff5cd1a76 in TrackGenerator::initializeSegments (this=0x1a31180)
    at src/TrackGenerator.cpp:1992
#13 0x00007ffff5cba064 in Solver::initializeFSRs (this=this@entry=0x1a0aed0)
    at src/CPUSolver.cpp:134
#15 0x00007ffff5cba9df in Solver::computeEigenvalue (
    this=this@entry=0x1a0aed0, max_iters=max_iters@entry=1000, 
    mode=mode@entry=FORWARD, res_type=FISSION_SOURCE) at src/Solver.cpp:1039
#16 0x00007ffff5c78286 in _wrap_Solver_computeEigenvalue (
    args=<optimized out>, kwargs=<optimized out>)
    at openmoc/openmoc_wrap.cpp:17082
```

I've redacted the part of the trace that is not tied back directly to the OpenMOC source code.

samuelshaner:
It seems to work just fine for the SPH pin cell problem when run serially, but fails in parallel. Hmmm...looking at the code I'm not quite sure what the issue is.

wbinventor:
I'm not sure either without really digging into the problem. Might be worth starting by commenting out parallel regions in `TrackGenerator` (i.e., canceling out the changes you made in this PR) go identify the problematic one (and it may not be segment initialization...these multi-sim issues can be really tricky to identify).

src/ParallelHashMap.h
geogunow: I don't understand. It seems you're adding whitespace.

samuelshaner:
It keeps failing for me during the `FixedHashMap<std::string, fsr_data*>::at(std::string)` call in the `ParallelHashMap<std::string, fsr_data*>::at(std::string)` function.

@geogunow - are ideas on whether there might be a thread safety issue in `FixedHashMap<std::string, fsr_data*>::at(std::string)`?

wbinventor:
The fact that it consistently fails with this PR vs. "develop" would
indicate that *if* it is an issue with the hash map, it is a new one, or at
least newly exposed.

samuelshaner:
I believe it is just a newly exposed issue with the parallel hash map. I just checked out develop and put a `pragma omp parallel` and `pragma omp for` around the loops in `TrackGenerator::initializeSegments()` and get the same error so I would argue that it is an issue with `ParallelHashMap`.

Perhaps we should go through @geogunow's PR first to fix this issue and then come back to this PR?

wbinventor:
One thing you might consider doing is adding a new "test_multisim_flux" test
This could simply be the same as the "test_multisim_fixed_source" test but
would run a fixed source calculation instead (the naming is messed up, the
existing fixed source test actually computes the source given an external
source so perhaps should be renamed to "test_multisim_ext_source")

Anyway...the reason I suggest this is that this new test would mimic the
multi-sim pattern in the SPH calculator, which runs many fixed source
calculations in sequence on the same geometry and with the same tracks. But
the test would cut away a lot of the extra data processing that is needed
for SPH.

wbinventor:
Did you try merging @geogunow's changes in from his latest PR? Perhaps that
resolves the issue.

samuelshaner:
Tried that and it doesn't fix it :(

geogunow:
@samuelshaner is it a seg fault?

samuelshaner:
yes, it is a seg fault

geogunow:
Can you give me the line of code where ``ParallelHashMap`` is called?

geogunow:
Is it perhaps line 925 in ``Geometry.cpp`` because that line has an incorrect usage of ``ParallelHashMap`` (with my update).

samuelshaner:
It is line 1196 of `Geometry.cpp` and the line in `ParallelHashMap.h` that is failing is line 215.

wbinventor:
I know it's not what you want to hear, but I wouldn't be convinced that
this bug is from the parallel hash map. I spent a non-trivial amount of
time debugging memory allocation/deallocation issues to enable multi-sim,
and in most of those cases, GDB did *not* point me to the correct line of
code for a segfault. Instead, it often pointed me to the line of code where
an error prevailed to cause a segfault (but not always in a
non-deterministic way), but typically the error came from something else.
If this segfault is indeed 100% reproducible and always points to the
parallel hash map then that may give us some confidence that it is
originating there, but I wouldn't bet on it. I'd definitely add in the test
I suggested and if it fails with a segfault, it will be easier to debug
than SPH.

geogunow:
@samuelshaner try merging my fix now

samuelshaner:
@geogunow - Just tried that out and got a seg fault at the same place.

geogunow:
@samuelshaner then I'm not sure how the problem could be with ``ParallelHashMap``

samuelshaner:
Okay, all tests are passing but I can't find the source of the issue with sph solver. Was it introduced in this PR? Perhaps. Or was it merely exposed by this PR? Perhaps. I've spent a lot of time trying to find the root cause but came up empty.

@geogunow and @wbinventor - what's the call? My vote is to merge this in.

wbinventor:
I don't want to merge it in if it breaks the SPH calculator. That took
several months to implement and fully debug all of the memory allocation
issues. Even though it is not part of the test suite, it is something that
I and perhaps others do need to use. Let's talk about what we might do to
track down the source of the issue.

wbinventor:
@samuelshaner how did you approach debugging this issue? The key here may be to abstract away as much of the SPH calculator as possible to identify the source of the bug. I feel your pain, this appears to be amongst the toughest class of bugs to sort out. 

Did you try creating a multi-sim test for `Solver::computeFlux()` to mirror "test_multi_sim_fixed_source" for `Solver::computeSource()`? If so, did that test pass or fail? I'd bet my money that such a test would have a good chance of failing if SPH is failing.

However, if such a test passes, you might construct a slightly more complicated multi-sim test based on this to better mirror the SPH scheme. In particular, consider something combining the following snippets of "openmoc/materialize.py":

* lines [450-457](https://github.com/mit-crpg/OpenMOC/blob/develop/openmoc/materialize.py#L449) to initialize a **single** `TrackGenerator` and `Solver`
* lines [511-525](https://github.com/mit-crpg/OpenMOC/blob/develop/openmoc/materialize.py#L511) to iterate over fixed source flux calculations
* line [552](https://github.com/mit-crpg/OpenMOC/blob/develop/openmoc/materialize.py#L552) to load an `openmc.mgxs.Library` on each iteration (if test with two snippets above passes)

This will strip away the cross section updates to really isolate the code in the SPH scheme which is causing a failure. Note that a test with the third bullet point implement will work on your local machine with an accessible `openmc.mgxs.Library`, but will not on Travis without installing OpenCG and OpenMC's Python API.

Finally, it looks like GitHub can't merge this PR in anyway due to merge conflicts. Any idea why? Perhaps this is due to the discretization PRs I just merged in?

samuelshaner:
This PR was broken up and addressed in PRs #276 #277 #278 #279 #280 #281 #282 #283 #285

PR #266, opened 2016-02-16
Multi-Simulation Tests

wbinventor:
This PR adds seven new "multi-simulation" tests to the suite. These tests are intended to ensure that one can run multiple simulations - with different runtime parameters, discretizations, materials, etc. - in the same Python script. With these tests I hope that we can avoid running into the memory deallocation issues that led to a number of my PRs over the last few months. 

Although @samuelshaner may want to review this in order to merge in #264, it is probably also a good idea for @geogunow to review this to have an idea of some of the "design criteria" that we need to meet when sync'ing up the 3D MOC code.


tests/test_multisim_fixed_source/test_multisim_fixed_source.py
```diff
@@ -0,0 +1,125 @@
...
+        water_univ = openmoc.Universe(name='water')
+        water_univ.addCell(water_cell)
+
+        self.source_cell = openmoc.Cell(name='source')
+        self.source_cell.setFill(self.materials['Water'])
+        source_univ = openmoc.Universe(name='source')
+        source_univ.addCell(self.source_cell)
+
+        # Create 2D array of Universes in each lattice cell
+        universes = [[[water_univ]*num_x for _ in range(num_y)]]
```
samuelshaner: Could this just be:

    universes = [[[water_univ]*num_x] * num_y]

wbinventor: Actually no - this will produce a faulty result which is why I used this notation in the equivalent "point-source" sample input file. The reason is that multiplying a nested list actually creates a list of pointers. For example, the current notation will produce the following homogeneous array of cells with a tiny green fixed source cell:

cells-z-0


Your simpler notation however would produce the following vertical line of fixed source cells:

cells-z-0

samuelshaner: hmmm...interesting. I didn't know that.

samuelshaner:
This PR looks really good. I'm glad we are going to have these multi-sim tests in there and have that capability supported. As soon as the merge conflicts are resolved I think it is ready to merge.

samuelshaner:
I test merged the wbinventor/multi-sim branch into the mit-crpg/develop branch and ran the tests on my laptop and AWS with matplotlib 1.5.1 installed and all the plotting tests failed. I'm perplexed by what is going on. Are there any issues you could think of? Any other packages that need to be updated (e.g. PIL, numpy, etc)?

geogunow:
@samuelshaner I've run into issues in the past with Travis. Make a trivial change and re-submit. Maybe when it runs again it will be fine? But we really should get to the bottom of this.

wbinventor:
This doesn't appear to be a Travis issue. Instead it is an issue where
matplotlib created plots of slightly different pixel counts on different
machines. When the testing harness goes to hash those figures, although
they appear the same, they hash to a different result. This can be seen by
running the suite on your local machine and a cluster with the `_cleanup()`
method commented out so you can view the resultant images.

I thought I had fixed this by using matplotlib 1.5 (Anaconda's default) but
it appears not. And it also appears that we aren't the first to face this
issue - there is a lot of chatter on the web about it as well. It looks
like matplotlib provides a testing interface with "fuzzy" image comparisons
to allow for stuff like this, so I'll give that a try. The latest 2.0
version claims to get around this issue entirely, but I don't think we
should make the a dependency just yet.

samuelshaner:
Looking at the commit log, the matplotlib issue seems frustrating. Thanks for working to fix it @wbinventor!

wbinventor:
Ok, so I *think* the plotting tests should work now. I made a lot of changes, most notably we now store small versions of each image for comparison, and "hash" the true and test images using histograms of the RGB values. Kind of hacky, but so is matplotlib it seems...

tests/pickle-rcparams.py
```diff
@@ -0,0 +1,294 @@
...
+          u'ytick.labelsize': u'medium',
+          u'ytick.major.pad': 4.0,
+          u'ytick.major.size': 4.0,
+          u'ytick.major.width': 0.5,
+          u'ytick.minor.pad': 4.0,
+          u'ytick.minor.size': 2.0,
+          u'ytick.minor.visible': False,
+          u'ytick.minor.width': 0.5}
+
+pickle.dump(rcparams, open('rcparams.pkl', 'wb'))
```
geogunow: I'm a bit confused about this. Why are we including this file to dump the default ``rcParams``? And a pickle for it as well?

tests/test_multisim_materials/test_multisim_materials.py
```diff
@@ -0,0 +1,48 @@
...
+            cells = self.input_set.geometry.getAllMaterialCells()
+            materials = self.input_set.geometry.getAllMaterials()
+
+            # Exchange all of the materials for their clones
+            for cell_id in cells:
+                material = cells[cell_id].getFillMaterial()
+                clone = material.clone()
+                cells[cell_id].setFill(clone)
+
+            # Turn on SWIG flag to register old Materials
```
geogunow: Do we want Materials capitalized? It is not in the previous comment.

tests/pickle-rcparams.py
wbinventor: The goal is to make the images generated on different machines as similar as possible. We are using Anaconda's matplotlib on Travis, and this matplotlib have its own default parameters which may be different from those installed with `apt-get` or `macports`. Hence, I thought it a good idea to always use Anaconda's `rcparams` for the test suite, even when run locally.

tests/test_multisim_materials_cmfd/test_multisim_materials_cmfd.py
```diff
@@ -0,0 +1,61 @@
...
+        """Run multiple OpenMOC eigenvalue calculations with CMFD with
+        different materials."""
+
+        for i in range(self.num_simulations):
+
+            # Extract all of the material-filled cells in the geometry
+            cells = self.input_set.geometry.getAllMaterialCells()
+            materials = self.input_set.geometry.getAllMaterials()
+
+            # Exchange all of the materials for their clones
```
geogunow: Ditto on consistency for capitalization of materials

tests/testing_harness.py
```diff
@@ -271,42 +274,123 @@ class PlottingTestHarness(TestHarness):
...
                 os.remove(output)
             elif os.path.isdir(output):
                 shutil.rmtree(output)
 
         super(PlottingTestHarness, self)._cleanup()
+
+    def _compare_images(self, img1, img2):
+        """Compare two PIL images using in RGB space with pixel histograms."""
+
+        # Extract RGBA data from each PIL Image
```
geogunow: RGBA or RGB?

tests/testing_harness.py
```diff
@@ -271,42 +274,123 @@ class PlottingTestHarness(TestHarness):
...
+        distance = np.sqrt(np.dot(diff, diff))
+        return distance
+
+
+class MultiSimTestHarness(TestHarness):
+    """Specialized TestHarness for testing multi-simulation capabilities."""
+
+    def __init__(self):
+        super(MultiSimTestHarness, self).__init__()
+        self.num_simulations = 3
```
geogunow: It looks here like we run the exact same test three times. I see why you might want to test multiple runs but isn't twice enough?

wbinventor: Yes you have a good point and I'm open to changing it to 2. The reason I set it to 3 is that many of the PRs I've submitted the last year to enable multiple simulations with a single `Solver`+`TrackGenerator`+`Generator` fixed memory allocation/deallocation issues that randomly broke the code. Similar to some of the issues you faced when in #270, the code would randomly segmentation fault after 2+ simulations rather than always on the second simulation. Since I was burned before, I set this to 3, though I recognize that my argument for doing so would actually indicate that there is no reason to believe that 3 is actually enough. It doesn't dramatically impact the runtime of the test suite as it stands, so I'd prefer to keep it at 3, but am willing to drop to 2 if you strongly prefer it.

tests/testing_harness.py
wbinventor: RGBA where A is the alpha value.

tests/test_multisim_materials/test_multisim_materials.py
wbinventor: Got it

tests/test_multisim_materials_cmfd/test_multisim_materials_cmfd.py
wbinventor: And this

samuelshaner:
This PR looks good to go. @geogunow - is there any additional items you would like to see addressed?

PR #268, opened 2016-02-16
Ring radius extension

friedmud:
When using rings to subdivide the moderator the rings won't reach all the way to the outer corners like so:

pasted image at 2016_02_15 07_07 pm

I've modified the selection of the maximum ring radius to allow for reaching the corners like so:

flat-source-regions-z-0 0

### DIFF
--- a/src/Geometry.cpp
+++ b/src/Geometry.cpp
@@ -780,7 +780,7 @@ void Geometry::subdivideCells() {
   }
 
   /* Recursively subdivide Cells into rings and sectors */
-  _root_universe->subdivideCells(max_radius);
+  _root_universe->subdivideCells(max_radius, _ringify_type);
 }
 
 
@@ -1164,8 +1164,7 @@ void Geometry::setFSRsToKeys(std::vector<std::string>* FSRs_to_keys) {
  * @brief Sets the method for selection of the maximum ring radius for subdivision
  * @param ringify_type The method to use for selection of maximum ring radius
  */
-void Geometry::setRingifyType(ringifyType ringify_type)
-{
+void Geometry::setRingifyType(ringifyType ringify_type) {
   _ringify_type = ringify_type;
 }
 
--- a/src/Geometry.cpp
+++ b/src/Geometry.cpp
@@ -765,9 +765,11 @@ std::string Geometry::getFSRKey(LocalCoords* coords) {
  */
 void Geometry::subdivideCells() {
 
-  /* Compute equivalent radius with the same area as the Geometry */
+  double x_rad = getWidthX()/2.0;
+  double y_rad = getWidthY()/2.0;
+
   /* This is used as the maximum radius for all ringified Cells */
-  double max_radius = sqrt(getWidthX() * getWidthY() / M_PI);
+  double max_radius = sqrt(x_rad*x_rad + y_rad*y_rad);
 
   /* Recursively subdivide Cells into rings and sectors */
   _root_universe->subdivideCells(max_radius);
--- a/src/Geometry.cpp
+++ b/src/Geometry.cpp
@@ -20,6 +20,9 @@ Geometry::Geometry() {
 
   /* Initialize CMFD object to NULL */
   _cmfd = NULL;
+
+  /* Default value for ringify type */
+  _ringify_type = EQUIVALENT_AREA;
 }
 
 
@@ -765,11 +768,16 @@ std::string Geometry::getFSRKey(LocalCoords* coords) {
  */
 void Geometry::subdivideCells() {
 
-  double x_rad = getWidthX()/2.0;
-  double y_rad = getWidthY()/2.0;
+  double max_radius = 0;
 
-  /* This is used as the maximum radius for all ringified Cells */
-  double max_radius = sqrt(x_rad*x_rad + y_rad*y_rad);
+  if (_ringify_type == EQUIVALENT_AREA)
+    max_radius = sqrt(getWidthX() * getWidthY() / M_PI);
+  else {
+    double x_rad = getWidthX()/2.0;
+    double y_rad = getWidthY()/2.0;
+
+    max_radius = sqrt(x_rad*x_rad + y_rad*y_rad);
+  }
 
   /* Recursively subdivide Cells into rings and sectors */
   _root_universe->subdivideCells(max_radius);
@@ -1149,6 +1157,16 @@ void Geometry::setFSRKeysMap(ParallelHashMap<std::string, fsr_data*>*
  */
 void Geometry::setFSRsToKeys(std::vector<std::string>* FSRs_to_keys) {
   _FSRs_to_keys = *FSRs_to_keys;
+}
+
+
+/**
+ * @brief Sets the method for selection of the maximum ring radius for subdivision
+ * @param ringify_type The method to use for selection of maximum ring radius
+ */
+void Geometry::setRingifyType(ringifyType ringify_type)
+{
+  _ringify_type = ringify_type;
 }
 
 
--- a/src/Geometry.h
+++ b/src/Geometry.h
@@ -13,6 +13,7 @@
 #include "Python.h"
 #endif
 #include "Cmfd.h"
+#include "ringify_type.h"
 #include <limits>
 #include <sys/types.h>
 #include <sys/stat.h>
@@ -25,19 +26,6 @@
 
 /** Forward declaration of Cmfd class */
 class Cmfd;
-
-/**
- * @enum ringifyType
- * @brief The method to use when subdividing cells into rings
- */
-enum ringifyType{
-
-  /** Max ring radius chosen such that it has the same area as the total geometry */
-  EQUIVALENT_AREA,
-
-  /** Max ring radius chosen to be the distance from the center to a corner of the geometry */
-  MAX_DISTANCE
-};
 
 /**
  * @struct fsr_data
--- a/src/Geometry.h
+++ b/src/Geometry.h
@@ -27,6 +27,19 @@
 class Cmfd;
 
 /**
+ * @enum ringifyType
+ * @brief The method to use when subdividing cells into rings
+ */
+enum ringifyType{
+
+  /** Max ring radius chosen such that it has the same area as the total geometry */
+  EQUIVALENT_AREA,
+
+  /** Max ring radius chosen to be the distance from the center to a corner of the geometry */
+  MAX_DISTANCE
+};
+
+/**
  * @struct fsr_data
  * @brief A fsr_data struct represents an FSR with a unique FSR ID
  *        and a characteristic point that lies within the FSR that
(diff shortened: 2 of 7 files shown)


wbinventor:
Thanks @friedmud! I think @geogunow will have some thoughts to offer on this one.

samuelshaner:
We had quite a lengthy discussion on this line a few months ago and I see pros and cons of both approaches. At the time, I wasn't really considering the case of having a ton of rings, which is clearly a deficiency of the current method.

* **current method**: This created equi-volume rings, assuming the radius of the outer ring was less than half the cell width/height.
    * Pros: Favors *smaller* radius rings, so better resolution close to the fuel where the flux gradient will be larger.
    * Cons: If lots of rings are used, the outermost ring is enormous compared to the all other rings. 
* **proposed method**: This created equi-volume rings, assuming the radius of the outer ring is equal to the distance from the center of the bounding box to the corner.
    * Pros: No concerns about large regions in the corners for large ring counts. 
    * Cons: Favors *larger* radius rings, so not as good of resolution near the fuel for low ring counts.

I can see the argument for going this new approach, but do we want the option? I would vote for adding in the option of the current method (perhaps call it the **equivalent radius method**) vs the proposed method (perhaps call it the **max radius method**) since I would expect the current method to be more accurate when only a few moderator rings are used. Thoughts?

samuelshaner:
Also, note that the line 1155 of in the `Lattice::subdivideCells(...)` method of `Universe.cpp` would need to be modified as well to properly discretize pin cells not in the root universe.

samuelshaner:
Thinking about this a bit more and talking to @wbinventor I would say we implement the option of both methods. I have a pretty clear idea of one way to do this:

* **Cell::setNumRings(int num_rings, ringifyType ringify_type)**: Allow the ringify method to accept a ringifyType (with a default value) and ringify the cell accordingly.
* **Pass both radii**: Pass both the equivalent radius and max radius to the `Universe::subdivideCells(...)` and `Cell::ringify(...)` methods.
* **Ringify cell with appropriate method**: Ringify each Cell in `Cell::ringify(...)` with the method that was specified in the `Cell::setNumRings(...)` method call. 

@friedmud - do you want to give a crack at this? If not, I'm happy to implement and PR it in.

friedmud:
Ok - I'll take a swipe at this - I'll go with the `Cell::setNumRings(int num_rings, ringifyType ringify_type)` option

friedmud:
Ok - added the option.  Look it over and let me know if anything needs to change.

friedmud:
Oh - and I don't know how you guys feel about squashing commits.  I can squash these two into one if you like.  Just let me know.

src/Geometry.h
```diff
@@ -27,6 +27,19 @@
 class Cmfd;
 
 /**
+ * @enum ringifyType
+ * @brief The method to use when subdividing cells into rings
+ */
+enum ringifyType{
+
+  /** Max ring radius chosen such that it has the same area as the total geometry */
```
samuelshaner: Can this be something like:

    /** Max ring radius chosen such that it has the same area as the bounding box of the universe it lies in */

We can solve more than just pin cells with OpenMOC!

src/Geometry.h
```diff
@@ -27,6 +27,19 @@
...
+  EQUIVALENT_AREA,
+
+  /** Max ring radius chosen to be the distance from the center to a corner of the geometry */
```
samuelshaner: ditto

src/Geometry.h
```diff
@@ -27,6 +27,19 @@
 class Cmfd;
 
 /**
+ * @enum ringifyType
+ * @brief The method to use when subdividing cells into rings
+ */
+enum ringifyType{
```
samuelshaner: Per the [google c++ style guide](https://google.github.io/styleguide/cppguide.html#Enumerator_Names), there should be a space between the `ringifyType` and `{`.

src/Geometry.cpp
```diff
@@ -1151,6 +1161,16 @@ void Geometry::setFSRsToKeys(std::vector<std::string>* FSRs_to_keys) {
 
 
 /**
+ * @brief Sets the method for selection of the maximum ring radius for subdivision
+ * @param ringify_type The method to use for selection of maximum ring radius
+ */
+void Geometry::setRingifyType(ringifyType ringify_type)
+{
```
samuelshaner: Per the [google c++ style guide](https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions), the open curly brace is always on the end of the last line of the function declaration, not the start of the next line.

samuelshaner:
I like the short and simple solution! The ringify type will need to be propagated through calls to `Universe::subdivideCells()` and `Lattice::subdivideCells()` in `Universe.cpp` to properly ringify cells in nested universes/lattices. Line 1155 of `Universe.cpp` should also be modified to reflect the option of ringify type.

src/Geometry.h
friedmud: No problem!  But just so you know... I copied the style from `universeType` here:  https://github.com/mit-crpg/OpenMOC/blob/develop/src/Universe.h#L41

I even double checked that it didn't have a space there... but I just wanted to make mine look the same ;-)

src/Geometry.h
friedmud: I suppose I don't understand... but the Geometry I'm referring to here is the current geometry... I'm using the width of the Geometry (not of a pin cell).  As far as I know I'm not querying the bounding box of the universe... I'm using the bounds of the current Geometry object.  But maybe I'm misunderstanding the code...

src/Geometry.cpp
friedmud: Heh - yes, cross project habits ;-)

friedmud:
If it needs to be propagated, why does it work as is?  Which example should I be running to show that this needs to be propagated?  Maybe an assembly that uses a Lattice?

Thanks for reviewing this!

samuelshaner:
I would recommend `sample-input/simple-lattice/simple-lattice.py` and playing around with the number of rings for one of the fuel pins.

Thanks for implementing this!

samuelshaner:
It probably works as is because you have not tested out on problems with nested universes and lattices. If you do so, all nested cells will be ringified with the old method.

friedmud:
Cool - thanks.  I'll try it out with some more complicated geometries.

src/Geometry.h
samuelshaner: Yikes, thanks for pointing that out! Perhaps you could add the space in there as well?

src/Geometry.h
samuelshaner: The subdivide cells process is a recursive process that traverses the CSG tree, subdividing all cells filled with materials as it goes. When a cell located in some universe of some lattice cell is subdivided, the bounding box of the lattice cell is used to get the max radius. For pin cell problems that don't have any nesting, the bounding box of the root universe is used and this bounding box is computed implicitly via the Geometry::getWidthX()` method. I hope that makes sense!

friedmud: I see it now that I'm working with a lattice.  Thanks for the explanation.  I'll make the changes!

src/Geometry.h
friedmud: Can do!

samuelshaner:
I have personally never squashed a commit, so I don't think it is necessary for this PR.

friedmud:
Ok - I think I addressed the comments.

Unfortunately, the simple patch got more complicated.  I ended up moving `ringifyType` out to its own header file because making `Universe.h` `#include` `Geometry.h` just wasn't a good idea.  If there is another way you would like me to address this, let me know.

src/ringify_type.h
```diff
@@ -0,0 +1,25 @@
+/**
+ * @file ringify_type.h
+ * @details The ringifyType enum.
+ * @date February 16, 2016
+ * @author Derek Gaston, MIT, Course 22 (wboyd@mit.edu)
```
samuelshaner: add your email here.

samuelshaner:
Just tested this out and everything looks good. As soon as the email is changed, I think it is ready to merge.

friedmud:
Thanks guys, I'll fix up that email first thing in the morning. I
appreciate the help with this first PR!

geogunow:
Sorry for being afk on this one. Kind of consumed by ClosedMOC at the moment. Is there not a way to combine both methods to create something that gets the best of both worlds? I think its really a deficiency in how we deal with the outermost region currently. Maybe we could treat that specific region (or regions if there's sectors) differently. @samuelshaner do you think that would be possible?

samuelshaner:
There is a solution to get the best of both worlds - linear source! That
would take considerable time though. Hmmm...I bet we could think of some
scheme to get the best of both worlds. Perhaps using the MAX_DISTANCE and
weighting the inner radii by r^3 instead of r^2. I think you could call
this weighting by the first spatial moment of the area, which makes it
sound pretty cool. This would extend the rings out to the corner, but favor
smaller rings closer to the fuel as desired to get better accuracy with
less discretization.

What do you think about that? @geogunow - do you have any ideas?

On Wednesday, February 17, 2016, Geoffrey Gunow <notifications@github.com>
wrote:

> Sorry for being afk on this one. Kind of consumed by ClosedMOC at the
> moment. Is there not a way to combine both methods to create something that
> gets the best of both worlds? I think its really a deficiency in how we
> deal with the outermost region currently. Maybe we could treat that
> specific region (or regions if there's sectors) differently. @samuelshaner
> <https://github.com/samuelshaner> do you think that would be possible?
>
> —
> <https://github.com/mit-crpg/OpenMOC/pull/268#issuecomment-185054321>.
>


-- 
Regards,
Sam Shaner
Graduate Student

friedmud:
Fixed up the email

samuelshaner:
Thanks @friedmud! I'm happy with this and think it is ready to merge. @geogunow - are you okay with merging?

geogunow:
@samuelshaner I'd have to look deeper to see what you mean exactly but it
sounds good!

samuelshaner:
To clarify, what I said before about the r^3 weighting was actually incorrect - that would make the ring next to the fuel *bigger*, not smaller. Weighting by r instead of r^2 would make the ring next to the fuel smaller. For comparison, the plots below show the `pin-cell.py` geometry with 5 moderator rings. The left two plots show the two methods in this PR and the right plot shows MAX_DISTANCE with r-weighting (the delta_radius is the same between all rings).

pin_cell_wgting

wbinventor:
Although I don't want to slowdown this process, I thought I'd point out that with @samuelshaner's work recently we have moved closer to lockstep updates of code and documentation. @geogunow recently updated the docs on radial discretization. If this PR changes that we should update the docs accordingly. I know we haven't abided by this rule at all in the past, so it may not be fair to enforce it here for the first time with @friedmud's first PR.

friedmud:
I'm all for changing the documentation as we go.  Has there been a decision on what direction we want to go with this capability?

I thought the outcome of our discussion the other day is to go to the "one true" ring laydown option (which is @samuelshaner 's modified MAX_DISTANCE above).

wbinventor:
I'm fine with this decision, but @samuelshaner and @geogunow may have stronger opinions on this than I do.

samuelshaner:
I'll create another PR that implements the new modified ring generation method and update the docs accordingly either tonight or tomorrow.

Issue #265, opened 2016-02-15
Seg fault in ParallelHashMap.h


samuelshaner:
I've been running into a seg fault when trying to run some LRA cases that appears to be in `ParallelHashMap.h` on line 190. I ran the profile version of OpenMOC with debugging turned on with an LRA-like input file included below. The bug occurs during segmentation and can be reproduced by compiling the input file below (with debugging turned on) and running with gdb. The second file below includes the gdb output showing the error.

@geogunow - any ideas on why there would be a seg fault on line 190 of `ParallelHashMap.h`?

[lra-cmfd.cpp.txt](https://github.com/mit-crpg/OpenMOC/files/131291/lra-cmfd.cpp.txt)
[lra-cmfd-output.txt](https://github.com/mit-crpg/OpenMOC/files/131298/lra-cmfd-output.txt)

geogunow:
Yeah, It's probably during a resize. I'll take a look at it.

samuelshaner:
Also, is there any reason for the `#ifdef OPENMP` preprocessor directives? Elsewhere in OpenMOC we just assume that OpenMP is present.

geogunow:
I don't think this is the problem, but I found that there will be a seg fault if the user tries to set more threads than the number of physical cores

geogunow:
I have come up with a solution that I will create a PR for tomorrow once I polish it for merging. The problem is that having an ``at`` function in ``ParallelHashMap`` that can assign values is fundamentally flawed. That's probably also why you don't find that kind of function in any mainstream parallel hash map implementation. Looks like I tried to get a little too cute when I first created it :)

wbinventor:
Closed with #270.

PR #270, opened 2016-02-16
[Hotfix] Parallel Hash Map Concurrency

geogunow:
As noted in #265, the current ``ParallelHashMap`` has the potential to seg fault. This is because the write buffer for each core doesn't necessarily get flushed immediately after a write. This isn't generally a problem, except the lock-free contains function needs to announce its presence to other threads. If this announcement gets stuck in a write buffer without getting flushed, then there is the potential during a resize for the table which the contains thread is looking at to be deleted by a thread attempting a resize. In order to fix this, ``#pragma omp flush`` directives are added after announcements to force the thread to wait for the write buffer to flush before proceeding. 

In addition, there is another error in the parallel hash map. Right now,the ``at(K key)`` function of ``ParallelHashMap`` returns a reference to the value associated with the ``key``. This allows for assignment of a new value to the key/value pair. However, this is fundamentally flawed since even a thread-safe parallel hash map will not be able to produce thread safe behavior in this case since the act of returning a reference to a value and assigning a new value are two distinct operations, thus there are no correction safeguards for this behavior as it occurs outside of ``ParallelHashMap``. Therefore the ``at(K key)`` function of ``ParallelHashMap`` is changed so that it only returns the value, not a reference. This will cause an error to be thrown any time an update to the value is attempted using ``at(K key)``. To allow values to be updated the ``update(K key, V value)`` function is added so that updates can take place completely inside the ``ParallelHashMap`` where correctness over multiple threads is ensured. This new ``update(K key, V value)`` ensures correctness by acquiring the lock associated with the ``key`` before updating the underlying hash table using ``FixedHashMap::at(K key)``.

### DIFF
--- a/src/ParallelHashMap.h
+++ b/src/ParallelHashMap.h
@@ -15,6 +15,8 @@
 #include<functional>
 #include<omp.h>
 
+#include "log.h"
+
 
 /**
  * @class FixedHashMap ParallelHashMap.h "src/ParallelHashMap.h"
@@ -211,11 +213,12 @@ V& FixedHashMap<K,V>::at(K key) {
 
   /* search bucket for key and return the corresponding value if found */
   node *iter_node = _buckets[key_hash];
-  while (iter_node != NULL)
+  while (iter_node != NULL) {
     if (iter_node->key == key)
       return iter_node->value;
     else
       iter_node = iter_node->next;
+  }
 
   /* after the bucket has been completely searched without finding the key,
      throw an exception */
@@ -293,9 +296,7 @@ int FixedHashMap<K,V>::insert_and_get_count(K key, V value) {
   /* increment counter and return number */
   size_t N;
 #pragma omp critical (node_incr)
-  {
       N = _N++;
-  }
 
   return (int) N;
 }
@@ -367,11 +368,9 @@ V* FixedHashMap<K,V>::values() {
 
   /* fill array with values */
   size_t ind = 0;
-  for (size_t i=0; i<_M; i++)
-  {
+  for (size_t i=0; i<_M; i++) {
     node *iter_node = _buckets[i];
-    while (iter_node != NULL)
-    {
+    while (iter_node != NULL) {
       values[ind] = iter_node->value;
       iter_node = iter_node->next;
       ind++;
@@ -417,9 +416,9 @@ template <class K, class V>
 void FixedHashMap<K,V>::print_buckets() {
   for (size_t i=0; i<_M; i++) {
     if (_buckets[i] == NULL)
-      std::cout << i << " -> NULL" << std::endl;
+      log_printf(NORMAL, "%d -> NULL", i);
     else
-      std::cout << i << " -> " << _buckets[i] << std::endl;
+      log_printf(NORMAL, "%d -> %p", i, _buckets[i]);
   }
 }
 
@@ -655,8 +654,7 @@ void ParallelHashMap<K,V>::resize() {
     omp_set_lock(&_locks[i]);
 
   /* recheck if resize needed */
-  if (2*_table->size() < _table->bucket_count())
-  {
+  if (2*_table->size() < _table->bucket_count()) {
     /* release locks */
     for (size_t i=0; i<_num_locks; i++)
       omp_unset_lock(&_locks[i]);
@@ -693,7 +691,8 @@ void ParallelHashMap<K,V>::resize() {
 
   /* wait for all threads to stop reading from the old table */
   for (size_t i=0; i<_num_threads; i++)
-    while (_announce[i].value == old_table) {};
+    while (_announce[i].value == old_table)
+      continue;
 
   /* free memory associated with old table */
   delete old_table;
--- a/src/ParallelHashMap.h
+++ b/src/ParallelHashMap.h
@@ -84,7 +84,7 @@ class ParallelHashMap {
     volatile long pad_L5;
     volatile long pad_L7;
     volatile long pad_L8;
-    FixedHashMap<K,V>* volatile value;
+    FixedHashMap<K,V> volatile* value;
     volatile long pad_R1;
     volatile long pad_R2;
     volatile long pad_R3;
--- a/src/ParallelHashMap.h
+++ b/src/ParallelHashMap.h
@@ -110,7 +110,8 @@ class ParallelHashMap
     ParallelHashMap(size_t M = 64, size_t L = 64);
     virtual ~ParallelHashMap();
     bool contains(K key);
-    V& at(K key);
+    V at(K key);
+    void update(K key, V value);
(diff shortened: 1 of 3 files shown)


src/ParallelHashMap.h
```diff
@@ -566,23 +565,55 @@ void ParallelHashMap<K,V>::insert(K key, V value)
...
-  #ifdef OPENMP
+#ifdef OPENMP
+  omp_unset_lock(&_locks[lock_hash]);
+#endif
+}
+
+
+/**
+ * @brief Updates the value associated with a key in the parallel hash map.
+ * @details The thread first acquires the lock for the bucket associated with
```
wbinventor: Reads well save for small typo: "acquires the lock for the bucket associated with the key is acquired" -> "acquires the lock for associated bucket"

src/ParallelHashMap.h
```diff
@@ -566,23 +565,55 @@ void ParallelHashMap<K,V>::insert(K key, V value)
...
+ *      for the key. If the key is not found, an exception is returned. When
+ *      the key is found, the value is updated and the lock is released.
+ * @param key the key of the key/value pair to be updated
+ * @param value the new value for the key/value pair
+ */
+template <class K, class V>
+void ParallelHashMap<K,V>::update(K key, V value) {
+
+  /* get lock hash */
+#ifdef OPENMP
```
wbinventor: If I understand correctly, the major change introduced in this PR is this lock around the `at(...)`?

src/ParallelHashMap.h
```diff
@@ -820,19 +860,20 @@ void ParallelHashMap<K,V>::clear()
...
-  #endif
+#endif
 
   /* get pointer to table, announce it will be searched */
   FixedHashMap<K,V> *table_ptr;
-  do{
+  do {
     table_ptr = _table;
     _announce[tid].value = table_ptr;
+#pragma omp flush(_announce)
```
wbinventor: Oh and these `flushes`

samuelshaner:
Can all the `#ifdef OPENMP` be removed? Everywhere else in OpenMOC we assume that OpenMP can be used.

geogunow:
@samuelshaner sure

src/ParallelHashMap.h
geogunow: Right. Just the ``flushes``, an updated ``at(K key)``, and a new ``update(K key, V value)`` function

src/ParallelHashMap.h
```diff
@@ -208,8 +204,8 @@ bool FixedHashMap<K,V>::contains(K key)
  * @return value associated with the given key
  */
 template <class K, class V>
-V& FixedHashMap<K,V>::at(K key)
-{
+V& FixedHashMap<K,V>::at(K key) {
+
   /* get hash into table assuming M is a power of 2, using fast modulus */
   size_t key_hash = std::hash<K>()(key) & (_M-1);
```
samuelshaner: Per the google c++ style guide, can you put braces around the while loop starting on line 214?

src/ParallelHashMap.h
```diff
@@ -273,8 +270,8 @@ void FixedHashMap<K,V>::insert(K key, V value)
  *      key was already present in map.
  */
 template <class K, class V>
-int FixedHashMap<K,V>::insert_and_get_count(K key, V value)
-{
+int FixedHashMap<K,V>::insert_and_get_count(K key, V value) {
+
   /* get hash into table using fast modulus */
   size_t key_hash = std::hash<K>()(key) & (_M-1);
```
samuelshaner: Can you remove the braces around the critical statement on line 298. Also what effect of the `(node_incr)` after critical? I've never seen a variable in parentheses after critical before and am more curious what that does.

src/ParallelHashMap.h
```diff
@@ -363,8 +360,8 @@ K* FixedHashMap<K,V>::keys()
  *      key/value pairs in the table.
 */
 template <class K, class V>
-V* FixedHashMap<K,V>::values()
-{
+V* FixedHashMap<K,V>::values() {
+
   /* allocate array of values */
   V *values = new V[_N];
```
samuelshaner: Can you place the `{` on lines 371 and 374 after their corresponding statements on the previous line?

src/ParallelHashMap.h
```diff
@@ -417,10 +414,8 @@ void FixedHashMap<K,V>::clear()
  *      screen.
  */
 template <class K, class V>
-void FixedHashMap<K,V>::print_buckets()
-{
-  for (size_t i=0; i<_M; i++)
-  {
+void FixedHashMap<K,V>::print_buckets() {
```
samuelshaner: Can you change this method to use `log_printf()` instead of `std::cout`?

src/ParallelHashMap.h
```diff
@@ -639,22 +648,18 @@ int ParallelHashMap<K,V>::insert_and_get_count(K key, V value)
...
+
   /* acquire all locks in order */
-  #ifdef OPENMP
   for (size_t i=0; i<_num_locks; i++)
     omp_set_lock(&_locks[i]);
-  #endif
 
   /* recheck if resize needed */
   if (2*_table->size() < _table->bucket_count())
   {
```
samuelshaner: This should be on the line with the if statement

src/ParallelHashMap.h
```diff
@@ -676,16 +681,15 @@ void ParallelHashMap<K,V>::resize()
...
-  #endif
 
   /* delete key and value list */
-  delete[] key_list;
-  delete[] value_list;
+  delete [] key_list;
+  delete [] value_list;
 
   /* wait for all threads to stop reading from the old table */
   for (size_t i=0; i<_num_threads; i++)
```
samuelshaner: Per the google c++, the while loop on line 692 should have `continue` after it instead of empty brackets.

src/ParallelHashMap.h
geogunow: I didn't put it there. I thought you did?

geogunow: I previously had ``#pragma omp atomic capture`` on that line but someone must have changed it to ``#pragma omp critical(node_incr)``

samuelshaner: According to the file history, Tim He did. Since this PR modifies some things in `ParallelHashMap.h` I was just looking for anything that should be updated in this file. Can you just remove the braces?

geogunow: I just did that in my last push

geogunow:
@samuelshaner I believe I have now addressed all your comments

wbinventor:
Any reason to delay this further?

samuelshaner:
I think this PR is ready to go. Since @geogunow is really the expert on this, I just want to get his confirmation he believes this is ready before merging it in.

PR #272, opened 2016-02-19
Standalone FSR Centroid Initialization

wbinventor:
This short PR refactors `TrackGenerator::generateFSRCentroids()` to allow it to be called from Python. I've found this method to be useful if one wishes to query the centroids without actually running a calculation with a `Solver` object.

### DIFF
--- a/src/Solver.cpp
+++ b/src/Solver.cpp
@@ -596,7 +596,7 @@ void Solver::initializeFSRs() {
   _FSR_volumes = _track_generator->getFSRVolumes();
 
   /* Generate the FSR centroids */
-  _track_generator->generateFSRCentroids(_FSR_volumes);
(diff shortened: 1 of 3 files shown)


samuelshaner:
Looks good to me. Any idea why the test suite is not being run?

wbinventor:
That is a very good question. The test suite did run and passed on my commit as shown [here](https://travis-ci.org/wbinventor/OpenMOC). I'll look into this right now and see if there is a reason for this.

wbinventor:
Well, looks like the web hook didn't trigger Travis the first time. Closing and re-opening did the trick!

samuelshaner:
Good to know! As soon as it (presumably) passes, I'll merge it in.

PR #273, opened 2016-02-23
Linear Source Approximation

samuelshaner:
This PR introduces a solver that uses the linear source approximation as described by Ferrer, Rhodes, and Smith [1,2]. The list below contains the remaining work to be done before this PR should be considered ready to merge. 

c5g7_lsr
- [x] **ExpEvaluator look-up table**: The ExpEvaluator look-up table is implemented for the linear source solver and the three additional exponential expressions present, but the impact of the table size on problem error needs to be better quantified.
- [ ] **Plotting LSR flux**: The linear source region flux varies across a region and therefore the flux can be queried at any point. A rudimentary plotting function is included in this PR, but should be more formally integrated into the plotting module.
- [x] **Linear Source Tests**: Tests should be written for the linear source solver.
- [x] **Documentation**: The documentation should be updated for both the flat and linear source solvers to reflect changes in this PR and changes over the last few months since the documentation was last updated.
- [ ] **FSR -> SR**: The use of FSR should be generalized to SR throughout the code.
- [ ] **Linear source in GPUSolver/VectorizedSolver**: The linear source approximation was only implemented for the CPUSolver. Implementing it for the GPUSolver and possibly VectorizedSolver should be considered.

The linear source solver, `CPULSSolver`, has been tested on homogeneous, 1D-gradient, 2D-gradient, and C5G7 test problems. In all cases, the LS solver was able to converge the flux/pin powers and eigenvalue with many fewer source regions than the FS solver. 

**1-D Gradient**: Homogeneous material with one-group xs and a vacuum boundary on the left side.

one_d_lsr

**2-D Gradient**: Homogeneous material with one-group xs and a vacuum boundary on the right and bottom sides.

two_d_lsr

**C5G7**: C5G7 geometry with coarse mesh source region discretization. These represent cases 1 and 4 from Ferrer [1,2] and the eigenvalue and pin power errors for these cases show good agreement with the reported values from literature.

c5g7_pp_comp_ls3

**References**
[1] R. Ferrer and J. Rhodes III, "A Linear Source Approximation Scheme for the Method of Characteristics", Nuclear Science and Engineering, Volume 182, Issue 2, 2016.
[2] R. Ferrer, J. Rhodes, and K. Smith, "Linear Source Approximation in CASMO 5", PHYSOR 2012.


wbinventor:
This new functionality looks pretty awesome @samuelshaner! Do you want to take charge of reviewing this one @geogunow since you are more familiar with higher order source approximations? I'm happy to jump in too and test things out, but probably won't dive in and verify the implementation of the mathematical formulation.

Also, it looks like this PR builds upon #264 is that correct? I think we should fix the issues with that PR and merge it in before this one, though that is probably not an issue since it looks like there is still some work to do before this one will be ready.

samuelshaner:
Correct, this builds on PR #264 and this PR should shrink quite a bit once that one is merged. I'll update the documentation for the MOC solver and go through all the math for the LS method, so \*hopefully\* the reviewer can verify the math by reading through the docs and following along in the code.

samuelshaner:
In order to compare the FS and LS solvers, I ran a series of trials for the 2D C5G7 test case with 128 azimuthal angles, 0.05 cm track spacing, and 3 polar angles. The table below provides the "grind time" (time to integrate across one segment divided by number of polar angles times the number of energy groups) to compare the kernels in the two source approximations. The NSE article by Ferrer shows an increase in MOC solve time of ~1.5x when going from FS to LS while we see ~2x increase in MOC solve time. There is room for improving this by combining the flat and linear components of the flux and source arrays into contiguous arrays, but this might negatively impact code readability, so the flat and linear components of the flux and source have been kept in separate arrays. Additionally, caching the segment start and end points (in local coordinates) might further reduce the run time of the LS solver.

Num Threads* | LS grind time** | FS grind time** | LS / FS | LS Strong Scaling | FS Strong Scaling |
:------------------:|:--------------------:|:---------------------:|:-----------:|:-----------------------:|:------------------------:|
             1         |    17.75             |           8.67         |   2.05     |            -                |                 -             |
             2         |     9.03              |           4.40         |   2.05     |            1.97          |                 1.97       |
             4         |     4.93              |           2.39         |   2.06     |            3.60          |                 3.63       |
             8         |     2.59              |           1.29         |   2.01     |            6.85          |                 7.88       |
\* All tests were run on an AWS c4.4xlarge node (2 Intel Xeon E5-2666 2.9 GHz processors, each with 8 vCPUs [4 hyperthreaded CPUs], and 30 GiB shared memory)
 \** units of (nanoseconds / num segments / num energy groups / num polar angles)

The 1D slab problem was further investigated to determine the order of convergence of the L2 norm and L-inf norm of the one group flux as the number of discrete source regions is increased. The plots below show the flux profile produced by the FS solver converges ~linearly (i.e. O(h^1)) with source region width, while the flux profile by the LS solver converges ~quadratically (i.e. O(h^2)) with source region width, as expected.

flux_l2_norm

flux_linf_norm

Another concern that was further investigated was the error on the exponential terms. The F2, G1, and H exponentials (shown below) can incur significant error for optically thin regions when using a first-order Taylor series expansion interpolation table, as alluded to by Ferrer in the NSE article. Thus, the ExpEvaluator was changed to use a second-order Taylor series expansion interpolation table. It is important to note that only one interpolation table is created for the F1 exponential; all other exponentials are approximated by interpolating the F1 exponential at the desired optical length and then using the formula for the exponential that includes the F1 term to approximate the desired exponential term (e.g. F1, G1, or H). 

The error and relative error plots show that, while significant relative error is incurred in optically thin regions for the F2, G1, and H exponentials, the magnitude of the error is quite small as these exponentials are close to zero for optically thin regions. 

I will also note that the G2 exponential is computed using the exponential intrinsic function to avoid approximation error. This shouldn't impact the run time very much because the terms that use the G2 exponential are precomputed before the MOC solve.

exp_intrin

exp_error

exp_rel_error

All this evaluation will be included in the documentation that will eventually be written for this PR.

wbinventor:
Wow! The preliminary results for this PR look remarkable. It sounds like any MOC code worth its weight definitely should be using at least a linear source approximation. I'm def stoked to use this in my research soon.

Just curious, can you add the number of segments and nber of FSRs to you analysis of the grind times? Would be nice to round that out with the metrics needed to compute and compare the total runtimes for a given accuracy.

samuelshaner:
Sure, I just updated my comment with the number of segments and number of source regions for the case I used in the analysis.

wbinventor:
Can you add the number of segments and FSRs for the fixed source case for comparison?

wbinventor:
@samuelshaner any idea of a rough ETA for when this PR might be ready for showtime?

samuelshaner:
I'll shoot to have it ready at the end of this week (friday 5/20). I've been writing up documentation on the implementation of the stable exponentials this weekend (see attached pdf). In stabilizing the exponentials, the MOC equations start to look a bit different than what we are typically used to (and what is shown in Mike Ferrer's paper), so I want to make sure it is well documented so future developers are not confused when they look at the terms.

[moc_ls_exponentials.pdf](https://github.com/mit-crpg/OpenMOC/files/265482/moc_ls_exponentials.pdf)

wbinventor:
Sounds good, just let me know when to start reviewing / testing things out.
I'll read over the math at that point. Will probably need @geogunow's
prowess at this as well.

GiudGiud:
Hi Sam, thanks for the PR

@tjlaboss needs this for research and I think it'd be a good first step towards bringing 3D MOC to develop, so I will merge this soon after testing it extensively.

geogunow:
Hey Guillaume,

Be aware that there might be some bugs in that linear source
implementation. It would be best to compare with my 3D MOC implementation
(which handles both 2D and 3D MOC) since it has many bug fixes in it -
including corrections for strange behavior in optically thin regions.

Best regards,
Geoff

GiudGiud:
If anything major I can't fix shows up when I test it, I wont merge that one.
Long term goal is to have your implementation of the solvers, track generation etc replace the current one, but with continuity in the test suite. My idea is that I would rather merge your code, which I've been working with for a year now, into develop,  than merge develop into 3D-MOC. To make that easier, we'll ramp up develop's capabilities.

GiudGiud:
Merged indirectly through #431

PR #274, opened 2016-02-23
changed sectorize to create diagonal plane for first sector division

samuelshaner:
This PR changes the `Cell:sectorize()` method to use a diagonal plane to create the first sector division instead of a horizontal plane, as is done in CASMO:

sectorize


wbinventor:
The documentation needs to be updated. The Travis build seems to fail since SWIG wasn't installed correctly for some reason. I'm going to close and then re-open to see if that triggers it to install...?

wbinventor:
As with #275, this should be quick merge once the docs are updated and the Travis build works.

wbinventor:
Well looks like Travis was flaking out earlier - closing and reopening triggered it to install swig correctly and the tests all pass.

samuelshaner:
The documentation has been updated to reflect the new sectorize methodology.

PR #284, opened 2016-02-24
Exponential table look-up evaluated at interval mid-point

samuelshaner:
This PR changes the exponential table to store the slope and intercept values of the exponential at the midpoint of the interval instead of at the start point. The documentation and tests have been updated to reflect this change. The plots below show the error in the exponential when evaluated using the start point (left) and mid-point (right):

exp_rel_error_comp

The plot below shows the error in the exponential using the default exponential precision of 1e-5 in the tau interval [0.3, 0.303]:

exp_old_new_labeled

The integrated error is about **4x** larger when using the start point compared with the mid-point.


wbinventor:
Looks good to me!

Issue #286, opened 2016-03-04
Images in Documentation Not Showing Up


wbinventor:
I've been updating some of the documentation on `openmoc.materialize` and realized that many of the images in the section on radial/angular discretization are not tracked by Git. There were some recent PRs by @geogunow and @samuelshaner to update this section - when you guys build the docs, do all of the images show up properly? Here is the list of warnings (really, errors) that I see when I run sphinx:

```bash
/home/wboyd/Documents/NSE-CRPG-Codes/OpenMOC/docs/source/usersguide/input.rst:402: WARNING: Duplicate explicit target name: "figa".
/home/wboyd/Documents/NSE-CRPG-Codes/OpenMOC/docs/source/usersguide/input.rst:402: WARNING: Duplicate explicit target name: "figb".
/home/wboyd/Documents/NSE-CRPG-Codes/OpenMOC/docs/source/usersguide/input.rst:428: WARNING: Duplicate explicit target name: "figa".
/home/wboyd/Documents/NSE-CRPG-Codes/OpenMOC/docs/source/usersguide/input.rst:428: WARNING: Duplicate explicit target name: "figb".
/home/wboyd/Documents/NSE-CRPG-Codes/OpenMOC/docs/source/usersguide/input.rst:457: WARNING: Duplicate explicit target name: "figa".
/home/wboyd/Documents/NSE-CRPG-Codes/OpenMOC/docs/source/usersguide/input.rst:457: WARNING: Duplicate explicit target name: "figb".
/home/wboyd/Documents/NSE-CRPG-Codes/OpenMOC/docs/source/usersguide/input.rst:632: WARNING: Duplicate explicit target name: "figa".
/home/wboyd/Documents/NSE-CRPG-Codes/OpenMOC/docs/source/usersguide/input.rst:632: WARNING: Duplicate explicit target name: "figb".
/home/wboyd/Documents/NSE-CRPG-Codes/OpenMOC/docs/source/usersguide/input.rst:670: WARNING: Duplicate explicit target name: "figa".
/home/wboyd/Documents/NSE-CRPG-Codes/OpenMOC/docs/source/usersguide/input.rst:670: WARNING: Duplicate explicit target name: "figb".
/home/wboyd/Documents/NSE-CRPG-Codes/OpenMOC/docs/source/usersguide/input.rst:None: WARNING: image file not readable: usersguide/../../img/annular_pin_materials.png
[...]

PR #287, opened 2016-03-04
Added png files not currently being tracked and fixed linking issues in rst files

samuelshaner:
This PR addresses issue #286 by adding the images not showing up in the documentation and fixing a few issues with linking in the rst files.

Some extra whitespace was also removed from `cmfd.rst`.


wbinventor:
Thanks for the quick fix @samuelshaner! The manual image linking is a real shortcoming in Sphinx I think, though perhaps there is a feature to do it automatically that I am unaware of.

Issue #194, opened 2015-08-05
Update documentation


wbinventor:
This is intended to start a conversation on the various features that have been implemented to the code over the last 6 months and which should be added to the documentation. Some items that come to mind that should be added to the online documentation before the next release include:

**User's Guide**
* polar quadrature API
* fixed source calculation API
* FSR volume correction API
* adjoint calculations
* plotting centroids, eigenmodes

**Theory & Methodology**
* *k*-nearest neighbor CMFD updates
* krylov subspace solver
* adjoint solver

wbinventor:
Added periodic boundary conditions to this list.

wbinventor:
I compiled a list of updates we need to make to the documentation based on our "All Hands" day last week. 

**Installation**
- [ ] Remove deprecated ``--no-numpy`` flag
- [ ] Add Matplotlib as required dependency

**User's Guide**
- [x] Polar quadratures
- [x] Fixed source and criticality calculations
- [x] Moderator rings
- [x] Convergence criteria
- [ ] Change references to `Circle` to `ZCylinder`
- [ ] Changes to `Material` and `openmoc.materialize`
- [ ] Add new capabilities in `openmoc.plotter`
- [ ] Discussion of 3D geometries
- [ ] Adjoint calculations
- [ ] SPH factor calculator
- [ ] FSR volume correction
- [ ] Exponential evaluation options
- [ ] Multi-eigenmode Krylov solver

**Theory & Methodology**
- [ ] *k*-Nearest Neighbor CMFD 
- [ ] Incorporation of nu-scatter into k-eigenvalue
- [ ] SuPerHomogenisation (SPH) factors

**Developer's Guide**
- [x] Writing documentation
- [ ] References

Issue #60, opened 2014-03-08
Doxygen comments to docstrings


wbinventor:
The Doxygen comments should be converted to docstrings. This will be especially helpful for those using IPython to construct OpenMOC input files. SWIG includes the ability to do embed docstrings for the parameters names and types for method calls with the following interface file decorator:

%feature("autodoc")

This works but doesn't capture the content in the Doxygen comments. The 2013 Google Summer of Code included a project to incorporate parse Doxygen to docstrings, but this is still in an as yet unreleased branch of SWIG (http://sourceforge.net/p/swig/code/HEAD/tree/).

A workaround is to use Doxygen to target to XML, and then use the "doxy2swig.py" script to parse the XML to generate a SWIG interface file with the docstrings. The interface file can then be included in the main interface file for wrapping OpenMOC. This method is commonly used all over the place in Python/C++ open source projects and is documented here:

http://www.enricozini.org/2007/tips/swig-doxygen-docstring/

Moving forward, we should include the doxy2swig.py file in our gh-pages branch and generate an interface file every time we rebuild the documentation.

wbinventor:
Closed with #288.

PR #288, opened 2016-03-05
Reaction Rate Tallies, Documentation, Docstrings, PEP8

wbinventor:
This PR adds reaction rate mesh tally feature to the `openmoc.process` module. The goal for this is to make it easy to compare pin-by-pin or assembly-by-assembly reaction rates (nuclide-integrated fission, U-238 capture, etc.) between OpenMOC and other codes (*e.g.*, OpenMC).

In addition, this PR addresses issue #60 by introducing the widely used ["doxy2swig.py"](https://github.com/m7thon/doxy2swig) script to "docs/Doxygen" to convert our Doxygen comments in the C/C++/CUDA source code into docstrings embedded in the SWIG-generated Python module. It is *very* helpful to have docstrings for rapid interactive prototyping in ipython. Similarly, all of the Python sub-modules have been revised to use pure Python docstrings instead of Doxygen comments. 

Finally, the majority of this code changes in this PR are dedicated to new tests for `openmoc.process` and updated documentation on `openmoc.materialize`, `openmoc.plotter` and `openmoc.process` modules to address sub-issues in #194.

All of the changes made in this PR are summarized below:

* Added `Mesh` class to `openmoc.process` module
* Added "doxy2swig.py" to inject Doxygen comments as docstrings in SWIG-generated `openmoc` module
* Revised all Python modules to be (closer to) PEP8 conformant  (*i.e.*, 4 spaces / tab)
* Turned all Python module Doxygen comments into docstrings
* Updated documentation on `openmoc.materialize` module
* Updated docs on `openmoc.plotter` module
* Updated docs on `openmoc.process` module
* Updated docs on `openmoc.log` module
* Renamed `Circle` as `ZCylinder` in the docs
* Removed deprecated ``-no-numpy`` compile time option from docs
* Revised `Options` class to use Python properties instead of getter methods
* Revised error checking in Python modules to use new `openmc.checkvalue` module
* Added test for `openmoc.process` store/restore simulation states
* Added test for `openmoc.process` mesh tallies
* Added test for `openmoc.log` module
* Fixed bug in `Solver::getFSRSource(...)` method
* Removed unsupported `openmoc.compatible.casmo` module


docs/source/usersguide/input.rst
```diff
@@ -172,71 +169,63 @@ OpenMOC uses multi-group macroscopic nuclear cross-sections, provided by the use
...
-   material.setSigmaT(sigma_f)
+   # Load the cross section data into the material
+   material.setSigmaT(sigma_t)
+   material.setSigmaF(sigma_f)
    ...
 
-For many simulations, defining the nuclear data cross-sections by hand in a Python script is cumbersome and error-prone. As a result, OpenMOC includes the ``openmoc.materialize`` module for importing nuclear data cross-sections from an HDF5_ or a Python pickle_ binary file. The ``materialize(...)`` routine is used to import data and instantiate ``Material`` objects returned via a Python dictionary_. The use of the ``openmoc.materialize`` module to import HDF5 and pickle binary files is illusrated in the following snippet:
+For many simulations, defining the nuclear data cross sections by hand in a Python script is cumbersome and error-prone. As a result, OpenMOC includes the ``openmoc.materialize`` module for importing nuclear data cross sections from an HDF5_ binary file. The ``load_from_hdf5(...)`` routine is used to import data and instantiate ``Material`` objects returned via a Python dictionary_. The use of the ``openmoc.materialize`` module to import HDF5 binary files is illusrated in the following snippet:
```
samuelshaner: illusrated -> illustrated

docs/source/usersguide/input.rst
```diff
@@ -172,71 +169,63 @@ OpenMOC uses multi-group macroscopic nuclear cross-sections, provided by the use
...
-    # objects for each material and returns them in a dictionary
-    # indexed by a name string defined in the pickle file
-    pickle_materials = mat.materialize('materials-data.pkl')
-
-    # Retrieve the material called 'fuel' in the pickle file
-    fuel = pickle_materials['fuel']
+The ``openmoc.materialize`` module defines a standard for cross section data stored in binary files. First, HDF5 files must include an ``'# groups'`` attribute with the integer number of groups in the top level of the file hierarchy. Second, the string domain type - ``'material'`` or ``'cell'`` - must be specified in the top level of the file hierarchy. This must match the ``domain_type`` keyword argument passed to ``load_from_hdf5(...)`` which can be either ``'material'`` (default) or ``'cell'``. The ``domain_type`` keyword argument is discussed in more detail at the end of this section. Finally, multi-group cross sections to assign by material or cell must be defined as an `HDF5 group`_ with a string name or integer ID to identify the material or cell. The material group must contain the following floating point `HDF5 datasets`_ of multi-group cross section data:
```
samuelshaner: I feel like `include a ``# groups`` attribute` would be more appropriate.

docs/source/usersguide/input.rst
```diff
@@ -811,7 +790,7 @@ OpenMOC has an integrated CMFD acceleration framework that allows users to great
     cmfd.setOpticallyThick(False)
     cmfd.setSORRelaxationFactor(1.5)
     cmfd.setMOCRelaxationFactor(0.66)
```
samuelshaner: Can you remove this line since we no longer have an MOC relaxation factor?

docs/source/usersguide/input.rst
```diff
@@ -826,7 +805,7 @@ These lines of code should be placed in your input file at the location where th
   * ``setGroupStructure`` (default: same as MOC group structure) - OpenMOC is able to perform CMFD on a coarse energy group structure to allow fine energy group problems to be accelerated with CMFD without incurring a significant computational overhead for CMFD. This function takes a python list as input with the first value of 1 (to indicate the first energy group) followed by an increasing values ending with the number of energy groups plus 1. In the example above, a 7 group MOC problem is broken up into 2 energy groups for CMFD.
   * ``setOpticallyThick`` (default: False) - OpenMOC uses an correction factor on the material diffusion coefficients as described in the Theory and Methodology section. This correction factor is turned off by default.
```
samuelshaner: While you are editing this file can you change `an` to `a`?

docs/source/usersguide/processing.rst
```diff
@@ -103,18 +103,119 @@ The code snippet below illustrates one possible configuration of parameters to t
...
+The ``openmoc.process`` module supports reaction rate through structured Cartesian mesh tallies, similar to how one might compute such quantities in a Monte Carlo code such as MCNP, Serpent or OpenMC. The ``Mesh`` class allows one to tally a reaction rate in each of its structured cells. :ref:`Table 4 <table_mesh_params>` below itemizes each of the property attributes of the ``Mesh`` class which can be used to tally reaction rates.
+
+.. _table_mesh_params:
+
+========================  ===================  ==========  ==========================================================
+Property                  Type                 Default     Note
+========================  ===================  ==========  ==========================================================
+``dimension``             list of integers     None        The number of mesh cells in each direction
+``lower_left``            list of floats       None        The lower-left corner of the structured mesh
+``upper_right``           list of floats       None        The upper-right corner of the structrued mesh
```
samuelshaner: Do we want to go with `lower_left` and `upper_right` or `origin` and `offset` as is done with a lattice?

src/CPUSolver.cpp
```diff
@@ -889,12 +889,12 @@ void CPUSolver::computeFSRFissionRates(double* fission_rates, int num_FSRs) {
     fission_rates[r] = 0.0;
 
   /* Loop over all FSRs and compute the volume-averaged nu-fission rate */
```
samuelshaner: nu-fission -> fission

samuelshaner:
Could the `tally_fission_rates` method just create a `domains_to_coeffs` dict and call the `tally_on_mesh` method? I feel like that would shorten it up a bit.

samuelshaner:
Line 900 of `process.py` looks like it was copied from the doc string for the `tally_fission_rates` method and should be updated for the `tally_on_mesh` method.

samuelshaner:
@wbinventor this PR looks great! I'm really excited to use the new mesh tally features!

docs/source/usersguide/input.rst
wbinventor: good catch

docs/source/usersguide/input.rst
wbinventor: agreed

docs/source/usersguide/input.rst
wbinventor: got it

wbinventor: I also removed a full paragraph description of `Cell:setMOCRelaxationFactor()` below.

docs/source/usersguide/input.rst
wbinventor: got it

docs/source/usersguide/input.rst
samuelshaner: Great, thanks!

docs/source/usersguide/processing.rst
wbinventor: The reason I went with `lower_left` and `upper_right` was to mirror the `openmc.Mesh` class as much as possible. At least from my perspective, this mesh tally feature is intended to make it very easy to compare spatially-integrated reaction rates between the two codes. The way it is implemented now, it is very easy to create an `openmoc.process.Mesh` from an `openmc.mesh`:

```python
...

# Extract mesh for mesh tallies from OpenMC StatePoint
openmc_mesh = statepoint.meshes[mesh_id]

# Create OpenMOC Mesh to mirror OpenMC Mesh
openmoc_mesh = openmoc.process.Mesh()
openmoc_mesh.dimension = openmc_mesh.dimension
openmoc_mesh.lower_left = openmc_mesh.lower_left
openmoc_mesh.upper_right = openmc_mesh.upper_right
openmoc_mesh.width = openmc_mesh.width

...
```

I think we should eitheradopt a standard akin to our own `Lattice` or `openmc.Mesh`. Based on the use case, I would go with the latter since it simplifies the code above (though I recognize that `lower_left` and `upper_right` can easily be backed out of `origin` and `offset` and vice versa).

src/CPUSolver.cpp
wbinventor: Haha, yes indeed. I changed it here as well as in `GPUSolver`

wbinventor:
@samuelshaner I addressed all of your comments. In addition, I added a new parent/ancestor feature to the `Cell` class to facilitate easier comparison of reaction rates between non-discretized OpenMC meshes and discretized OpenMOC meshes.

As I've gone about trying to compare various nuclide-specific reaction rates between OpenMC and OpenMOC, I've realized that it isn't as straightforward as the nuclide-integrated fission rates. The reason is that an OpenMC MGXS `Library` is defined for the `Cells` or `Materials` used to create the OpenMOC `Geometry` *prior* to discretization. Once discretized, the `Cells` and `Materials` in the `Geometry` do not easily map back to those in the `Library`. By tracking a *parent `Cells`* when a `Cell` is cloned during spatial discretization, we can easily keep retain a mapping from pre- to post-discretized geometries.

wbinventor:
I'm unsure why the test suite continues to fail, but will figure this out next.

docs/source/usersguide/processing.rst
samuelshaner: Ok, I'm on board with that.

samuelshaner:
@wbinventor I left two comments in the conversations of this PR that address some items in `process.py` that it appears you might have missed. I wasn't able to have the comment address a particular line because the diff was too large.

samuelshaner:
I'm curious, is there any chance the parent cell could be garbage collected?

wbinventor:
@samuelshaner - I fixed the copy-paste docstring issue. I had considered making `tally_fission_rates` use the more general `tally_on_mesh(...)` method but decided against it for two reasons. First, the process of constructing a `domains_to_coeffs` dictionary actually increases the amount of code in the method overall. Second, we already have the `Solver::computeFSRFissionRates(...)` method implemented in C++/CUDA which is faster than computing each of the rates fsr-by-fsr and group-by-group in Python. This second reason is a minor one since the computation is so fast anyway, but `tally_fission_rates` is noticeably faster than `tally_on_mesh` for some of the problems I've looked at.

wbinventor:
I don't think that the parent `Cell` can be garbage collected. Its `thisown` should be set to `False` once it is assigned to a `Universe`, meaning that we have actually allowed these parent `Cells` to be a memory leak (*i.e.*, the user can still manipulate the parent `Cell`, but neither Python nor C++ will delete its memory). I didn't fix this memory leak here since 1) it's not clear how to do so and 2) it is so little memory that it likely will never matter.

wbinventor:
I think the test suite issues have been fixed.

samuelshaner:
Ok, I think this is about ready to merge. Do you know why `test_simulation_state.py` is failing for Python 3?

wbinventor:
Yeah, I think I just fixed the Python 3 issue. Evidently Python 3 reports more decimal places for `float` variables than does Python 2.

I also re-ran Doxygen and "doxy2swig.py" to reflect the new parent `Cell` feature. @samuelshaner @geogunow - you should keep in mind that all future PRs should only be merged in after running Doxgyen and "doxy2swig.py" to update "openmoc/docstring.i" with docstrings synchronized to the C++ source code. This is akin to having to update the XML schema and statepoint revision for every PR to OpenMC.

PR #290, opened 2016-03-06
Inline ExpEvaluator::computeExponential()

samuelshaner:
This PR inlines the ExpEvaluator::computeExponential() which reduces the segment integration time by ~12%.


wbinventor:
Looks good!

PR #292, opened 2016-03-16
Generalize Option Class

wbinventor:
This short PR generalizes the `openmoc.Options` class to make it sub-classable. This can be nice for users who want to add their own options for certain types of simulations, but leverage the structure already implemented in `openmoc.Options`.

### DIFF
--- a/openmoc/options.py
+++ b/openmoc/options.py
@@ -129,10 +129,6 @@ class Options(object):
         """This method parses command line options and assigns the appropriate
         values to the corresponding class attributes."""
 
-        print(sys.argv[1:])
-        print(self.short_args)
-        print(self.long_args)
-
         try:
             self._opts, self._args = \
                 getopt.getopt(sys.argv[1:], self.short_args, self.long_args)
--- a/openmoc/options.py
+++ b/openmoc/options.py
@@ -10,7 +10,7 @@ else:
     from openmoc.log import py_printf
 
 
-class Options:
+class Options(object):
     """Command-line options for runtime configuration of OpenMOC.
 
     This class parses, interprets and encapsulates the runtime options for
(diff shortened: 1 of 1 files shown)


openmoc/options.py
```diff
@@ -65,30 +75,69 @@ def __init__(self):
...
+    @property
+    def num_omp_threads(self):
+        return self._num_omp_threads
+
+    @property
+    def num_thread_blocks(self):
+        return self._num_thread_blocks
+
+    @property
+    def num_threads_per_block(self):
```
samuelshaner: All other attributes besides `_num_gpu_threads` use their name in the property method getter. What do you think about having this `def num_gpu_threads(self):`?

wbinventor: Oh yes, I agree and will fix it accordingly now.

openmoc/options.py
```diff
@@ -58,37 +58,86 @@ def __init__(self):
...
 
+        self._short_args = 'hfa:s:i:c:t:b:g:r:l:'
+        self._long_args = ['help',
+                           'num-azim=',
+                           'track-spacing=',
+                           'tolerance=',
+                           'max-iters=',
+                           'num-omp-threads=',
+                           'num-thread-blocks=',
+                           'num-gpu-threads=']
```
samuelshaner: can it be changed here as well?

openmoc/options.py
```diff
@@ -146,32 +196,4 @@ def parseArguments(self):
             elif opt in ('-b', '--num-thread-blocks'):
                 self._num_thread_blocks = int(arg)
             elif opt in ('-g', '--num-gpu-threads'):
```
samuelshaner: and here

openmoc/options.py
```diff
@@ -122,14 +171,15 @@ def parseArguments(self):
...
-                num_gpu_threadblocks += 'The number of GPU threadblocks\n'
-                print(num_gpu_threadblocks)
+                num_threadblocks += 'The number of GPU threadblocks\n'
+                print(num_threadblocks)
 
-                num_gpu_threads = '\t{: <35}'.format('-g, --num-gpu-threads=<64>')
-                num_gpu_threads += 'The number of GPU threads per block\n'
-                print(num_gpu_threads)
+                num_threads_per_block = \
+                    '\t{: <35}'.format('-g, --num-gpu-threads=<64>')
```
samuelshaner: and here

openmoc/options.py
wbinventor: agreed, it's standardized here and below now.

samuelshaner:
Looks good. Thanks for another short and easy to review PR @wbinventor!

PR #293, opened 2016-03-18
New Lattice Universe Update Method

wbinventor:
This is a short PR that introduces a new `Lattice::setUniverses(...)` method. This allows a user to assign a `Universe` to an individual `Lattice` cell as a corollary to assigning all `Universes` in an array to a `Lattice`.


wbinventor:
I updated this PR with two new items today:

* Fixed a bug in the deallocation of data from `CPUSolver`
* The "docstring.i" file has been updated from Doxygen

The first item is actually a big deal if one runs 100s-1000s of simulations with different `CPUSolver` instances in the same Python script. Previously the data in a `CPUSolver` wouldn't be deallocated since it did not call the parent `Solver` class' destructor. Now it does so memeory is deallocated as one would expect.

wbinventor:
And one more update - a super tiny tweak to fix a memory leak of `LocalCoords` in the `openmoc.plotter` module. This would not have been noticeable, except for cases like mine where you may generate 100s of plots with high pixel resolution. I think this was a bug I actually fixed sometime in the last year, but probably crept back into the code with my PR to refactor `openmoc.plotter` back in December.

wbinventor:
Wanna take this one up @geogunow? It's actually relatively short.

openmoc/__init__.py
```diff
@@ -32,6 +32,7 @@
 time = (now.month, now.day, now.year, now.hour, now.minute, now.second)
 year_string = '-'.join(map(left_pad, map(str, (now.month, now.day, now.year))))
 today_string = ':'.join(map(left_pad, map(str, (now.hour, now.minute, now.second))))
+today_string += ':' + str(now.microsecond)[:2]
```
samuelshaner: Have you run into cases where this is needed to avoid having multiple runs write to the same log file? If so, do you think there should be some kind of check to see whether a log file with the desired name already exists?

openmoc/docstring.i
```diff
@@ -2821,61 +2802,40 @@ the number of threads per block
...
 computeResidual(residualType res_type) -> double  
 
-Computes the residual between source/flux iterations.  
+Computes the residual between successive flux/source iterations.  
 
 Parameters
 ----------
 * res_type :  
-    the type of residuals to compute (SCALAR_FLUX, FISSION_SOURCE, TOTAL_SOURCE)  
+    the type of residual (FLUX, FISSION_SOURCE, TOTAL_SOURCE)
```
samuelshaner: This should still be `SCALAR_FLUX`

openmoc/docstring.i
```diff
@@ -2821,61 +2802,40 @@ the number of threads per block
...
-    the number of threads per block  
 ";
 
 %feature("docstring") GPUSolver::initializePolarQuadrature "
 initializePolarQuadrature()  
 
-Creates a polar quadrature object for the GPUSolver on the GPU.  
+Initializes a new PolarQuad object.  
+
+Deletes memory old PolarQuad if one was previously allocated.
```
samuelshaner: Maybe and `of` between `memory` and `old`.

openmoc/docstring.i
```diff
@@ -2966,70 +2911,51 @@ the FSR scalar flux
...
 computeKeff()  
 
-Compute $ k_{eff} $ from successive fission sources.  
-
-This method computes the current approximation to the multiplication factor on this
-iteration as follows: $ k_{eff} = \\frac{\\displaystyle\\sum_{i \\in I}
-\\displaystyle\\sum_{g \\in G} \\nu \\Sigma^F_g \\Phi V_{i}} {\\displaystyle\\sum_{i \\in
-I} \\displaystyle\\sum_{g \\in G} (\\Sigma^T_g \\Phi V_{i} - \\Sigma^S_g \\Phi V_{i} -
-L_{i,g})} $  
+Compute $ k_{eff} $ from total fission and absorption rates in each FSR and energy group.
```
samuelshaner: I think this should now just be `..from the total fission rate in each...`

openmoc/docstring.i
```diff
@@ -3509,6 +3381,26 @@ Returns
 the offset of the Cell  
 ";
 
+%feature("docstring") Lattice::updateUniverse "
+updateUniverse(int lat_x, int lat_y, int lat_z, Universe *universe)  
+
+Update the Universe in a particular Lattice cell.  
+
+This method may only be used after an array of Universe has been assigned with the
```
samuelshaner: Can `Universe` be made plural or `objects` be added after it?

src/Universe.cpp
```diff
@@ -1214,6 +1214,41 @@ void Lattice::setUniverses(int num_z, int num_y, int num_x,
 
 
 /**
+ * @brief Update the Universe in a particular Lattice cell.
+ * @details This method may only be used after an array of Universe
```
samuelshaner: I think I should have commented on Universe being plural here instead of in `docstring.i`

samuelshaner:
Glad to see you're pushing the OpenMOC so hard! I think this PR looks good to go as soon as my comments are addressed.

openmoc/__init__.py
wbinventor: That's precisely why I added the sub-second resolution to the log filename here.

openmoc/docstring.i
wbinventor: Weird...it shows up as `SCALAR_FLUX` in "GPUSolver.cpp" but not here. I'll run "doxy2swig.py" again and see if that fixes things.

wbinventor: Oh I see, it is pulling this comment from "Solver.h" where it was indeed incorrect. Good thing to note for the future - "doxy2swig.py" is smart enough to pull comments from parent classes and inject them as docstrings into their children.

openmoc/docstring.i
wbinventor: Got it.

openmoc/docstring.i
wbinventor: Actually this comment is wrong - we compute keff from successive fission sources. I'm not sure how this changed in this PR since I never touched it, but anyway.

openmoc/docstring.i
wbinventor: yep

src/Universe.cpp
wbinventor: No problem, nice to see these things show up in `docstring.i` anyway.

wbinventor:
Alright, I updated this based on your comments @samuelshaner. In addition, I added two more things to this PR:

* A new timer report for the `TrackGenerator`
* Moved the loop over x,y coordinates in `openmoc.plotter` into C++ in `Geometry` to dramatically speed up plotting of high-resolution plots

openmoc/numpy_typemaps.i
```diff
@@ -75,3 +75,9 @@
 
 /* The typemap used to match the method signature for Solver::setFluxes */
 %apply (FP_PRECISION* INPLACE_ARRAY1, int DIM1) {(FP_PRECISION* in_fluxes, int num_fluxes)}
+
+/* The typemap used to match the method signature for
+ * Geometry::getSpatialDataOnGrid */
+%apply (double* IN_ARRAY1, int DIM1) {(double* gridx, int numx)}
```
samuelshaner: Can `numx` be made `num_x` (and same for numy and numxy) here and in `Geometry.cpp` to be consistent with how other methods have `num_x` as inputs?

wbinventor: Done

src/Geometry.cpp
```diff
@@ -1011,6 +1011,114 @@ void Geometry::computeFissionability(Universe* univ) {
...
+ * @param numy the number of y-coordinates in the grid
+ * @param domains a NumPy array of the domain IDs
+ * @param numxy the number of coordinates in the 2D grid
+ * @param zcoord the z-coordinate to use to find the domain IDs
+ * @param domain_type the type of domain ('fsr', 'material', 'cell')
+ */
+void Geometry::getSpatialDataOnGrid(double* gridx, int numx,
+				    double* gridy, int numy,
+				    int* domains, int numxy,
+				    double zcoord, const char* domain_type) {
```
samuelshaner: We do have the ability to make plots in the xz and yz planes with the plotter. Is that functionality preserved? It looks like this is specific to making xy plots.

src/TrackGenerator.cpp
```diff
@@ -2036,3 +2049,38 @@ double TrackGenerator::getPhi(int azim) {
...
     return 2 * M_PI - _phi[azim - _num_azim];
 }
+
+
+/**
+ * @brief Deletes the Timer's timing entries for each timed code section
+ *        code in the loop over track for ray tracing.
+ */
+void TrackGenerator::clearTimerSplits() {
+  _timer->clearSplit("Total time");
```
samuelshaner: Could this be `_timer->clearSplits()` in case further timer splits are added?

wbinventor: Sure, changed it here and in `Solver`

src/Geometry.cpp
wbinventor: As we've determined, 3D plotting is not currently supported by `openmoc.plotter` and it is not entirely trivial to add in.

wbinventor:
I believe this PR is ready for final review / merging. Do note that I made a few updates since the last review:

* performance optimization for `openmoc.process.get_scalar_fluxes(...)` method when retrieving fluxes for *all* FSRs and energy groups
* 4-space indentation in `thisown.i` since the previous 2-space indentation breaks some newer versions of SWIG shipped with Anaconda
* bug fix for `openmoc.process.Mesh.get_mesh_cell_indices(...)` to work for `Meshes` which cover only a subset of the entire `Geometry`
* small change to the way the `plots` directory is created in `openmoc.plotter` to eliminate race conditions if multiple MPI processes are in use with `mpi4py`

@samuelshaner @geogunow can we expedite this merge in the next day or two? I'm trying to sync @logty with a lot of my stuff including this code, and it helps to have everything on the "develop" branches for each repo if possible.

openmoc/numpy_typemaps.i
```diff
@@ -75,3 +75,9 @@
 
 /* The typemap used to match the method signature for Solver::setFluxes */
 %apply (FP_PRECISION* INPLACE_ARRAY1, int DIM1) {(FP_PRECISION* in_fluxes, int num_fluxes)}
+
+/* The typemap used to match the method signature for
+ * Geometry::getSpatialDataOnGrid */
+%apply (double* IN_ARRAY1, int DIM1) {(double* grid_x, int num_x)}
+%apply (double* IN_ARRAY1, int DIM1) {(double* grid_y, int num_y)}
+%apply (int* ARGOUT_ARRAY1, int DIM1) {(int* domains, int num_xy)}
```
geogunow: Why not just use ``std::vector`` ?

openmoc/thisown.i
```diff
@@ -107,6 +107,23 @@
         universes[i][j][k].thisown = False
 %}
 
+/* A Lattice owns the memory for each Universe it contains */
+%pythonappend Lattice::updateUniverse %{
+  # SWIG 3
+  if 'universe' in locals():
+    universe = locals()['universe']
```
geogunow: This confuses me. From my interpretation of what ``locals()`` does, this line should do nothing

src/Geometry.cpp
```diff
@@ -1011,6 +1011,114 @@ void Geometry::computeFissionability(Universe* univ) {
...
+ * @param num_y the number of y-coordinates in the grid
+ * @param domains a NumPy array of the domain IDs
+ * @param num_xy the number of coordinates in the 2D grid
+ * @param zcoord the z-coordinate to use to find the domain IDs
+ * @param domain_type the type of domain ('fsr', 'material', 'cell')
+ */
+void Geometry::getSpatialDataOnGrid(double* grid_x, int num_x,
+				    double* grid_y, int num_y,
+				    int* domains, int num_xy,
+				    double zcoord, const char* domain_type) {
```
geogunow: So yeah, this could be made much more compact with ``std::vector`` and easier to translate with Python lists

src/Material.cpp
```diff
@@ -563,7 +563,7 @@ void Material::setSigmaT(double* xs, int num_groups) {
 
     /* If the cross-section is near zero (e.g., within (-1E-10, 1E-10)) */
     if (fabs(xs[i]) < ZERO_SIGMA_T) {
-      log_printf(INFO, "Overriding zero cross-section in "
+      log_printf(INFO, "Overriding zero tot. MGXS in "
```
geogunow: Why "tot." instead of "total"?

src/Material.cpp
```diff
@@ -586,7 +586,7 @@ void Material::setSigmaTByGroup(double xs, int group) {
 
     /* If the cross-section is near zero (e.g., within (-1E-10, 1E-10)) */
     if (fabs(xs) < ZERO_SIGMA_T) {
-      log_printf(INFO, "Overriding zero cross-section in "
+      log_printf(INFO, "Overriding zero tot. MGXS in "
```
geogunow: ditto

openmoc/numpy_typemaps.i
wbinventor: I'm open to using `std::vector`, but since I don't know how to, I went with NumPy typemaps. Do we have any examples which use `std::vector`? Is this documented in our Developer's Guide as are the NumPy typemaps? If not, I would opt to leave this as is for now in lieu of an overhaul of our NumPy typemaps to Python lists / `std::vectors` for this method as well as many others in the future.

openmoc/thisown.i
wbinventor: This line should extract the `universe` variable from the list of local variables. Basically, it's a way of getting access to the parameter called `locals` for the `Lattice::updateUniverse(...)` method.

src/Geometry.cpp
wbinventor: It would be great to minimize this code since I do recognize this is quite clunky. I'd just rather not go hunting on another SWIG expedition to learn how to use `std::vector`. But if we have some examples in place already then I'll build upon them.

geogunow: If you use std::vector no typemap is needed. You just need to include
std_vector.i  (which we already do) and it will just work.

src/Material.cpp
wbinventor: Yeah so the reason is simply that the message length with "total" exceeds 80 characters and is split across two lines. This can clutter an output log when you have lots (100s-1000s) of `Materials` with lots (10s) of energy groups. If you think "tot." is confusing then I'll switch it to "total" though.

src/Geometry.cpp
wbinventor: Ok then, so what would this method signature look like in that case? And how would one call it from Python?

geogunow: Just like you would in C++. SWIG can automatically translate lists/numpy
arrays to vectors and vice versa.

src/TrackGenerator.cpp
```diff
@@ -2036,3 +2049,38 @@ double TrackGenerator::getPhi(int azim) {
...
+void TrackGenerator::printTimerReport() {
+
+  std::string msg_string;
+
+  log_printf(TITLE, "TRACKGENERATOR TIMING REPORT");
+
+  /* Get the total runtime */
+  double tot_time = _timer->getSplit("Total time");
+  msg_string = "Total time for ray tracing";
+  msg_string.resize(53, '.');
```
samuelshaner: What do you think about having 53 be a constant, like `REPORT_WIDTH`? I'm putting some reports for memory and run parameters in ClosedMOC and am finding that I'm using this value a lot and a constant might be a little cleaner to use throughout.

src/Geometry.cpp
wbinventor: Wow, the `std_vector.i` typemap is fantastic! It was such an easy drop-in replacement, and didn't require nearly the painful wrangling that the NumPy typemaps often incur. We should absolutely start using `std_vector.i` in place of NumPy typemaps everywhere, if we can define our own `UniverseVector`, `CellVector`, `MaterialVector`, etc.

Also @geogunow just to follow up on our conversation from yesterday, the input args and return value are indeed NumPy arrays. Though I presume the typemap may also be compatible with lists.

src/TrackGenerator.cpp
wbinventor: Sure, I just added a new `REPORT_WIDTH` macro to "constants.h" and dropped it into the timer reports in `Solver` and `TrackGenerator'. Perhaps a formalized "report" interface in the `log` module would be useful (and minimize code reuse) for these new timer / memory footprint reports you speak of @samuelshaner.

wbinventor:
@geogunow @samuelshaner this up-to-date with respect to your comments. I'm going to test out the `GPUSolver` and `VectorizedSolver` to ensure that they are still in working order, but otherwise believe this is ready to go.

wbinventor:
The `GPUSolver` looks good but the `VectorizedSolver` is not test-able for an unrelated compilation issue recorded in #300.

wbinventor:
Just a heads up that I tacked on another minor feature in this PR - a new optional `store` parameter to `TrackGenerator::generateTracks()` which can allow one to toggle whether or not to store tracks to a file. I've found that in some cases when multiple OpenMOC processes (*e.g.*, jobs running concurrently on a cluster) attempt to store their tracks to the same file at the same time it may lead to `std::bad_alloc` errors. In these cases, it is beneficial to just dumping the tracks to a file altogether.

wbinventor:
Just a heads up that I updated the SPH factor sample inputs to be compatible with:

1) the latest version of the `openmc.mgxs` module (which added a new attribute to the `MGXS`)
2) the reformatted `openmoc.opencg_compatible` module from #292.

PR #294, opened 2016-03-21
Small bugfix in materialize scatter matrix read-in

jtramm:
A small bug I discovered that @wbinventor was able to fix regarding the scattering matrix materialize read in function.


wbinventor:
Thanks @jtramm!

PR #295, opened 2016-03-26
implemented faster successive over-relaxation iteration scheme for linear solver

samuelshaner:
This PR updates the linearSolve method in `linalg.cpp` to improve the performance of the CMFD solve. The if statement in the inner loop of the SOR iteration was removed by iterating over all the values in the A matrix with the diagonal removed. With these changes, the run time for `c5g7-cmfd.py` goes down by ~33%.

### DIFF
--- a/src/Matrix.cpp
+++ b/src/Matrix.cpp
@@ -7,10 +7,14 @@
  *         a map of lists) to allow for easy setting and incrementing of the
  *         values in the object. When the matrix is needed to perform linear
  *         algebra operations, it is converted to compressed row storage (CSR)
- *         form. The matrix is ordered by cell (as opposed to by group) on the
- *         outside. Locks are used to make the matrix thread-safe against
+ *         form [1]. The matrix is ordered by cell (as opposed to by group) on
+ *         the outside. Locks are used to make the matrix thread-safe against
  *         concurrent writes the same value. One lock locks out multiple rows of
  *         the matrix at a time reprsenting multiple groups in the same cell.
+ *
+ *            [1] "Sparse matrix", Wikipedia,
+ *                https://en.wikipedia.org/wiki/Sparse_matrix.
+ *
  * @param cell_locks Omp locks for atomic cell operations
  * @param num_x The number of cells in the x direction.
  * @param num_y The number of cells in the y direction.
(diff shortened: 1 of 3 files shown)


src/Matrix.cpp
```diff
@@ -283,6 +325,21 @@ FP_PRECISION* Matrix::getA() {
...
 
 /**
+ * @brief Get the A component (excluding the diagonal) of the CSR form of the
+ *        matrix object.
+ * @return A pointer to the A component (excluding the diagonal) of the CSR
+ *         form matrix object.
+ */
+FP_PRECISION* Matrix::getAD() {
+
+  if (_modified)
```
wbinventor: perhaps just add a comment preceding this block to describe what this actually means - *e.g.*, what does `_modified` represent?

src/Matrix.cpp
```diff
@@ -296,6 +353,21 @@ int* Matrix::getIA() {
...
 /**
+ * @brief Get the IA component (excluding the diagonal) of the CSR form of the
+ *        matrix object.
+ * @return A pointer to the IA component (excluding the diagonal) of the CSR
+ *         form matrix object.
+ */
+int* Matrix::getIAD() {
+
+  if (_modified)
+    convertToCSR();
```
wbinventor: ditto

src/Matrix.cpp
```diff
@@ -309,6 +381,21 @@ int* Matrix::getJA() {
...
 
 /**
+ * @brief Get the JA component (excluding the diagonal) of the CSR form of the
+ *        matrix object.
+ * @return A pointer to the JA component (excluding the diagonal) of the CSR
+ *         form matrix object.
+ */
+int* Matrix::getJAD() {
+
+  if (_modified)
```
wbinventor: ditto

src/Matrix.cpp
```diff
@@ -283,6 +325,21 @@ FP_PRECISION* Matrix::getA() {
 
 
 /**
+ * @brief Get the A component (excluding the diagonal) of the CSR form of the
```
wbinventor: Perhaps it might be nice to include a `@details` comment block describing what the `A` component physically represents?

src/Matrix.cpp
```diff
@@ -296,6 +353,21 @@ int* Matrix::getIA() {
 
 
 /**
+ * @brief Get the IA component (excluding the diagonal) of the CSR form of the
```
wbinventor: Perhaps it might be nice to include a `@details` comment block describing what the `IA` component physically represents?

src/Matrix.cpp
```diff
@@ -309,6 +381,21 @@ int* Matrix::getJA() {
 
 
 /**
+ * @brief Get the JA component (excluding the diagonal) of the CSR form of the
```
wbinventor: Perhaps it might be nice to include a `@details` comment block describing what the `JA` component physically represents?

wbinventor:
Looks good to me!

src/Matrix.cpp
samuelshaner: Sure, I added a comment before every `if (_modified)`.

src/Matrix.cpp
samuelshaner: I added more description on the A component and changed the AD component to LU (since it is really the lower + upper components of the matrix). I also added a reference to the wikipedia page where the A, IA, and JA components are described.

wbinventor:
Sweet, can't wait to try out new accelerated acceleration method!

PR #296, opened 2016-03-27
Update to Quadrature

geogunow:
In this PR I update the quadrature used in OpenMOC to better generalize. In particular the following major changes were made:

* The ``PolarQuad`` class has been changed to ``Quadrature`` to imply generalizations outside of just the polar angle
* All weights have been moved to the ``Quadrature`` class. This includes azimuthal weights and azimuthal spacings.
* The ``Quadrature`` class is no longer owned by instances of the ``Solver`` class - but rather instances of the ``TrackGenerator`` class
* Total weights are precomputed inside the ``Quadrature`` class
* The ``TrackGenerator`` has been changed so that all angles in [0, 2*pi] are stored. This improves the readability so that the definition of ``_num_azim`` is consistent. Although this does mean storing duplicate data, there should be no performance hit.

This PR is the first step to syncing OpenMOC with ClosedMOC's 3D-MOC solver.


wbinventor:
Cool stuff @geogunow! I'll look this over later today. The test suite appears to have failed, any idea why? Also, would you mind putting in one or more new tests of the various `Quadrature` options? I don't think we have any tests explicitly devoted to testing that part of our API.

wbinventor:
Also, two other things:

* update @samuelshaner's docs on the old `PolarQuad` to reflect these changes
* run `doxygen` and the "doxy2swig.py" script to update our docstrings in "docstring.i"

geogunow:
@wbinventor @samuelshaner this PR is ready for review

wbinventor:
Great, I'll review this tonight.

src/CPUSolver.cpp
```diff
@@ -748,6 +748,11 @@ void CPUSolver::tallyScalarFlux(segment* curr_segment, int azim_index,
   FP_PRECISION* sigma_t = curr_segment->_material->getSigmaT();
   FP_PRECISION delta_psi, exponential;
 
+  /* Extract weights for all polar angles */
+  FP_PRECISION weights[_num_polar];
```
wbinventor: So we are using allocating `weights` on the stack each and every time `CPUSolver::tallyScalarFlux(...)` is called? Might it not be better to call `_quad->getWeight(...)` directly within the nested loop over energies and groups? If you inline this `Quadrature::getWeight()` I doubt that this would lead to a performance hit, and would reduce the code in this method by a few lines.

src/Cmfd.cpp
```diff
@@ -759,10 +759,10 @@ FP_PRECISION Cmfd::computeLarsensEDCFactor(FP_PRECISION dif_coef,
 
   /* Loop over polar angles */
   for (int p = 0; p < _num_polar; p++) {
-    mu = cos(asin(_polar_quad->getSinTheta(p)));
+    mu = cos(asin(_quadrature->getSinTheta(0,p)));
```
wbinventor: space between arguments `0` and `p`

src/Cmfd.cpp
```diff
@@ -759,10 +759,10 @@ FP_PRECISION Cmfd::computeLarsensEDCFactor(FP_PRECISION dif_coef,
...
     expon = exp(-delta / (3 * dif_coef * mu));
     alpha = (1 + expon) / (1 - expon) - 2 * (3 * dif_coef * mu) / delta;
-    rho += mu * _polar_quad->getWeight(p) * alpha;
+    rho += 2.0 * mu * _quadrature->getPolarWeight(0,p) * alpha;
```
wbinventor: ditto

src/Cmfd.cpp
```diff
@@ -10,7 +10,7 @@
 Cmfd::Cmfd() {
 
   /* Initialize Geometry and Mesh-related attribute */
-  _polar_quad = NULL;
+  _quadrature = NULL;
```
wbinventor: This attribute is called `_quad` in `Solver` and `_quadrature` in `Cmfd`. Perhaps they should use consistent notation since they are one and the same? I personally prefer the more verbose `_quadrature` but it is up to you.

src/Cmfd.cpp
```diff
@@ -1426,12 +1426,12 @@ void Cmfd::setSourceConvergenceThreshold(FP_PRECISION source_thresh) {
...
- * @param polar_quad A PolarQuad object pointer from the Solver
+ * @brief Sets the Quadrature object in use by the MOC Solver.
+ * @param quadrature A Quadrature object pointer from the Solver
  */
-void Cmfd::setPolarQuadrature(PolarQuad* polar_quad) {
-  _polar_quad = polar_quad;
-  _num_polar = polar_quad->getNumPolarAngles();
+void Cmfd::setQuadrature(Quadrature* quadrature) {
+  _quadrature = quadrature;
+  _num_polar = quadrature->getNumPolarAngles() / 2;
```
wbinventor: What is going on here to divide out the number of polar angles?

src/Cmfd.cpp
```diff
@@ -1862,7 +1862,8 @@ void Cmfd::tallyCurrent(segment* curr_segment, FP_PRECISION* track_flux,
         int g = getCmfdGroup(e);
 
         for (int p=0; p < _num_polar; p++)
-          currents[g] += track_flux(p, e) * polar_weights[p] / 2.;
+          currents[g] += track_flux(p, e) * _quadrature->getWeight(azim_index,
```
wbinventor: Why divide the weight by 2.0 here? Can't the `Quadrature` simply store the "correct" weight? Dividing by 2. is a non-negligible operation for each and every `segment` and polar angle.

src/Cmfd.cpp
```diff
@@ -1881,7 +1882,8 @@ void Cmfd::tallyCurrent(segment* curr_segment, FP_PRECISION* track_flux,
...
```
wbinventor: Ditto

src/ExpEvaluator.cpp
```diff
@@ -9,7 +9,7 @@
 ExpEvaluator::ExpEvaluator() {
   _interpolate = true;
   _exp_table = NULL;
-  _polar_quad = NULL;
+  _quad = NULL;
```
wbinventor: Ditto, inconsistent naming convention for the `Quadrature` class attribute between `Solver`, `Cmfd` and `ExpEvaluator`

src/ExpEvaluator.cpp
```diff
@@ -167,7 +167,7 @@ void ExpEvaluator::initialize() {
   log_printf(INFO, "Initializing exponential interpolation table...");
 
   /* Set size of interpolation table */
-  int num_polar = _polar_quad->getNumPolarAngles();
+  int num_polar = _quad->getNumPolarAngles();
```
wbinventor: Here the number of polar angles between `ExpEvaluator` is consistent with `Quadrature`, but this was not the case for `Cmfd` above. I'm sure there is a good reason for this inconsistency, but it should be detailed in some comments.

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+    log_printf(ERROR, "Attempted to retrieve theta for polar angle = %d "
+               "and azim angle = %d but the thetas have not been "
+               "initialized", polar, azim);
+
+  return _thetas[azim][polar];
+}
+
+
+/**
+ * @brief Returns the azimuthal angle value.
```
wbinventor: Add to this comment whether the angle is in units of radians or degrees (radians I believe).

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+ *        polar indexes
+ * @details Angular weights are multiplied by Track spcings
+ * @param azim index of the azimuthal angle of interest
+ * @param polar index of the polar angle of interest
+ * @return the total weight of each Track with the given indexes
+ */
+FP_PRECISION Quadrature::getWeight(int azim, int polar) {
+
+  if (polar < 0 || polar >= _num_polar)
+    log_printf(ERROR, "Attempted to retrieve the multiple for polar angle = "
```
wbinventor: This error message won't make much sense anymore when it refers to "multiple" since this method is now returning the "weight." Perhaps just reconstruct the error messages to align with the method signature.

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+    log_printf(ERROR, "Attempted to retrieve multiple for polar angle = %d "
+               "and azimuthal angle = %d but the multiples have not been "
+               "initialized", polar, azim);
+
+  return _total_weights[azim][polar];
+}
+
+
+/**
+ * @brief Returns a pointer to the Quadrature's array of \f$ sin\theta_{p} \f$.
```
wbinventor: Perhaps just clarify that these are the sines of the polar angles.

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+    log_printf(ERROR, "Attempted to retrieve the thetas array "
+               "but it has not been initialized");
+
+  return _thetas;
+}
+
+
+/**
+ * @brief Returns a pointer to the Quadrature's array of \f$ \phi \f$.
+ * @return a pointer to the array of \f$ \phi \f$
```
wbinventor: Perhaps just clarify that these are the azimuthal angles.

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+               "and azim angle = %d but the sin thetas have not been "
+               "initialized", polar, azim);
+
+  return _sin_thetas[azim][polar];
+}
+
+
+/**
+ * @brief Returns the polar angle value for a given azimuthal and polar
+ *        angle index.
```
wbinventor: Add to this comment whether the angle is in units of radians or degrees (radians I believe).

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+}
+
+
+/**
+ * @brief Set the Quadrature's array of sines of each polar angle.
+ * @details This method is a helper function to allow OpenMOC users to assign
+ *          the Quadrature's sin thetas in Python. A user must initialize a
+ *          NumPy array of the correct size (e.g., a float64 array the length
+ *          of the number of polar angles) as input to this function. This
+ *          function then fills the NumPy array with the data values for the
```
wbinventor: This comment seems to be a bit mis-leading since it talks about how the method "fills the NumPy array with the data values...". In reality, this method takes the NumPy array as input and transfers the values into its own internal array. This was probably a copy-paste mistake of mine in my initial implementation of the `PolarQuad` class.

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+      _thetas[_num_azim/2 - a - 1][_num_polar - p - 1] = M_PI - thetas[ap];
+      ap++;
+    }
+  }
+}
+
+
+/**
+ * @brief Set the Quadrature's array of polar weights
+ * @param weights The polar weights
```
wbinventor: Perhaps include a Python code snippet for this method as is done for `Quadrature::setThetas(...)`?

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  _polar_weights = new FP_PRECISION*[_num_azim/2];
+  for (int i=0; i < _num_azim/2; i++)
+    _polar_weights[i] = new FP_PRECISION[_num_polar];
+
+  /* Extract polar weights from user input */
+  int ap=0;
+  for (int a=0; a < _num_azim/4; a++) {
+    for (int p=0; p < _num_polar/2; p++) {
+      if (weights[ap] < 0. || weights[ap] > M_PI_2)
+        log_printf(ERROR, "Unable to polar weight to %f which is "
```
wbinventor: should this error check instead be in the `Quadrature::setPolarValues(...)`?

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+
+      setPolarValues(_polar_weights, a, p, weights[ap]);
+      ap++;
+    }
+  }
+}
+
+
+/**
+ * @brief Sets the polar angle for the given indexes
```
wbinventor: End comment with period

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+ * @param theta the value of the polar angle to be set
```
wbinventor: Add the units in which theta must be specified

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  _thetas[azim][polar] = theta;
+  _thetas[_num_azim/2 - azim - 1][polar] = theta;
+  _thetas[azim][_num_polar - polar - 1] = M_PI - theta;
+  _thetas[_num_azim/2 - azim - 1][_num_polar - polar - 1] = M_PI - theta;
+}
+
+
+/**
+ * @brief Sets the azimuthal angle for the given index
+ * @param phi the value of the azimuthal angle to be set
```
wbinventor: Add the units in which phi must be specified

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+
+  if (phi <= 0.0 || phi >= M_PI_2)
+    log_printf(ERROR, "Unable to set phi for azim = %d to %f which is not "
+               "in the range (0.0, PI/2)", azim, phi);
+
+  else if (azim >= _num_azim/4)
+    log_printf(ERROR, "Unable to set phi for azim = %d since azim is not in"
+               " the range (0, _num_azim/4)", azim);
+
+  else if (_phis == NULL)
```
wbinventor: Although it doesn't actually matter here, I think it is better to make this memory allocation block a standalone `if` rather than an `else if` since it doesn't logically connect to the error checking blocks above.

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  else if (azim >= _num_azim/4)
+    log_printf(ERROR, "Unable to set polar weight for azim = %d and polar = %d "
+               "since azim is not in the range (0, _num_azim/4)", azim, polar);
+
+  else if (polar >= _num_polar/2)
+    log_printf(ERROR, "Unable to set polar weight for azim = %d and polar = %d "
+               "since polar is not in the range (0, _num_polar/2)", \
+               azim, polar);
+
+  else if (_polar_weights == NULL) {
```
wbinventor: Although it doesn't actually matter here, I think it is better to make this memory allocation block a standalone `if` rather than an `else if` since it doesn't logically connect to the error checking blocks above.

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+
+  if (weight <= 0.0 || weight >= M_PI_2)
+    log_printf(ERROR, "Unable to set azim weight for azim = %d to %f which is "
+               "not in the range (0.0, PI/2)", azim, weight);
+
+  else if (azim >= _num_azim/4)
+    log_printf(ERROR, "Unable to set azim weight for azim = %d since azim is "
+               "not in the range (0, _num_azim/4)", azim);
+
+  else if (_azim_weights == NULL)
```
wbinventor: Although it doesn't actually matter here, I think it is better to make this memory allocation block a standalone `if` rather than an `else if` since it doesn't logically connect to the error checking blocks above.

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  if (spacing <= 0.0)
+    log_printf(ERROR, "Unable to set azimuthal spacing for azim = %d to %f "
+                      "which is not strictly greater than zero", azim,
+                      spacing);
+
+  else if (azim >= _num_azim/4)
+    log_printf(ERROR, "Unable to set azimuthal spacing for azim = %d since "
+                      " azim is not in the range (0, _num_azim/4)", azim);
+
+  else if (_azim_spacings == NULL)
```
wbinventor: Although it doesn't actually matter here, I think it is better to make this memory allocation block a standalone `if` rather than an `else if` since it doesn't logically connect to the error checking blocks above.

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+    _phis = new double[_num_azim/2];
+
+  _phis[azim] = phi;
+  _phis[_num_azim/2 - azim - 1] = M_PI - phi;
+}
+
+
+/**
+ * @brief Sets the azimuthal spacing for the given index
+ * @param spacing the spacing in the azimuthal direction to be set
```
wbinventor: specify the units in for the azimuthal spacing

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+
+  else if (_azim_spacings == NULL)
+    _azim_spacings = new FP_PRECISION[_num_azim/2];
+
+  setAzimuthalValues(_azim_spacings, azim, spacing);
+}
+
+
+/**
+ * @brief Sets the azimuthal weight for the given index
```
wbinventor: End all doxygen comments with a period.

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+    for (int a=0; a < _num_azim/2; a++)
+      _polar_weights[a] = new FP_PRECISION[_num_polar];
+  }
+
+  setPolarValues(_polar_weights, azim, polar, FP_PRECISION(weight));
+}
+
+
+/**
+ * @brief Dummy routine to initialize the polar quadrature.
```
wbinventor: Is this really a "dummy" routine? If this is called by the `Solver` (*not* by the user) then that perhaps should be clarified here in a `@details` comment block.

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+    _phis = new double[_num_azim/2];
+
+  /* Compute a desired azimuthal angles */
+  for (int a = 0; a < _num_azim/2; a++)
+    _phis[a] = 2.0 * M_PI / _num_azim * (0.5 + a);
+}
+
+
+/**
+ * @brief This private routine computes the produce of the sine thetas and
```
wbinventor: typo: "the produce" -> "the product"

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+      for (int p = 0; p < _num_polar/2; p)
+        string << " (" << a << "," << p << "): " << _total_weights[a][p] << ", ";
+    }
+  }
+
+  return string.str();
+}
+
+
+quadratureType Quadrature::getQuadratureType() {
```
wbinventor: No doxygen comment

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+ * @brief Dummy constructor calls the parent constructor.
+ */
+TYPolarQuad::TYPolarQuad(): Quadrature() {
+  _quad_type = TABUCHI_YAMAMOTO;
+}
+
+
+/**
+ * @brief Set the number of polar angles to initialize.
+ * @param num_polar the number of polar angles (maximum 3)
```
wbinventor: The maximum of 3 in the comment is inconsistent with the actual permissible maximum of 6 in the method.

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+ * @brief Routine to initialize the polar quadrature.
+ * @details This routine uses the tabulated values for the Tabuchi-Yamamoto
+ *          polar angle quadrature, including the sine thetas and weights.
+ */
+void TYPolarQuad::initialize() {
+
+  /* Call parent class initialize routine */
+  Quadrature::initialize();
+
+  /* Allocate temporary arrays for tabulated quadrature values */
```
wbinventor: This should have a conditional to check if `thetas` was already initialized or not. If it is not `NULL` then it should first be deallocated and then re-allocated. These sorts of deallocation issues creep up and bite for multi-simulation situations.

wbinventor: Oh nvm, I see that this is a temporary array.

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+ * @brief Dummy constructor calls the parent constructor.
+ */
+LeonardPolarQuad::LeonardPolarQuad(): Quadrature() {
+  _quad_type = LEONARD;
+}
+
+
+/**
+ * @brief Set the number of polar angles to initialize.
+ * @param num_polar the number of polar angles (2 or 3)
```
wbinventor: Again inconsistent max polar angles in the comment with what is actually permitted by the method

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  Quadrature::precomputeWeights(solve_3D);
+}
+
+
+
+/**
+ * @brief Dummy constructor calls the parent constructor.
+ */
+LeonardPolarQuad::LeonardPolarQuad(): Quadrature() {
+  _quad_type = LEONARD;
```
wbinventor: Perhaps this should have a default for `_num_polar`? Otherwise the `LeonardPolarQuad::initialize()` method below will pass without doing anything right?

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+quadratureType Quadrature::getQuadratureType() {
+  return _quad_type;
+}
+
+
+/**
+ * @brief Dummy constructor calls the parent constructor.
+ */
+TYPolarQuad::TYPolarQuad(): Quadrature() {
+  _quad_type = TABUCHI_YAMAMOTO;
```
wbinventor: Perhaps this should have a default for `_num_polar`? Otherwise the `TYPolarQuad::initialize()` method below will pass without doing anything right?

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  Quadrature::precomputeWeights(solve_3D);
+}
+
+
+
+/**
+ * @brief Dummy constructor calls the parent constructor.
+ */
+GLPolarQuad::GLPolarQuad(): Quadrature() {
+  _quad_type = GAUSS_LEGENDRE;
```
wbinventor: Perhaps this should have a default for `_num_polar`? Otherwise the `GLPolarQuad::initialize()` method below will pass without doing anything right?

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+}
+
+
+/**
+ * @brief Set the number of polar angles to initialize.
+ * @param num_polar the number of polar angles (maximum 6)
```
wbinventor: ditto, inconsistent max polar angles in comment wrt to the method implementation.

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+
+  /* Set the arrays of sin thetas and weights */
+  Quadrature::setThetas(thetas, _num_polar/2*_num_azim/4);
+
+  /* Deallocate temporary arrays */
+  delete [] thetas;
+}
+
+
+void GLPolarQuad::precomputeWeights(bool solve_3D) {
```
wbinventor: I don't see `solve_3D` used anywhere - is it a stub for something to come later?

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  delete [] thetas;
+}
+
+
+/**
+ * @brief Calculates total weights for every azimuthal/polar combination based
+ *        on the Leonard polar quadrature
+ * @param solve_3D Boolean indicating whether this is a 3D quadrature
+ */
+void LeonardPolarQuad::precomputeWeights(bool solve_3D) {
```
wbinventor: I don't see `solve_3D` used anywhere - is it a stub for something to come later?

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  Quadrature::precomputeWeights(solve_3D);
+}
+
+
+
+/**
+ * @brief Dummy constructor calls the parent constructor.
+ */
+EqualWeightPolarQuad::EqualWeightPolarQuad(): Quadrature() {
+  _quad_type = EQUAL_WEIGHT;
```
wbinventor: ditto, perhaps use a default number of polar angles here so that `initialize()` won't errantly complete?

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+
+  /* Set the arrays of sin thetas and weights */
+  Quadrature::setThetas(thetas, _num_polar/2*_num_azim/4);
+
+  /* Deallocate temporary arrays */
+  delete [] thetas;
+}
+
+
+void EqualWeightPolarQuad::precomputeWeights(bool solve_3D) {
```
wbinventor: No doxygen comment

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  Quadrature::precomputeWeights(solve_3D);
+}
+
+
+
+/**
+ * @brief Dummy constructor calls the parent constructor.
+ */
+EqualAnglePolarQuad::EqualAnglePolarQuad(): Quadrature() {
+  _quad_type = EQUAL_ANGLE;
```
wbinventor: ditto, default # polar angles?

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  /* Call parent class initialize routine */
+  Quadrature::initialize();
+
+  /* Allocate temporary arrays for tabulated quadrature values */
+  double* thetas = new double[_num_polar/2*_num_azim/4];
+
+  double cos_theta_a, cos_theta_b;
+  double theta_a, theta_b;
+  double delta_theta = M_PI / (_num_polar);
+
```
wbinventor: extra blank line here

src/Quadrature.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  /* Set the arrays of sin thetas and weights */
+  Quadrature::setThetas(thetas, _num_polar/2*_num_azim/4);
+
+  /* Deallocate temporary arrays */
+  delete [] thetas;
+}
+
+
+void EqualAnglePolarQuad::precomputeWeights(bool solve_3D) {
```
wbinventor: No doxygen comment

src/Quadrature.h
```diff
@@ -0,0 +1,217 @@
+/**
+ * @file Quadrature.h
+ * @brief The Quadrature abstract class and subclasses.
+ * @details The Quadrature subclasses are defined with tabulated or functional
+ *          quadrature sets given in the "Lattice Physics Computations",
+ *          Handbook of Nuclear Engineering, Dave Knott, Akio Yamamoto, 2010.
+ * @date April 6, 2015
```
wbinventor: You can update this date and put your name/email here if you wish (or remove it altogether)

src/Quadrature.h
```diff
@@ -0,0 +1,217 @@
...
+  FP_PRECISION** _polar_weights;
+
+  /** An array of the total weights for each azimuthal/polar angle pair */
+  FP_PRECISION** _total_weights;
+
+
+  /* Templates for setting the same values to complimentary or supplimentary
+   * angles */
+  template <typename T>
+  void setPolarValues(T** array, int azim_index, int polar_index, T value) {
```
wbinventor: What is the point of using a template here in place of `FP_PRECISION`? I know this question can be flipped around on its head - e.g., "what is the point of using `FP_PRECISION` in place of fancy templating?" - but I'm not sure I see the point of templating here since there are only two real options for single vs. double precision (as far as I can tell).

src/Quadrature.h
```diff
@@ -0,0 +1,217 @@
...
+    array[azim_index][polar_index] = value;
+    array[_num_azim/2 - azim_index - 1][polar_index] = value;
+    array[azim_index][_num_polar - polar_index - 1] = value;
+    array[_num_azim/2 - azim_index - 1][_num_polar - polar_index - 1] = value;
+  }
+  template <typename T>
```
wbinventor: add blank line between templates

src/Solver.cpp
```diff
@@ -361,40 +337,18 @@ void Solver::setTrackGenerator(TrackGenerator* track_generator) {
                "since the TrackGenerator has not yet generated tracks");
 
   _track_generator = track_generator;
-  _num_azim = _track_generator->getNumAzim() / 2;
+  _num_polar = _track_generator->getQuadrature()->getNumPolarAngles() / 2;
   _num_parallel_track_groups = _track_generator->getNumParallelTrackGroups();
   _tot_num_tracks = _track_generator->getNumTracks();
   _tracks = _track_generator->getTracksByParallelGroup();
+  _quad = track_generator->getQuadrature();
```
wbinventor: perhaps retrieve the `Quadrature` from the `TrackGenerator` into the `_quad` attribute first, then retrieve `_num_polar`? Would have more of a logical flow to it.

src/TrackGenerator.cpp
```diff
@@ -6,19 +6,19 @@
...
-                               const double spacing) {
+TrackGenerator::TrackGenerator(Geometry* geometry, int num_azim,
+                               double spacing) {
 
   setNumThreads(1);
-
   _geometry = geometry;
   setNumAzim(num_azim);
-  setTrackSpacing(spacing);
+  setDesiredAzimSpacing(spacing);
```
wbinventor: As a side note, from a terminology standpoint, are using "azimuthal spacing" in place of "track spacing" now? If so, update the `Options` class to reflect this change.

src/TrackGenerator.cpp
```diff
@@ -233,15 +234,16 @@ Track **TrackGenerator::getTracksByParallelGroup() {
...
  */
-FP_PRECISION* TrackGenerator::getAzimWeights() {
-  if (!_contains_tracks)
-    log_printf(ERROR, "Unable to return Track azimuthal angle quadrature "
-               "weights since Tracks have not yet been generated.");
+Quadrature* TrackGenerator::getQuadrature() {
 
-  return _azim_weights;
+  if (_quad == NULL)
+    log_printf(ERROR, "Unable to return the Solver's Quadrature "
```
wbinventor: Update this error message to use "TrackGenerator" in place of "Solver"

src/TrackGenerator.cpp
```diff
@@ -439,12 +441,12 @@ void TrackGenerator::setNumAzim(int num_azim) {
  * @brief Set the suggested track spacing (cm).
  * @param spacing the suggested track spacing
  */
-void TrackGenerator::setTrackSpacing(double spacing) {
+void TrackGenerator::setDesiredAzimSpacing(double spacing) {
   if (spacing < 0)
     log_printf(ERROR, "Unable to set a negative track spacing %f for the "
```
wbinventor: If we are using "azimuthal spacing" in place of "track spacing" now, then this error message should be changed to reflect that.

src/TrackGenerator.cpp
```diff
@@ -607,17 +633,37 @@ void TrackGenerator::generateTracks(bool neighbor_cells) {
     log_printf(ERROR, "Unable to generate Tracks since no Geometry "
                "has been set for the TrackGenerator");
 
+  /* Check for valid quadrature */
+  if (_quad != NULL) {
+    if (_quad->getNumAzimAngles() != _num_azim) {
+      if (_user_quad) {
+        log_printf(ERROR, "User defined quadrature does not match the "
+                          " number of azimuthal angles in the TrackGenertor");
```
wbinventor: typo: "TrackGenertor" -> "TrackGenerator"

src/TrackGenerator.cpp
```diff
@@ -843,7 +894,7 @@ void TrackGenerator::initializeTracks() {
     for (int j = 0; j < _num_y[i]; j++) {
 
       /* If Track points to the upper right */
-      if (i < _num_azim / 2)
+      if (i < _num_azim / 4)
```
wbinventor: In some places you use spaces when you divide `_num_azim` by 2 or 4, and in other places you do not. Just pick a standard and be consistent throughout (all files, not just this one).

src/TrackGenerator.cpp
```diff
@@ -1357,7 +1409,7 @@ void TrackGenerator::dumpTracksToFile() {
   if (!_contains_tracks)
     log_printf(ERROR, "Unable to dump Tracks to a file since no Tracks have "
       "been generated for %d azimuthal angles and %f track spacing",
```
wbinventor: ditto: "track spacing" -> "azimuthal spacing"

wbinventor: this may be the case in a lot of the error messages btw

src/TrackGenerator.cpp
```diff
@@ -2027,12 +2064,9 @@ double TrackGenerator::getPhi(int azim) {
     log_printf(ERROR, "Unable to get Phi since the TrackGenerator does not"
                " contain tracks.");
 
-  if (azim < 0 || azim >= _num_azim*2)
+  if (azim < 0 || azim >= _num_azim)
     log_printf(ERROR, "Unable to get Phi for azimuthal angle %d since there"
-               " %d azimuthal angles", azim, _num_azim*2);
+               " %d azimuthal angles", azim, _num_azim);
```
wbinventor: looks like we are missing the word "are" in this error message

wbinventor: also, perhaps "Phi" shouldn't be capitalized in these error msgs

wbinventor:
@geogunow I gave this my first thorough review and like how you are generalizing the quadrature. I do think @samuelshaner should also review this since he knows more about where we are headed with the 3D merge than I do. Aside from my comments, a few more things that need to be addressed:

- [ ] Add one or more new tests to the suite for the plug-n-play `Quadratures`
- [ ] Update the docstrings with Doxygen and "doxy2swig.py"
- [ ] Update "thisown.i" since it still refers to `Solver::setPolarQuad(...)`

src/CPUSolver.cpp
geogunow: If I inline the ``Quadrature::getWeight(...)`` function I would probably need to get rid of the error checking conditionals. Do you think I should do that?

src/Cmfd.cpp
geogunow: Well in ``Cmfd`` we always use half the number of polar angles - we never actually use the full number. Maybe I could define a variable like ``_half_num_polar``?

wbinventor:
I just added two more tasks above to test 1) the `GPUSolver` and 2) the SPH factor calculations.

wbinventor:
@geogunow note that if you have issues testing the SPH factor sample inputs, perhaps try using the compatibility fixes I just pushed to #293. I think that PR should be merged in first since it is ready anyway so this shouldn't be an issue by the time you get around to testing out SPH.

src/Quadrature.cpp
geogunow: No. The ``Quadrature::setPolarValues(...)`` is a private method that does not face with the user.

src/Quadrature.cpp
geogunow: Yes, it is

src/Quadrature.cpp
geogunow: ditto

src/Quadrature.h
geogunow: It could also potentially be ``int``

geogunow:
@wbinventor the sph-factor files are not standalone and hence do not work on my computer -- or for anyone doesn't have the pre-requisites that you have.

wbinventor:
Those files never were standalone - you need to install the release
versions of the OpenMC Python API and OpenCG.

geogunow:
@wbinventor If those files need OpenMC and OpenCG should we really be including them in OpenMOC?

samuelshaner:
The documentation for the azim spacing in `docs/source/usersguide/input.rst` should be updated.

samuelshaner:
It looks like the inputs in the profile directory have not been updated. Could you update those?

src/CPUSolver.cpp
```diff
@@ -745,10 +745,10 @@ void CPUSolver::tallyScalarFlux(segment* curr_segment, int azim_index,
 
   /* Compute change in angular flux along segment in this FSR */
   for (int e=0; e < _num_groups; e++) {
-    for (int p=0; p < _num_polar; p++) {
+    for (int p=0; p < _num_polar_2; p++) {
       exponential = _exp_evaluator->computeExponential(sigma_t[e] * length, p);
       delta_psi = (track_flux(p,e)-_reduced_sources(fsr_id,e)) * exponential;
-      fsr_flux[e] += delta_psi * _polar_weights(azim_index,p);
+      fsr_flux[e] += delta_psi * _quadrature->getWeightInline(azim_index, p);
```
samuelshaner: I like how clean this is and that it avoids making a weights pointer attribute in the Solver class!

samuelshaner:
The `tests/test_plot_quadrature/true-0.png` looks like it has 4 quadrature points instead of 3.

src/Quadrature.cpp
```diff
@@ -0,0 +1,1370 @@
...
+ * @code
+ *          polar_weights = numpy.array([0.1, 0.2, 0.05, ... ])
+ *          polar_quad = openmoc.Quadrature()
+ *          polar_quad.setNumAzimAngles(num_azim)
+ *          polar_quad.setNumPolarAngles(len(polar_weights) / num_azim)
+ *          polar_quad.setPolarWeights(polar_weights)
+ * @endcode
+ *
+ * @param weights The polar weights
+ * @param num_azim_times_polar the total number of angles (azimuthal x polar)
```
samuelshaner: Can you note that this is just the total number of angles in one octant?

src/Quadrature.h
```diff
@@ -0,0 +1,236 @@
...
+  FP_PRECISION* _azim_weights;
+
+  /** An array of the quadrature polar weights */
+  FP_PRECISION** _polar_weights;
+
+  /** An array of the total weights for each azimuthal/polar angle pair */
+  FP_PRECISION** _total_weights;
+
+
+  /* Templates for setting the same values to complimentary or supplimentary
```
samuelshaner: complementary and supplementary.

openmoc/plotter.py
```diff
@@ -1235,16 +1235,16 @@ def plot_quadrature(solver, get_figure=False):
 
     # Retrieve data from TrackGenerator
     track_generator = solver.getTrackGenerator()
-    polar_quad = solver.getPolarQuad()
+    quad = track_generator.getQuadrature()
     num_azim = track_generator.getNumAzim()
-    azim_spacing = track_generator.getTrackSpacing()
-    num_polar = polar_quad.getNumPolarAngles()
+    azim_spacing = track_generator.getDesiredAzimSpacing()
+    num_polar = quad.getNumPolarAngles()
```
samuelshaner: This should be `num_polar_2 = quad.getNumPolarAngles() / 2`

samuelshaner: I think that should fix the 4 quadrature points showing up on the test plot.

geogunow:
@samuelshaner I don't think I ever failed the ``tests/test_plot_quadrature`` test. So it must have been that way before.

samuelshaner:
You updated the results for `tests/test_plot_quadrature` in this PR and if you look at the files that were changed in this PR it should be clear that the plot quadrature test results have 4 points, instead of the desired 3 points. Do you see the updated fill in the files changed for this PR?

geogunow:
@samuelshaner you're right. It did change and the new plot looks incorrect. It is corrected now. However, I was right too - I never failed the ``tests/test_plot_quadrature`` test. It seems the test is broken in that it delivers a false positive test result. Maybe @wbinventor knows how to fix this? We might have the same situation for *all* plotting tests.

wbinventor:
@geogunow our plotting tests all use a thresholded comparison of a 3D RGB histogram of the pixels in the reference and generated images ([link](https://github.com/mit-crpg/OpenMOC/blob/develop/tests/testing_harness.py#L346)). This (or something similar) is necessary since the size of the matplotlib images generated on different machines is inconsistent. The threshold used is completely arbitrary right now, and is probably too high to catch just the few pixel difference between the quadrature plots with 3 vs. 4 points. It *should* be large enough, however, to catch blatant discrepancies in plots, but it is certainly not good enough to catch everything. You can always tighten the tolerance for the quadrature test if you want to ensure this doesn't happen again.

samuelshaner:
I just ran the `test_plot_quadrature` locally with the fourth point removed and the distance metric used to judge the difference between two plots only changed slightly from 0.00301 to 0.00306 so I'm not sure tightening the tolerance would work for this case. I'm not really sure of an easy way to compare the plot files and get a clear pass/fail for the test. Perhaps just correct the plot and post an issue about that?

wbinventor:
I don't think the distance metric for our plotting tests needs to be fixed in this PR. But certainly making note of it as an issue as @samuelshaner proposes sounds like reasonable.

geogunow:
@wbinventor @samuelshaner The problem has been fixed for this PR. I just posted an issue about the test suite checking .png images.

wbinventor:
This looks up to par and ready to merge to me.

samuelshaner:
Thanks again for syncing up the quadrature @geogunow!

PR #301, opened 2016-04-06
[HOTFIX] Fixed bug in Lattice::getMinX

ljeure:
Lattice::getMinX had a '+' where it should have had a '*' and wasn't calculating minX properly.


wbinventor:
Thanks for this fix @ljeure!

PR #303, opened 2016-04-10
Improvement to CMFD Coarse Group Structure Setter

wbinventor:
Right now the `Cmfd::setGroupStructure(...)` method may be used to make `Cmfd` use a coarser energy group structure than MOC to accelerate the performance of CMFD. Presently, this coarse group structure must be specified as a monotonically increasing `list` of integers ranging from 1 to the maximum number of energy groups *plus one*. This final point is not intuitive and led to some head scratching for me this morning. 

This PR updates `Cmfd::setGroupStructure(...)` such that the list of group indices starts with 1 and simply ends with the number of energy groups (as I would imagine a novice user would expect). Our example inputs which use `CMFD` have been updated accordingly, as well as our test cases.

In addition, I have tried to come up with a fairly generic scheme to derive coarse group structures for arbitrary energy group structures which grow logarithmically with the number of MOC groups. In addition, this assigns group indices using logarithmic spacing in energy space in order to better capture thermal and resonance effects. Since we don't actually track the energies assigned to each energy group in OpenMOC, I've defined this scheme around the `openmc.mgxs.EnergyGroups` class which encapsulates both energy and groups, as shown below:

```python
# Initialize CMFD using OpenMC Mesh to define the dimensions
cmfd = openmoc.Cmfd()
cmfd.setLatticeStructure(17, 17)
cmfd.setKNearest(3)

# Designate a coarse energy group structure for CMFD
num_groups = max(np.ceil(np.log2(energy_groups.num_groups**2)), 1)
if num_groups < energy_groups.num_groups:
    start = np.log10(max(energy_groups.group_edges[0], 1E-8))
    stop = np.log10(energy_groups.group_edges[-1] * 0.99)
    energies = np.logspace(start, stop, num_groups+1)
    groups = np.unique([energy_groups.get_group(i) for i in energies[::-1]])
    cmfd.setGroupStructure(list(groups))
```

I'd be curious to hear @samuelshaner's thoughts on how this might perform. As a case study, the following is the number of groups derived when used for the CASMO energy group structures:

| # MOC groups | # CMFD groups |
|--------------|---------------|
| 1            | 1             |
| 2            | 2             |
| 4            | 4             |
| 8            | 6             |
| 16           | 8             |
| 25           | 10            |
| 40           | 11            |
| 70           | 13            |


wbinventor:
I also removed the requirement that the coarse group indices must be monotonically increasing. This may be a dangerous move, but seems necessary if one wishes to assign a coarse group structure which is identical to the fine MOC group structure (*i.e.*, [1,1,2] is equivalent to a 2-group MOC structure).

wbinventor:
Actually, restored the monotonic group index check, and simply updated my code snippet for generating CMFD coarse group structures to only assign them if they are indeed coarser than the MOC structure.

samuelshaner:
Thanks for working on a way to clear this up @wbinventor! I know setting the coarse group structure can be a bit confusing. As it stands now, the `_group_indices` attribute must start with `0`, end with `_num_moc_groups`, and be monotonically increasing. With your changes to `Cmfd.cpp`, the input files that set the cmfd group structure with:

    cmfd.setGroupStructure([1,4,7])

will not loop over the highest energy group (i.e. MOC group 0) in `collapseXS()` method since the first value in `_group_indices` will be `1`. I would have expected this to cause the tests to fail.

wbinventor:
I believe that the highest energy group index "0" is hard coded into
`_group_indices is it not?

samuelshaner:
You're right, it is.

wbinventor:
My results are identical with the current results when using group indices
of [1,3,7] for C5G7.

samuelshaner:
I'm just not a big fan of the input to the `setGroupStructure()` method for 2 CMFD groups for a 2 MOC group problem being [1,1,2]. That just seems as confusing to me as the previous approach.

The more I think about the setGroupStructure functionality, the more I think we should just have one option where we allow the user to set the number of CMFD groups to use:

    cmfd.setNumCMFDGroups(2)

If the user doesn't set the number of CMFD groups, it is set to the number of MOC groups. If they set a number of CMFD groups less than the number of MOC groups, the MOC groups are subdivided evenly to the CMFD groups with the highest CMFD group always containing the most MOC groups. For C5G7 this would be:

cmfd_group_structure

It's straightforward for the developer and the user to understand (I think) and accomplishes the goals of accelerating the MOC problem while reducing the time for the CMFD solve. I would guess that the group structure with this approach would be nearly the same as the method you proposed since the MOC energy bins are probably subdivided close to logarithmically already. Also, it could be hard coded into OpenMOC since it doesn't need the information on the energy of the bin edges.

What are your thoughts on just having one option for the user to set the number of CMFD groups?

wbinventor:
This proposal makes lot of sense to me. But I think we should have two
options - with the second one inspired by your diagram. The nested lists
would be much more expressive than what is currently implemented - i.e. ,
[[1], [2]] in place of [1, 1, 2]. This would give users a hook to
manipulate the coarse group structure if they wished. This could be
important since your heuristic to subdivide the group structure without
knowing the energies may not work well in general.
On Apr 10, 2016 12:30 PM, "sam" <notifications@github.com> wrote:

I'm just not a big fan of the input to the setGroupStructure() method for 2
CMFD groups for a 2 MOC group problem being [1,1,2]. That just seems as
confusing to me as the previous approach.

The more I think about the setGroupStructure functionality, the more I
think we should just have one option where we allow the user to set the
number of CMFD groups to use:

cmfd.setNumCMFDGroups(2)

If the user doesn't set the number of CMFD groups, it is set to the number
of MOC groups. If they set a number of CMFD groups less than the number of
MOC groups, the MOC groups are subdivided evenly to the CMFD groups with
the highest CMFD group always containing the most MOC groups. For C5G7 this
would be:

[image: cmfd_group_structure]
<https://cloud.githubusercontent.com/assets/1244537/14411377/62ddfdee-ff14-11e5-99a6-dbbb32efbcc8.png>

It's straightforward for the developer and the user to understand (I think)
and accomplishes the goals of accelerating the MOC problem while reducing
the time for the CMFD solve. I would guess that the group structure with
this approach would be nearly the same as the method you proposed since the
MOC energy bins are probably subdivided close to logarithmically already.
Also, it could be hard coded into OpenMOC since it doesn't need the
information on the energy of the bin edges.

What are your thoughts on just having one option for the user to set the
number of CMFD groups?

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/mit-crpg/OpenMOC/pull/303#issuecomment-208011512>

samuelshaner:
Sounds good to me. Do you want to implement the nested list approach to setting the CMFD group structure?

wbinventor:
Yes, I'll implement the nested lists but leave the heuristic for a later time.

wbinventor:
Ok, so the coarse energy group assignment as a nested list of lists is now implemented. This makes use of the typemaps in `std_vector.i`. The "c5g7-cmfd.py", "c4g7-3d-rodded-B.py" and "mox-assembly.ipynb" example inputs have all been updated, as well as the the tests which make use of coarse CMFD.

src/Cmfd.cpp
```diff
@@ -832,48 +832,50 @@ int Cmfd::getCmfdGroup(int group) {
...
-    if (group_indices[0] != 1)
-      log_printf(ERROR, "The first value in group indices must be 1!");
+  _num_cmfd_groups = group_indices.size();
+  _group_indices = new int[_num_cmfd_groups+1];
 
-    /* Set first group indice to 0 */
-    _group_indices[0] = 0;
+  if (group_indices[0][0] != 1)
+    log_printf(WARNING, "The first CMFD coarse group index is %d but is being "
+	       "over-ridden to 1", group_indices[0][0], group_indices[0][0]);
```
samuelshaner: you can remove the second `group_indices[0][0]`.

src/Cmfd.cpp
```diff
@@ -832,48 +832,50 @@ int Cmfd::getCmfdGroup(int group) {
...
+  /* Set first group index to 0 */
+  _group_indices[0] = 0;
 
-      _group_indices[i] = group_indices[i] - 1;
-      log_printf(DEBUG, "group indices %d: %d", i, group_indices[i]);
-    }
+  /* Set MOC group bounds for rest of CMFD energy groups */
+  for (int i=1; i < _num_cmfd_groups; i++) {
+    _group_indices[i] = group_indices[i][0] - 1;
+    log_printf(DEBUG, "group indices %d: %d", i, _group_indices[i]);
```
samuelshaner: Could we have a check to make sure the group indices are monotonically increasing? Most users will probably know very little about CMFD and this is one of the few user inputs, so I think it would be nice to have an error check.

src/Cmfd.cpp
```diff
@@ -832,48 +832,50 @@ int Cmfd::getCmfdGroup(int group) {
...
   }
 
-  _user_group_indices = true;
+  /* Set the last group index */
+  _group_indices[_num_cmfd_groups] =
+    group_indices[_num_cmfd_groups-1].back();
+  log_printf(DEBUG, "group indices %d: %d",
+	     _num_cmfd_groups, _group_indices[_num_cmfd_groups]);
```
samuelshaner: Could we also check that the last index is equal to `_num_cmfd_groups + 1`?

src/Cmfd.cpp
```diff
@@ -952,8 +954,19 @@ void Cmfd::initializeGroupMap() {
...
-    setGroupStructure(NULL, _num_moc_groups+1);
-    _user_group_indices = false;
+    _num_cmfd_groups = _num_moc_groups;
+
+    /* Delete old group indices array if it exists */
+    if (_group_indices != NULL)
+      delete [] _group_indices;
+
+    /* Allocate memory for new group indices */
+    _group_indices = new int[_num_cmfd_groups];
```
samuelshaner: This should be `_num_cmfd_groups + 1`

src/Cmfd.cpp
samuelshaner: Or rather `_num_cmfd_groups`

samuelshaner:
This looks great! As soon as the items are addressed I'll do a little testing and then it should be ready to merge. Implementing the general method should be a fairly small PR and I'll make that after this one is merged and update the documentation on CMFD.

wbinventor:
Alright, just for the record for posterity, the following is an updated version of the code snippet I provided above to auto-generate a logarithmically spaced coarse CMFD group structure. It's a bit hacky now but at least it works:

```python
# Designate a coarse energy group structure for CMFD
num_groups = int(max(np.ceil(np.log2(energy_groups.num_groups**2)), 1))
if num_groups < energy_groups.num_groups:
    start = np.log10(max(energy_groups.group_edges[0], 1E-8))
    stop = np.log10(energy_groups.group_edges[-1] * 0.999999)
    energies = np.logspace(start, stop, num_groups+1)[::-1]

    groups = []
    for i, energy in enumerate(energies[:-1:]):
        lower = energy_groups.get_group(energies[i])
        upper = energy_groups.get_group(energies[i+1])

        if i == len(energies)-1:
            upper = num_fine_groups + 1
        if i > 0:
            if upper in groups[-1]:
                continue
            if lower in groups[-1]:
                lower = groups[-1][-1] + 1

        groups.append(list(np.arange(lower, upper + 1)))

    cmfd.setGroupStructure(groups)
```

src/Cmfd.cpp
wbinventor: Well the problem is that neither `_num_cmfd_groups` nor `_num_moc_groups` is known when this method is called. Both are equal to zero from the constructor. So there isn't a way to properly do the error checking, at least not in this method.

wbinventor:
Ok, this is up-to-date per your comments @samuelshaner. I do think the figure you provided should absolutely be included in the updated User's Guide once it is updated to reflect the latest and greatest CMFD features. It is very helpful.

src/Cmfd.cpp
```diff
@@ -832,48 +832,53 @@ int Cmfd::getCmfdGroup(int group) {
...
+  if (group_indices[0][0] != 1)
+    log_printf(WARNING, "The first CMFD coarse group index is %d but "
+	       "is being over-ridden to 1", group_indices[0][0]);
+
+  /* Set first group index to 0 */
+  _group_indices[0] = 0;
+
+  /* Set MOC group bounds for rest of CMFD energy groups */
+  for (int i=1; i < _num_cmfd_groups; i++) {
+    if (group_indices[i][0] <= _group_indices[i-1])
```
samuelshaner: I just tried a case with repeated MOC groups in two different CMFD groups and this didn't pick it up. I think this should include another loop over the MOC groups within a CMFD group. You might also use a `last_moc_group` variable to not the last group that was iterated over (since it might be in the previous CMFD group).

src/Cmfd.cpp
samuelshaner: Good point. What do you think about adding a check in the `Cmfd::initializeGroupMap()` method?

src/Cmfd.cpp
wbinventor: Done.

src/Cmfd.cpp
```diff
@@ -832,48 +832,57 @@ int Cmfd::getCmfdGroup(int group) {
...
+
+  /* Set MOC group bounds for rest of CMFD energy groups */
+  for (int i=1; i < _num_cmfd_groups; i++) {
+    for (int j=0; j < group_indices[i].size(); j++) {
+      if (group_indices[i][j] <= last_moc_group)
+	log_printf(ERROR, "The CMFD coarse group indices are not "
+		   "monotonically increasing");
     }
+    _group_indices[i] = group_indices[i][0] - 1;
+    last_moc_group = _group_indices[i];
```
samuelshaner: I think this should be moved in the last if statement.

src/Cmfd.cpp
```diff
@@ -832,48 +832,57 @@ int Cmfd::getCmfdGroup(int group) {
...
+  if (group_indices[0][0] != 1)
+    log_printf(WARNING, "The first CMFD coarse group index is %d but "
+	       "is being over-ridden to 1", group_indices[0][0]);
+
+  /* Set first group index to 0 */
+  _group_indices[0] = 0;
+  int last_moc_group = group_indices[0][group_indices[0].size()];
+
+  /* Set MOC group bounds for rest of CMFD energy groups */
+  for (int i=1; i < _num_cmfd_groups; i++) {
```
samuelshaner: Can you start iterating at 0 to check to MOC groups in the first CMFD group and then remove line 866?

src/Cmfd.cpp
wbinventor: I'm unclear what you mean by this?

src/Cmfd.cpp
wbinventor: Sure, got it

src/Cmfd.cpp
samuelshaner: I was thinking the `last_moc_group` should actually be set to `group_indices[i][j]` (i.e. an MOC group in the 2D vector that the user input) and then as the `group_indices` 2D vector is looped over, the `if (group_indices[i][j] <= last_moc_group)` will make sure all the MOC groups in the `group_indices` 2D vector are monotonically increasing. In full:

    /* Set MOC group bounds for rest of CMFD energy groups */
    for (int i=0; i < _num_cmfd_groups; i++) {
        for (int j=0; j < group_indices[i].size(); j++) {
            if (group_indices[i][j] <= last_moc_group)
                log_printf(ERROR, "The CMFD coarse group indices are not "
                    "monotonically increasing");
            last_moc_group = group_indices[i][j];
        }
        _group_indices[i] = group_indices[i][0] - 1;
        log_printf(DEBUG, "CMFD group indices %d: %d", i, _group_indices[i]);
    }

wbinventor: Got it

samuelshaner:
Thanks for this PR and your patience in dealing with my meticulous comments @wbinventor! As soon as the test suite (presumably) passes, I'll merge this in.

wbinventor:
I have no idea why Python 3 CMFD tests are failing here as they all pass on my machine. I'll dig into it at some point tomorrow.

wbinventor:
I believe that the Python 3 test suite failures have been fixed. I'm surprised it was passing with Python 2, since I discovered a true bug in the underlying C++ code. I *think* this discrepancy must be the result of Python 2 and 3 initializing lists differently - in particular, with different max lengths and default values - which permitted them to be input into `Cmfd::setGroupStructure(...)` with different outcomes. Anyway, this is just something to be aware of as we move to using SWIG typemaps from `std_vector.i` in place of NumPy typemaps since they may be less forgiving of incorrect C++ code, resulting in different (errant) outcomes in Python 2 and 3.

samuelshaner:
Thanks for resolving those issues @wbinventor! Just tested `Cmfd::setGroupStructure()` out and everything works great!

PR #305, opened 2016-04-14
Updated FSR Volumes to be Stored in TrackGenerator

geogunow:
This PR changes the FSR volumes calculation to store the computed FSR volumes within the ``TrackGenerator`` attribute ``_FSR_volumes``. Previously, when ``getFSRVolumes`` was called it would allocate a new array of FSR volumes and return that array, leaving deallocation to be done outside the function and potentially outside the class. This update simplifies the memory management so that only the ``TrackGenerator`` allocates and deallocates memory associated with ``_FSR_volumes``. When external functions call ``getFSRVolumes`` to query the FSR volumes, a pointer to ``_FSR_volumes`` is returned.


wbinventor:
@geogunow this seems to be picking up changes from your quadrature PR #296 as well as the refactored volume calculation - can you sync this up with the develop branch so GitHub won't be confused and cloud the diff with changes from both sources?

PR #306, opened 2016-04-15
Updated FSR Volumes to be Stored in TrackGenerator

geogunow:
This PR changes the FSR volumes calculation to store the computed FSR volumes within the ``TrackGenerator`` attribute ``_FSR_volumes``. Previously, when ``getFSRVolumes`` was called it would allocate a new array of FSR volumes and return that array, leaving deallocation to be done outside the function and potentially outside the class. This update simplifies the memory management so that only the ``TrackGenerator`` allocates and deallocates memory associated with ``_FSR_volumes``. When external functions call ``getFSRVolumes`` to query the FSR volumes, a pointer to ``_FSR_volumes`` is returned.

### DIFF
--- a/src/Solver.cpp
+++ b/src/Solver.cpp
@@ -53,9 +53,6 @@ Solver::Solver(TrackGenerator* track_generator) {
  */
 Solver::~Solver() {
 
-  if (_FSR_volumes != NULL)
-    delete [] _FSR_volumes;
-
   if (_FSR_materials != NULL)
     delete [] _FSR_materials;
 
@@ -503,10 +500,7 @@ void Solver::initializeFSRs() {
 
   log_printf(INFO, "Initializing flat source regions...");
 
-  /* Delete old FSR arrays if they exist */
-  if (_FSR_volumes != NULL)
-    delete [] _FSR_volumes;
-
+  /* Delete old FSR materials array if it exists */
   if (_FSR_materials != NULL)
     delete [] _FSR_materials;
 
--- a/src/Solver.cpp
+++ b/src/Solver.cpp
@@ -119,7 +119,7 @@ FP_PRECISION Solver::getFSRVolume(int fsr_id) {
     log_printf(ERROR, "Unable to get the volume for FSR %d since the FSR "
                "IDs lie in the range (0, %d)", fsr_id, _num_FSRs);
 
-  else if (_FSR_volumes == NULL)
+  if (_FSR_volumes == NULL)
     log_printf(ERROR, "Unable to get the volume for FSR %d since the FSR "
                "volumes have not yet been computed", fsr_id);
 
@@ -511,6 +511,7 @@ void Solver::initializeFSRs() {
   _polar_times_groups = _num_groups * _num_polar_2;
 
   /* Get an array of volumes indexed by FSR  */
+  _track_generator->resetFSRVolumes();
   _FSR_volumes = _track_generator->getFSRVolumes();
 
   /* Generate the FSR centroids */
--- a/src/TrackGenerator.cpp
+++ b/src/TrackGenerator.cpp
@@ -19,7 +19,6 @@ TrackGenerator::TrackGenerator(Geometry* geometry, int num_azim,
   _use_input_file = false;
   _tracks_filename = "";
   _z_coord = 0.0;
-  _max_optical_length = std::numeric_limits<FP_PRECISION>::max();
   _FSR_volumes = NULL;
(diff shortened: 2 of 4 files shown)


src/TrackGenerator.cpp
```diff
@@ -268,9 +273,27 @@ double TrackGenerator::getZCoord() {
...
+  return _FSR_volumes;
+}
+
+
+/**
  * @brief Computes and returns an array of volumes indexed by FSR.
- * @details Note: It is the function caller's responsibility to deallocate
- *          the memory reserved for the FSR volume array.
+ * @details Note: The memory is stored in the FSR volumes buffer of the
+ *          TrackGenerator and is freed during deconstruction.
```
wbinventor: perhaps "memory deallocation" is more intuitive to users who are non-C/C++ developers?

src/TrackGenerator.cpp
```diff
@@ -268,9 +273,27 @@ double TrackGenerator::getZCoord() {
 
 
 /**
+ * @brief Return the array used to store the FSR volumes
+ * @return _FSR_volumes the FSR volumes array indexed by FSR ID
+ */
+FP_PRECISION* TrackGenerator::getFSRVolumesBuffer() {
+#pragma omp critical
```
wbinventor: Is this critical since the number of FSRs may be changing?

src/TrackGenerator.cpp
```diff
@@ -268,9 +273,27 @@ double TrackGenerator::getZCoord() {
 
 
 /**
+ * @brief Return the array used to store the FSR volumes
+ * @return _FSR_volumes the FSR volumes array indexed by FSR ID
+ */
+FP_PRECISION* TrackGenerator::getFSRVolumesBuffer() {
```
wbinventor: I'm not really sure I understand the point of abstracting this code away from `TrackGenerator::getFSRVolumes()`. What was your reasoning?

src/Solver.cpp
```diff
@@ -53,9 +53,6 @@ Solver::Solver(TrackGenerator* track_generator) {
  */
 Solver::~Solver() {
 
-  if (_FSR_volumes != NULL)
```
wbinventor: Does the `Solver` never actually use the FSR volumes? Or does it only retrieve the array from the `TrackGenerator` into temporary variables now?

src/TrackGenerator.cpp
geogunow: Maybe I should just take out the comment altogether. We don't mention how the memory is managed for other variables.

src/TrackGenerator.cpp
geogunow: It's critical because you don't want multiple threads to be allocating a new FSR volumes array. This function should be called infrequently so it shouldn't have any effect on performance.

src/TrackGenerator.cpp
geogunow: We don't want to allocate all new FSR volumes each time we call ``getFSRVolumes``. Also sometimes we might not want to recalculate all the FSR volumes, but rather just get the version previously calculated.

src/Solver.cpp
geogunow: It only retrieves the array from the ``TrackGenerator``

src/TrackGenerator.cpp
samuelshaner: I vote for keeping the comment, but changing it to `memory deallocation` as I had to look around to see how memory was handled for FSR volumes when you made this change in ClosedMOC.

src/TrackGenerator.cpp
wbinventor: Is it true that this is the only place in the code where we have a special routine to allocate a memory buffer? For better or worse, every array that we allocate is done inline within the routine that initializes the array. For consistency I would say simply do the same here, but can be convinced otherwise if there is a good reason for this allocation to be in its own method.

src/TrackGenerator.cpp
wbinventor: makes sense

src/Solver.cpp
wbinventor: cool

wbinventor:
This looks good to me aside from my comment on the memory allocation. Two more things to do before a full release:

- [ ] Make sure that the `GPUSolver` still works
- [ ] Make sure that the SPH inputs still work
- [ ] Run Doxygen and "doxy2swig.py" to update our docstrings

PR #307, opened 2016-04-21
Traverse Tracks Abstraction

geogunow:
This PR builds off #306 and abstracts the looping scheme for iterating over all long characteristics (ie ``Tracks``) and all short characteristics (ie ``segments``). This allows OpenMOC to be far more flexible as the algorithms can be agnostic of the track laydown and ray tracing scheme. This is particularly useful for extension to 3D MOC as the algorithms can be made largely agnostic of the dimensionality. It is also incredible useful for implementing on-the-fly ray tracing schemes as the fundamental algorithms can remain unchanged.

Specifically, the ``TraverseTracks`` class describes how to loop over ``Tracks`` and ``segments``. Then in ``TrackTraversingAlgorithms.h``, subclasses of ``TraverseTracks`` are created that implement a simple algorithm, such as calculating FSR volumes or doing a transport sweep. 

These subclasses only need to:
* define a constructor
* define an ``onTrack(...)`` function which describes what to do on each ``Track``
* define an ``execute`` function which describes the skeleton of the algorithm
* define a destructor (only if memory is allocated)

The ``execute`` function should include preprocessing steps (if any) and specify which ``MOCKernels`` to use in the algorithm. ``MOCKernels`` describe what to do on each ``segment`` as it is first encountered in an iteration of an algorithm. Though it is possible to apply functionality on ``segments`` in the ``onTrack(...)`` function, ``MOCKernels`` apply the functionality right away. The distinction is not apparent for schemes that have explicitly formed segments. However, for on-the-fly methods that do not store segments, the ``MOCKernels`` apply their functionality as soon as the segment is calculate whereas the ``onTrack(...)`` function is applied after all ``segments`` in the ``Track`` have been traversed. In the case of on-the-fly methods, correct ``segment`` information might not be available at the time ``onTrack(...)`` is called.

After the preprocessing steps and the declaration of the ``MOCKernels`` (can be ``NULL``) the algorithm should  call either ``loopOverTracks`` or ``loopOverTracksByParallelGroup`` to execute the functionality of described in the ``onTrack(...)`` function on ``Tracks`` and in the ``MOCKernels`` on ``segments``.

The goal of this PR is to introduce the abstracted looping scheme and eventually use it for every loop over ``Tracks`` and ``segments`` (as is done in ClosedMOC). However, to reduce to size of this initial PR, I have limited the algorithms for which the abstract looping is applied to the volume calculation and the transport sweep.


geogunow:
@friedmud and @jtramm might be interested in this PR

friedmud:
This looks really good @geogunow!

After Physor I'll have to show you the structure I ultimately ended up with in my new MOC code.  It is similar to what you have here, but, since I only do on the fly, it's a little simpler.  Still would be good to compare and contrast!

openmoc/openmoc.i
```diff
@@ -19,6 +19,7 @@
   #include "../src/LocalCoords.h"
   #include "../src/log.h"
   #include "../src/Material.h"
+  #include "../src/MOCKernel.h"
```
wbinventor: Do we need to wrap these new files with SWIG? Or should these algorithms all be under-the-hood and opaque to Python?

src/CPUSolver.cpp
```diff
@@ -648,75 +648,17 @@ void CPUSolver::transportSweep() {
...
-      }
-    }
-  }
+  /* Initialize flux in each FSR to zero */
+  flattenFSRFluxes(0.0);
 
-  return;
+  /* Tracks are traversed and the MOC equations from this CPUSolver are applied
+     to all Tracks and corresponding segments */
+  TransportSweep sweep_tracks(_track_generator);
```
wbinventor: WOW.

src/CPUSolver.h
```diff
@@ -74,8 +78,7 @@ class CPUSolver : public Solver {
   virtual void transferBoundaryFlux(int track_id, int azim_index,
                                     bool direction, FP_PRECISION* track_flux);
 
-public:
-  CPUSolver(TrackGenerator* track_generator=NULL);
+
```
wbinventor: Remove blank lines

src/CPUSolver.h
```diff
@@ -44,6 +45,9 @@ class CPUSolver : public Solver {
   /** OpenMP mutual exclusion locks for atomic FSR scalar flux updates */
   omp_lock_t* _FSR_locks;
 
+public:
+  CPUSolver(TrackGenerator* track_generator=NULL);
```
wbinventor: We already have this constructor - any idea why it is showing up as an addition in this diff?

src/MOCKernel.cpp
```diff
@@ -0,0 +1,165 @@
+#include "MOCKernel.h"
+#include "TrackGenerator.h"
+
+/**
+ * @brief Constructor for the MOCKernel assigns default values
+ * @param track_generator the TrackGenerator used to pull relevant tracking
+ *        data from
```
wbinventor: This `@param` comment is incomplete

src/MOCKernel.cpp
```diff
@@ -0,0 +1,165 @@
+#include "MOCKernel.h"
+#include "TrackGenerator.h"
+
+/**
+ * @brief Constructor for the MOCKernel assigns default values
```
wbinventor: The text "assigns default values" in the `@brief` comment block isn't very informative unless you expand upon this in a `@details` block (*i.e.*, a maximum optical path length is assigned).

src/MOCKernel.cpp
```diff
@@ -0,0 +1,165 @@
...
+  _max_tau = track_generator->retrieveMaxOpticalLength();
+}
+
+
+/**
+ * @brief Constructor for the VolumeKernel assigns default values, calls
+ *        the MOCKernel constructor, and pulls refernces to FSR locks and FSR
+ *        volumes from the provided TrackGenerator.
+ * @param track_generator the TrackGenerator used to pull relevant tracking
+ *        data from
```
wbinventor: ditto, incomplete `@param` comment

src/MOCKernel.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * @brief Constructor for the CounterKernel assigns default values and calls
+ *        the MOCKernel constructor
+ * @param track_generator the TrackGenerator used to pull relevant tracking
+ *        data from
+ * @param row_num the row index into the temporary segments matrix
+ */
+CounterKernel::CounterKernel(TrackGenerator* track_generator, int row_num) :
+                           MOCKernel(track_generator, row_num) {}
```
wbinventor: Add two spaces to make this continuation line match with the args in the line above.

src/MOCKernel.cpp
```diff
@@ -0,0 +1,165 @@
...
+  _quadrature = track_generator->getQuadrature();
+  _weight = 0;
+}
+
+
+/**
+ * @brief Constructor for the CounterKernel assigns default values and calls
+ *        the MOCKernel constructor
+ * @param track_generator the TrackGenerator used to pull relevant tracking
+ *        data from
```
wbinventor: ditto, incomplete `@param` block

src/MOCKernel.cpp
```diff
@@ -0,0 +1,165 @@
...
+ *        volumes from the provided TrackGenerator.
+ * @param track_generator the TrackGenerator used to pull relevant tracking
+ *        data from
+ * @param row_num the row index into the temporary segments matrix
+ */
+VolumeKernel::VolumeKernel(TrackGenerator* track_generator, int row_num) :
+                           MOCKernel(track_generator, row_num) {
+  _FSR_locks = track_generator->getFSRLocks();
+
+  if (_FSR_locks == NULL)
```
wbinventor: Is it ever possible for this error to occur? And if it does occur, is it the user's fault?

src/MOCKernel.cpp
```diff
@@ -0,0 +1,165 @@
...
+#include "TrackGenerator.h"
+
+/**
+ * @brief Constructor for the MOCKernel assigns default values
+ * @param track_generator the TrackGenerator used to pull relevant tracking
+ *        data from
+ * @param row_num the row index into the temporary segments matrix
+ */
+MOCKernel::MOCKernel(TrackGenerator* track_generator, int row_num) {
+  _count = 0;
```
wbinventor: Where is `row_num` used here?

src/MOCKernel.cpp
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * @brief Constructor for the VolumeKernel assigns default values, calls
+ *        the MOCKernel constructor, and pulls refernces to FSR locks and FSR
+ *        volumes from the provided TrackGenerator.
+ * @param track_generator the TrackGenerator used to pull relevant tracking
+ *        data from
+ * @param row_num the row index into the temporary segments matrix
+ */
+VolumeKernel::VolumeKernel(TrackGenerator* track_generator, int row_num) :
+                           MOCKernel(track_generator, row_num) {
```
wbinventor: Ditto, where is `row_num` used here?

src/MOCKernel.cpp
```diff
@@ -0,0 +1,165 @@
...
+  _FSR_locks = track_generator->getFSRLocks();
+
+  if (_FSR_locks == NULL)
+    log_printf(ERROR, "Unable to create a VolumeKernel without first creating "
+               "FSR locks");
+
+  _FSR_volumes = track_generator->getFSRVolumes();
+  _quadrature = track_generator->getQuadrature();
+  _weight = 0;
```
wbinventor: Perhaps add a comment above `weight` indicating what it represents?

src/MOCKernel.cpp
wbinventor: Perhaps add a comment above `count` indicating what it represents?

wbinventor: You might even consider renaming `_count` since it isn't obvious what it represents

src/MOCKernel.cpp
```diff
@@ -0,0 +1,165 @@
...
+ */
+MOCKernel::~MOCKernel() {};
+
+
+/*
+ * @brief Reads and returns the current count
+ * @details MOC kernels count how many times they are accessed. This value
+ *          returns the value of the counter (number of execute accesses)
+ *          since kernel creation or last call to newTrack.
+ * @return _count the counter value
```
wbinventor: Remove `_count` from comment

src/MOCKernel.cpp
```diff
@@ -0,0 +1,165 @@
...
+int MOCKernel::getCount() {
+  return _count;
+}
+
+
+/* @brief Resets the maximum optcal path length for a segment
+ * @details MOC kernels ensure that there are no segments with an optical path
+ *          length greater than the maximum optical path length by splitting
+ *          then when they get too large.
+ * @param the maximum optical path length for a segment
```
wbinventor: Add `max_tau` to beginning of `@param` comment block

src/MOCKernel.cpp
```diff
@@ -0,0 +1,165 @@
...
+  FP_PRECISION* sigma_t = mat->getSigmaT();
+  double max_sigma_t = 0;
+  for (int e=0; e < mat->getNumEnergyGroups(); e++)
+    if (sigma_t[e] > max_sigma_t)
+      max_sigma_t = sigma_t[e];
+
+  int num_cuts = std::max((int) std::ceil(length * max_sigma_t / _max_tau), 1);
+
+  /* Increment count */
+  _count += num_cuts;
```
wbinventor: It looks like `_count` is used for various different things in the different kernels? Above you define it in a comment as the number of times `execute` is accessed, but here it is obviously not that and is instead the number of segment cuts.

src/MOCKernel.cpp
```diff
@@ -0,0 +1,165 @@
...
+ * @param the maximum optical path length for a segment
+ */
+void MOCKernel::setMaxOpticalLength(FP_PRECISION max_tau) {
+  _max_tau = max_tau;
+}
+
+
+/*
+ * @brief Adds segment contribution to the FSR volume
+ * @details The VolumeKernel execute function adds the product of the
```
wbinventor: Add to the `@details` comment that this also cuts segments with lengths greater than the maximum permitted optical path length.

jtramm:
This looks awesome - really looking forward to adding in a random ray option to OpenMOC!

openmoc/openmoc.i
geogunow: Under-the-hood is fine

src/CPUSolver.h
geogunow: Yeah, git is a bit off on this one. It's because I moved some of the ``private`` methods to ``public`` and it interpreted that instead as moving the ``public:`` and the constructor rather than the functions.

src/MOCKernel.cpp
geogunow: In what way? Maybe the wording is awkward

src/MOCKernel.cpp
geogunow: ditto

src/MOCKernel.cpp
geogunow: ditto

src/MOCKernel.cpp
geogunow: It's possible if ``Tracks`` have not been generated yet

src/MOCKernel.cpp
geogunow: It's  because the ``SegmentationKernel`` (not provided here) needs ``row_num``. However, the reset of the kernels don't....maybe I can get rid of it for the rest.

src/MOCKernel.cpp
geogunow: Do you mean ``_weight``?

src/MOCKernel.cpp
geogunow: ``_count`` is the number of segments traversed since the creation or reset of the kernel

src/MOCKernel.cpp
geogunow: No, I'll take that out. It shouldn't need to do that.

src/CPUSolver.h
geogunow: What blank line?

wbinventor: Looks like you added in a line here

src/CPUSolver.h
wbinventor: I see now

src/MOCKernel.cpp
wbinventor: Yeah, I'd remove it for those that don't need or use it.

src/MOCKernel.cpp
wbinventor: Ok, I see that it isn't an incomplete comment, but it could be reworded to make it easier to read perhaps.

src/MOCKernel.cpp
wbinventor: Add that as a comment then. And perhaps reword the error message to reflect this? Otherwise, no user will understand 1) what the issue is or 2) how to fix it if they see this issue, even though it is as simple as generating tracks before setting this up.

src/MOCKernel.cpp
wbinventor: yes

src/MOCKernel.cpp
wbinventor: Ok, who or what cuts the segments then?

wbinventor: Oh I see it below

src/MOCKernel.cpp
```diff
@@ -0,0 +1,159 @@
...
+ * @details The CounterKernel execute function counts the number of segments
+ *          in a track by incrementing the counter variable upon execution. Due
+ *          to restrictions on maximum optical path length, the counter may be
+ *          incremented by more than one to account for splitting of the
+ *          segment into segments of allowed optical path length.
+ * @param length segment length
+ * @param mat Material associated with the segment
+ * @param id the FSR ID of the FSR associated with the segment
+ */
+void CounterKernel::execute(FP_PRECISION length, Material* mat, int id,
```
wbinventor: Might `SplitterKernel` be a more intuitive name than `CounterKernel` here?

src/MOCKernel.h
```diff
@@ -0,0 +1,117 @@
...
+ *          code. This class is the parent class of CounterKernel and
+ *          VolumeKernel. A generic MOCKernel should not be explicity
+ *          instantiated. Instead, an inhereting class should
+ *          be instantiated which describes the "execute" function.
+ */
+class MOCKernel {
+
+protected:
+
+  /** Count referring to the segment number */
```
wbinventor: The different comments on referring to what `_count` is seem to differ throughout the code. Can you come up with a consistent standard comment that you use everywhere for `_count`? And if need be, change `_count` to something more intuitive?

src/MOCKernel.cpp
```diff
@@ -0,0 +1,159 @@
...
+}
+
+
+/**
+ * @brief Constructor for the VolumeKernel assigns default values, calls
+ *        the MOCKernel constructor, and pulls refernces to FSR locks and FSR
+ *        volumes from the provided TrackGenerator.
+ * @param track_generator the TrackGenerator used to pull relevant tracking
+ *        data from
+ * @param row_num the row index into the temporary segments matrix
```
wbinventor: Is there really a segments "matrix"? I find `row_num` to be a non-intuitive name for a variable here, but perhaps that is just because I'm not as familiar with your OTF stuff to come.

src/MOCKernel.h
```diff
@@ -0,0 +1,117 @@
...
+
+  /* Sets the max optical path length to a different value */
+  void setMaxOpticalLength(FP_PRECISION max_tau);
+
+  /* Prepare MOCKernel for handling a new track */
+  virtual void newTrack(Track* track);
+
+  /* Executing function describes kernel behavior */
+  virtual void execute(FP_PRECISION length, Material* mat, int id,
+      int cmfd_surface_fwd, int cmfd_surface_bwd)=0;
```
wbinventor: tab this over inline with the parameters above

src/MOCKernel.h
```diff
@@ -0,0 +1,117 @@
...
+ * @details A CounterKernel inherets from MOCKernel and is a kernel which
+ *          counts the number of segments in a track by incrementing the
+ *          _count variable by the number of legitimate segment lengths
+ *          (less than the max optical path length) in the input length.
+ */
+class CounterKernel: public MOCKernel {
+public:
+  CounterKernel(TrackGenerator* track_generator, int row_num);
+  void execute(FP_PRECISION length, Material* mat, int id,
+      int cmfd_surface_fwd, int cmfd_surface_bwd);
```
wbinventor: tab this over to be inline with arguments above

src/MOCKernel.h
```diff
@@ -0,0 +1,117 @@
...
+
+};
+
+
+/**
+ * @class CounterKernel MOCKernel.h "src/MOCKernel.h"
+ * @brief Counts the number of segments of a track
+ * @details A CounterKernel inherets from MOCKernel and is a kernel which
+ *          counts the number of segments in a track by incrementing the
+ *          _count variable by the number of legitimate segment lengths
```
wbinventor: I think the last two lines of this comment may be a bit extraneous - referencing a private attribute `_count` in a public-facing comment seems like more info than necessary.

src/MOCKernel.h
```diff
@@ -0,0 +1,117 @@
...
+ *          (less than the max optical path length) in the input length.
+ */
+class CounterKernel: public MOCKernel {
+public:
```
wbinventor: Not a big deal, but you use spaces around `private` and `public` in your other class headers, but not here.

src/MOCKernel.h
```diff
@@ -0,0 +1,117 @@
...
+
+  /** The associated quadrature from which weights are derived */
+  Quadrature* _quadrature;
+
+public:
+
+  VolumeKernel(TrackGenerator* track_generator, int row_num);
+  void newTrack(Track* track);
+  void execute(FP_PRECISION length, Material* mat, int id,
+      int cmfd_surface_fwd, int cmfd_surface_bwd);
```
wbinventor: tab over to be inline with arguments above

src/Solver.cpp
```diff
@@ -298,6 +298,17 @@ FP_PRECISION Solver::getFlux(int fsr_id, int group) {
 
 
 /**
+ * @brief Returns the boundary flux array at the requested indexes
```
wbinventor: typo: "indexes" -> "indices"

wbinventor: actually, "flux array at the requested indexes" might be more clear if explained as "flux array for some track"

src/TrackGenerator.cpp
```diff
@@ -307,42 +313,14 @@ void TrackGenerator::calculateFSRVolumes() {
...
+    memset(_FSR_volumes, 0., num_FSRs*sizeof(FP_PRECISION));
 
-          /* Release FSR mutual exclusion lock */
-          omp_unset_lock(&_FSR_locks[fsr_id]);
-        }
-      }
-    }
-  }
+  /* Create volume calculator and calculate new FSR volumes */
+  VolumeCalculator volume_calculator(this);
```
wbinventor: Just to be clear, the idea is that this will work as is for a variety of tracking schemes including all combinations of explicit/OTF 2D/3D?

src/TrackGenerator.cpp
```diff
@@ -398,8 +376,13 @@ FP_PRECISION TrackGenerator::getFSRVolume(int fsr_id) {
 
 
 /**
- * @brief Finds and returns the maximum optical length amongst all segments.
- * @return the maximum optical path length
+ * @breif Calculates and returns the maximum optcial length for any segment
```
wbinventor: typo: "optcial" -> "optical"

src/TrackGenerator.cpp
```diff
@@ -398,8 +376,13 @@ FP_PRECISION TrackGenerator::getFSRVolume(int fsr_id) {
...
+ *        in the Geomtry.
+ * @details The _max_optical_length value is recomputed, updated, and returned.
```
wbinventor: Again, I lean against referencing a private variable `_max_optical_length` in a public-facing comment, and would instead opt for simply saying plain English "max optical path length"

src/TrackGenerator.cpp
```diff
@@ -398,8 +376,13 @@ FP_PRECISION TrackGenerator::getFSRVolume(int fsr_id) {
...
+ *          This value determines the when segments must be split during ray
```
wbinventor: typo: remove "the" from "the when"

src/TrackGenerator.cpp
```diff
@@ -398,8 +376,13 @@ FP_PRECISION TrackGenerator::getFSRVolume(int fsr_id) {
...
+ *          tracing.
+ * @return _max_optical_length the maximum optical length of any segment in the
```
wbinventor: Remove "_max_optical_length" from the `@return` doxygen comment

src/TrackGenerator.cpp
```diff
@@ -423,13 +406,27 @@ FP_PRECISION TrackGenerator::getMaxOpticalLength() {
           sigma_t = material->getSigmaT();
 
           for (int e=0; e < material->getNumEnergyGroups(); e++)
-            max_optical_length = std::max(max_optical_length, length*sigma_t[e]);
+            max_optical_length = std::max(max_optical_length,
```
wbinventor: Why isn't the max optical path length found in a kernel? Shouldn't all of this code iterating over tracks be abstracted into your kernel framework?

wbinventor: And can you simply update `_max_optical_length` here directly and remove line 416-417 below for simplicity?

src/TrackGenerator.cpp
```diff
@@ -866,6 +863,15 @@ void TrackGenerator::initializeTrackFileDirectory() {
 
 
 /**
+ * @brief Returns the type of ray tracing used for segment formation
+ * @return the segmentation type
```
wbinventor: Perhaps add a `@details` comment block here which enumerates the possible segment formation types for an amateur like me

src/TrackGenerator.h
```diff
@@ -94,6 +95,12 @@ class TrackGenerator {
   /** The z-coord where the 2D Tracks should be created */
   double _z_coord;
 
+  /** Determines the type of track segmentation to use */
+  segmentationType _segment_formation;
+
+  /** Max optical segment length for Tracks before splitting */
```
wbinventor: would be more accurate to say "max optical path length"  than "max optical segment length" to match the term used in the literature

src/TrackTraversingAlgorithms.cpp
```diff
@@ -0,0 +1,127 @@
...
+#include "CPUSolver.h"
+#include "Quadrature.h"
+
+
+/**
+ * @brief Constructor for SegmentSplitter calls the TraverseTracks
+ *        constructor
+ * @param track_generator The TrackGenerator to pull tracking information from
+ */
+VolumeCalculator::VolumeCalculator(TrackGenerator* track_generator)
```
wbinventor: Do you even need to specify this dummy constructor? Or is it implicitly defined to simply call the `TraverseTracks` constructor if you don't define it for `VolumeCalculator`? I think you can remove it altogether.

src/TraverseTracks.cpp
```diff
@@ -0,0 +1,153 @@
...
+void TraverseTracks::loopOverTracks(MOCKernel** kernels) {
+
+  switch (_segment_formation) {
+    case EXPLICIT_2D:
+      loopOverTracks2D(kernels);
+      break;
+  }
+
+  if (kernels != NULL) {
+    delete kernels[0];
```
wbinventor: What is this deletion of the first element in the array doing that the second deletion of the entire array does not do?

src/TraverseTracks.cpp
```diff
@@ -0,0 +1,153 @@
...
+ *        the functionality described in onTrack(...) to all Tracks.
+ * @details The segment formation method imported from the TrackGenerator
+ *          during construction is used to redirect to the appropriate looping
+ *          scheme. If kernels are provided (not NULL) then they are deleted at
+ *          the end of the looping scheme.
+ * @param kernels MOCKernels to apply to all segments
+ */
+void TraverseTracks::loopOverTracks(MOCKernel** kernels) {
+
+  switch (_segment_formation) {
```
wbinventor: Perhaps complete this `switch` block with the other cases, each with an `ERROR` or `WARNING` message?

src/TraverseTracks.cpp
```diff
@@ -0,0 +1,153 @@
...
+ *        applying the provided kernels to all segments and the functionality
+ *        described in onTrack(...) to all Tracks.
+ * @details The segment formation method imported from the TrackGenerator
+ *          during construction is used to redirect to the appropriate looping
+ *          scheme. If kernels are provided (not NULL) then they are deleted at
+ *          the end of the looping scheme.
+ * @param kernels MOCKernels to apply to all segments
+ */
+void TraverseTracks::loopOverTracksByParallelGroup(MOCKernel** kernels) {
+  loopOverTracksByParallelGroup2D(kernels);
```
wbinventor: Should this method body also have a `switch` over `_segment_formation`? If not, it seems extraneous to have this method to simply call `loopOverTracksByParallelGroup2D(...)`.

src/TraverseTracks.cpp
```diff
@@ -0,0 +1,153 @@
...
+  for (int a=0; a < num_azim/2; a++) {
+    int num_xy = _track_generator->getNumX(a) + _track_generator->getNumY(a);
+#pragma omp for
+    for (int i=0; i < num_xy; i++) {
+
+      Track* track_2D = &tracks_2D[a][i];
+
+      /* Operate on segments if necessary */
+      if (kernels != NULL) {
+        kernels[0]->newTrack(track_2D);
```
wbinventor: So this only uses the first kernel in the array? Why is the input an array of kernels then?

src/TraverseTracks.cpp
```diff
@@ -0,0 +1,153 @@
...
+  Track** tracks_2D = _track_generator->getTracks();
+  int num_azim = _track_generator->getNumAzim();
+  for (int a=0; a < num_azim/2; a++) {
+    int num_xy = _track_generator->getNumX(a) + _track_generator->getNumY(a);
+#pragma omp for
+    for (int i=0; i < num_xy; i++) {
+
+      Track* track_2D = &tracks_2D[a][i];
+
+      /* Operate on segments if necessary */
```
wbinventor: These comments seem confusing to me - this `if` block seems to be operating on a track (but the comment says "segments"), while the code in lines 90-91 seems to operating on segments (but the comment says "track"). I guess I'm confused about what is going on here, so perhaps bolster the comments a little bit?

src/TraverseTracks.cpp
```diff
@@ -0,0 +1,153 @@
...
+ * @param kernels MOCKernels to apply to all segments
+ */
+void TraverseTracks::loopOverTracksByParallelGroup(MOCKernel** kernels) {
+  loopOverTracksByParallelGroup2D(kernels);
+}
+
+
+/**
+ * @brief Loops over all explicit 2D Tracks
+ * @details The onTrack(...) function is applied to all 2D Tracks and the
```
wbinventor: You might add to you `@details` comment block that this `loopOverTracks2D(...)` only differs from `loopOverTracksByParallelGroup2D(...)` in terms of the parallelization (unless I've missed something). If I understand things correctly, `loopOverTracks2D(...)` does not enforce any chunking or ordering of operations across tracks, while `loopOverTracksByParallelGroup2D(...)` does enforce chunking / ordering of operations according to some angular criteria. This should be added to the comments for each method to really pinpoint what differentiates them.

src/TraverseTracks.cpp
```diff
@@ -0,0 +1,153 @@
...
+      Track* track = tracks[track_id];
+
+      /* Operate on segments if necessary */
+      if (kernels != NULL) {
+        kernels[0]->newTrack(track);
+        traceSegmentsExplicit(track, kernels[0]);
+      }
+
+      /* Operate on the Track */
+      segment* segments = track->getSegments();
```
wbinventor: If you pass in a `Track` to the `onTrack(...)` method, why do you need to extract the segments from the `Track` as a separate parameter? Can't the `onTrack(...)` method simply extract the segments from the `Track` parameter itself? The segments seem like an unnecessary parameter.

src/TraverseTracks.cpp
```diff
@@ -0,0 +1,153 @@
...
+ * @brief Loops over segments in a Track when segments are explicitly generated
+ * @details All segments in the provided Track are looped over and the provided
+ *          MOCKernel is applied to them.
+ * @param track The Track whose segments will be traversed
+ * @param kernel The kernel to apply to all segments
+ */
+void TraverseTracks::traceSegmentsExplicit(Track* track, MOCKernel* kernel) {
+  for (int s=0; s < track->getNumSegments(); s++) {
+    segment* seg = track->getSegment(s);
+    kernel->execute(seg->_length, seg->_material, seg->_region_id,
```
wbinventor: Whew, why unwrap all of these arguments from a `segment` for the kernel's `execute(...)` method? Why can't a `segment` simply be passed to a kernel, and then it can decide which attributes it needs to reference? Seems like it would really simplify the `execute(...)` method signature.

src/TraverseTracks.h
```diff
@@ -0,0 +1,77 @@
...
+  void loopOverTracks(MOCKernel** kernels);
+  void loopOverTracksByParallelGroup(MOCKernel** kernels);
+  virtual void onTrack(Track* track, segment* segments) = 0;
+
+  /* Returns a matrix of kernels of the requested type */
+  template <class KernelType>
+  MOCKernel** getKernels() {
+
+    /* Allocate kernels */
+    MOCKernel** kernels = new MOCKernel*[1];
```
wbinventor: So `getKernels()` will always return a length 1 array of kernels? Why use an array then?

wbinventor:
@geogunow I've completed my first full pass through this PR. This is certainly a paradigm shift and took some time to digest, but I think I can see where this is headed and the power of this abstraction. Nice work, looking forward to adding this to the code!

src/MOCKernel.cpp
geogunow: I actually have a ``SegmentSplitter`` in ClosedMOC -- but that one actually splits segments. The ``CounterKernel`` just counts how many segments there are along a ``Track``.

src/MOCKernel.h
geogunow: The definition of ``_count`` is always the same. It is the number of segments traversed since the last reset.

src/MOCKernel.h
geogunow: I'm not sure what you mean by this

src/Solver.cpp
geogunow: http://grammarist.com/usage/indexes-indices/

src/TrackGenerator.cpp
geogunow: yes

src/TrackGenerator.cpp
geogunow: Not quite because we first need to compute the maximum optical length then use it for splitting segments for every ``MOCKernel`` created from then on. ``MOCKernels`` are temporary but the ``TrackGenerator`` endures for the entire calculation so the maximum optical path length is stored there.

src/TrackGenerator.cpp
geogunow: Right now it would read "The only segmentation type currently supported is ``EXPLICIT_2D``". But that doesn't mean a whole lot right now. It would also be incorrect if we added 3D options and forgot to update that comment.

src/TraverseTracks.cpp
geogunow: Not really necessary since ``_segment_formation`` is an enum type that we define and all options should be supported.

src/TraverseTracks.cpp
geogunow: Sure

src/MOCKernel.h
samuelshaner: I believe what he means is that there should be a blank line between `class CounterKernel : public MOCKernel {` and `public`. Take a look at the other header files to see.

src/Solver.cpp
samuelshaner: So are you going to change it to indices or "flux array for some track". Your link seems to suggest indices is the way to go for this context.

src/TraverseTracks.cpp
geogunow: Not if your segments are formed on-the-fly

geogunow:
@wbinventor I think this PR is ready for further review

src/MOCKernel.cpp
```diff
@@ -0,0 +1,156 @@
...
+ */
+MOCKernel::MOCKernel(TrackGenerator* track_generator) {
+  _count = 0;
+  _max_tau = track_generator->retrieveMaxOpticalLength();
+}
+
+
+/**
+ * @brief Constructor for the VolumeKernel assigns default values, calls
+ *        the MOCKernel constructor, and pulls refernces to FSR locks and FSR
```
wbinventor: typo: "refernces" -> "references"

src/MOCKernel.cpp
```diff
@@ -0,0 +1,156 @@
...
+ *        volumes from the provided TrackGenerator.
+ * @param track_generator the TrackGenerator used to pull relevant tracking
+ *        data
+ */
+VolumeKernel::VolumeKernel(TrackGenerator* track_generator) :
+                           MOCKernel(track_generator) {
+  _FSR_locks = track_generator->getFSRLocks();
+
+  if (_FSR_locks == NULL)
+    log_printf(ERROR, "Unable to create a VolumeKernel without first creating "
```
wbinventor: Does the practical meaning of this error message mean that tracks must be generated first?

src/MOCKernel.cpp
```diff
@@ -0,0 +1,156 @@
...
+ */
+int MOCKernel::getCount() {
+  return _count;
+}
+
+
+/* @brief Resets the maximum optcal path length for a segment
+ * @details MOC kernels ensure that there are no segments with an optical path
+ *          length greater than the maximum optical path length by splitting
+ *          then when they get too large.
```
wbinventor: typo: "then" -> "them"

src/MOCKernel.cpp
```diff
@@ -0,0 +1,156 @@
...
+ *          in a track by incrementing the counter variable upon execution. Due
+ *          to restrictions on maximum optical path length, the counter may be
+ *          incremented by more than one to account for splitting of the
+ *          segment into segments of allowed optical path length.
+ * @param length segment length
+ * @param mat Material associated with the segment
+ * @param id the FSR ID of the FSR associated with the segment
+ */
+void CounterKernel::execute(FP_PRECISION length, Material* mat, int id,
+    int cmfd_surface_fwd, int cmfd_surface_bwd) {
```
wbinventor: it looks like this continuation line for the method signature isn't tabbed over inline with the line above

src/MOCKernel.h
```diff
@@ -0,0 +1,121 @@
...
+ * @file MOCKernel.h
+ * @brief An MOCKernel object
+ * @date May 5, 2015
+ * @author Geoffrey Gunow, MIT, Course 22 (geogunow@mit.edu)
+ */
+
+#ifndef MOCKERNEL_H_
+#define MOCKERNEL_H_
+
+#include "Python.h"
```
wbinventor: This include of "Python.h" should be wrapped with an `#ifdef` for the `SWIG` macro as so:

```c++
#ifdef SWIG
#include "Python.h"
#endif
```

This is how we include this file elsewhere. There is a very good reason for doing this, though it slips my mind at this time at night.

src/MOCKernel.h
```diff
@@ -0,0 +1,121 @@
...
+ *        segments.
+ * @details An MOCKernel is an object that owns some specified data and
+ *          contains an "execute" function which applies some functionality
+ *          to the data. This is useful in MOC where it is very common to
+ *          apply some function to segment data in either a nested loop
+ *          structure or from on-the-fly calculations. Kernels specify the
+ *          actions applied to the segments, reducing the need for repeated
+ *          code. This class is the parent class of CounterKernel and
+ *          VolumeKernel. A generic MOCKernel should not be explicity
+ *          instantiated. Instead, an inhereting class should
```
wbinventor: typo: "inhereting" -> "inheriting"

src/MOCKernel.h
```diff
@@ -0,0 +1,121 @@
...
+ *          be instantiated which describes the "execute" function.
```
wbinventor: to be more accurate, "describes" should be "implements" here

src/MOCKernel.h
```diff
@@ -0,0 +1,121 @@
...
+                       int cmfd_surface_fwd, int cmfd_surface_bwd)=0;
+
+};
+
+
+/**
+ * @class CounterKernel MOCKernel.h "src/MOCKernel.h"
+ * @brief Counts the number of segments of a track
+ * @details A CounterKernel inherets from MOCKernel and is a kernel which
+ *          tallies the number of legitamte segment lengths (less than the max
```
wbinventor: typo: "legitamte" -> "legitimate"

src/Solver.cpp
```diff
@@ -298,6 +298,17 @@ FP_PRECISION Solver::getFlux(int fsr_id, int group) {
 
 
 /**
+ * @brief Returns the boundary flux array for a Track
+ * @param track_id The Track's Unique ID
```
wbinventor: no need to capitalize "Unique" (nbd though)

src/TrackGenerator.cpp
```diff
@@ -398,8 +376,12 @@ FP_PRECISION TrackGenerator::getFSRVolume(int fsr_id) {
 
 
 /**
- * @brief Finds and returns the maximum optical length amongst all segments.
- * @return the maximum optical path length
+ * @breif Calculates and returns the maximum optical length for any segment
```
wbinventor: typo: "breif" -> "brief"

src/TrackGenerator.cpp
wbinventor: Sure, that's fine by me

src/TraverseTracks.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * @brief Destructor for TraverseTracks
+ */
+TraverseTracks::~TraverseTracks() {
+}
+
+
+/**
+ * @breif Loops over Tracks, applying the provided kernel to all segments and
```
wbinventor: typo: "breif" -> "brief"

src/TraverseTracks.cpp
```diff
@@ -0,0 +1,154 @@
...
+ *        the functionality described in onTrack(...) to all Tracks.
+ * @details The segment formation method imported from the TrackGenerator
+ *          during construction is used to redirect to the appropriate looping
+ *          scheme.
+ * @param kernel The MOCKernel to apply to all segments
+ */
+void TraverseTracks::loopOverTracks(MOCKernel* kernel) {
+  switch (_segment_formation) {
```
wbinventor: is this `switch-case` statement really necessary? Perhaps it's just a stub for the future.

src/TraverseTracks.cpp
```diff
@@ -0,0 +1,154 @@
...
+    case EXPLICIT_2D:
+      loopOverTracks2D(kernel);
+      break;
+  }
+}
+
+
+/**
+ * @breif Loops over Tracks by parallel groups to avoid potential conflicts,
```
wbinventor: ditto, typo: "breif" -> "brief"

src/TraverseTracks.cpp
```diff
@@ -0,0 +1,154 @@
...
+ *        applying the provided kernel to all segments and the functionality
+ *        described in onTrack(...) to all Tracks.
+ * @details The segment formation method imported from the TrackGenerator
+ *          during construction is used to redirect to the appropriate looping
+ *          scheme.
+ * @param kernel The MOCKernel to apply to all segments
+ */
+void TraverseTracks::loopOverTracksByParallelGroup(MOCKernel* kernel) {
+  switch (_segment_formation) {
```
wbinventor: ditto, same question about `switch-case` necessity

src/TraverseTracks.cpp
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+}
+
+
+/**
+ * @brief Loops over all explicit 2D Tracks by parallel group
+ * @details This function is the same as loopOverTracks2D except that
+ *          Tracks are traversed in a safe order such that there are no
+ *          conflicts in exchanging data between Tracks in parallel.
```
wbinventor: by "safe order" i presume you mean that there are no race conditions?

src/TraverseTracks.cpp
wbinventor: I'm still curious about your thoughts related to this previous comment of mine.

wbinventor:
@geogunow I've made a final review of this PR and it looks good pending my super minor comments. I do want to pull your branch and try it out with some of my MGXS datasets just to ensure that all works as expected. I'll let you know once I have a chance to do so. Would be good to get the go-ahead from @samuelshaner if he needs to review this in the meantime.

src/MOCKernel.cpp
geogunow: Yes, but it's possible that you could create locks without generating Tracks. If we go to broader methods like TRRM or Multi-group Monte Carlo, then Tracks would not be generated.

src/MOCKernel.h
geogunow: It's for the alternative C++ build

src/TraverseTracks.cpp
geogunow: Yeah, that's exactly what it is

src/TraverseTracks.cpp
geogunow: Same thing

src/TraverseTracks.cpp
geogunow: Yeah

geogunow:
@wbinventor I addressed your comments in a commit but for some reason the commit isn't showing up on here. Seems to be a problem with Github.

wbinventor:
@geogunow - I am testing out your code now. Have you tested it for `GPUSolver`? Can you take charge of doing that on AWS?

geogunow:
@wbinventor sure. I think I tested the GPU solver before the edits but I can do it again.

wbinventor:
Yeah that would be great. I'll redundantly test all of the other inputs, as well as some of the "multi-sim" use cases that I rely on for my work.

wbinventor:
I just ran all of the sample inputs save for the "c5g7-gpu.py" - all looks good, but can you change `xrange` to `range in "cosine-source.py" for Python 3 compatibility? Also, this works fantastic with my "multi-sim" use cases. I think it is good to go. Any reason to hesitate on merging this in @samuelshaner?

geogunow:
@wbinventor it seems there is a double free at the end of the ``GPUSolver`` (likely its destructor). I'm not sure where it is yet though.

src/TraverseTracks.cpp
```diff
@@ -0,0 +1,158 @@
...
+ * @param kernel The MOCKernel to apply to all segments
+ */
+void TraverseTracks::loopOverTracksByParallelGroup2D(MOCKernel* kernel) {
```
samuelshaner: Do you want to just remove the parallel group and store the old track fluxes since that's what we are transitioning to for 3D?

samuelshaner:
Once we get resolution on whether we want to remove the parallel track groups and confirm it works for the GPUSolver I think this is ready to merge.

wbinventor:
@geogunow is the double free in the `GPUSolver` something that is specific to the code in this PR, or is it reproducible with the develop branch? I would help test this out, but have recently discovered that StarCluster is not Python 3.5 compatible (which is all I'm using these days) and don't have ready access to AWS atm.

geogunow:
@wbinventor I fixed the bug in the GPU solver. Now the only topic to address is whether to remove the parallel track groups. Would you be okay with that?

wbinventor:
Awesome, thanks for tracking down the issue @geogunow. I don't know about
the parallel track groups, isn't that something @samuelshaner asked about?

samuelshaner:
I'd say just take out the parallel tracks group since that's where we are going with domain decomposition.

wbinventor:
Hold your breath!

PR #311, opened 2016-05-18
PolarQuad.cpp replaced with Quadrature.cpp in the profile/makefile

zechzhao:
As PolarQuad.cpp is not defined in src/


wbinventor:
Thanks for the patch @zechzhao! Do you want to also add the fix for the CMFD group structure setter that you pointed out?

zechzhao:
I was only able to make the compiler doesn't complain. However, I'm not quite sure what is the meaning of the original code.

And Am I using the pull function right? Not sure if I should start a new thread.

profile/Makefile
```diff
@@ -13,7 +13,7 @@ PRECISION   = single
 # Source Code List
 #===============================================================================
 
-case = c5g7/c5g7-cmfd.cpp
+case = c5g7/c5g7-cmfd.cpp
```
geogunow: Remove the space after ``c5g7/c5g7-cmfd.cpp``. Per our style guide, all trailing white spaces should be removed.

profile/models/c5g7/c5g7-cmfd.cpp
```diff
@@ -465,8 +465,8 @@ int main() {
   Cmfd cmfd;
   cmfd.setSORRelaxationFactor(1.5);
   cmfd.setLatticeStructure(51, 51);
-  int cmfd_group_structure[3] = {1,4,8};
-  cmfd.setGroupStructure(cmfd_group_structure, 3);
+  std::vector< std::vector<int> > cmfd_group_structure = {{1,4,7}};
```
geogunow: Since the previous update of the alternative C++ build, the syntax for setting the CMFD group structure has changed. Previously a single array was used to indicate how to collapse the groups (it was a bit complicated, so I'll skip the details). Now, we collapse groups much more straightforward. In particular, if you want to collapse to two groups for CMFD (which is what we want) you would write the following:
```
std::vector< std::vector<int> > cmfd_group_structure = {{1,2,3}, {4,5,6,7}};
```

which would indicate the first three groups would be collapsed into group one and the last four groups would be collapsed into group two.

geogunow:
@zechzhao Thanks for submitting this PR! It looks like we were a big negligent in updating the alternative C++ build for recent changes in the code. I have a few comments on the changes that need to be addressed before this is merged. Let me know if you run into any problems.

profile/models/c5g7/c5g7-cmfd.cpp
wbinventor: @zechzhao if you revise to the recommendation by @geogunow above we will merge in your PR right away.

wbinventor:
Thank you for the bug fix @zechzhao!

PR #313, opened 2016-05-30
Switching to Point Jacobi Iteration Scheme

geogunow:
In #307, the CPU solver changed the iteration scheme from updating boundary angular fluxes between connecting tracks immediately within parallel groups (sometimes loosely referred to as "Gauss-Siedel") to updating all boundary angular fluxes at the end of each iteration (referred to as Point Jacobi).

The "Gauss-Siedel" scheme is advantageous for a few reasons:
* More recent update on angular flux provides a better convergence rate
* Only one group of boundary angular fluxes need to be stored

However, it also has its disadvantages:
* Increased complexity due to the need to form "parallel groups" that can be run concurrently
* A slight reduction in parallelism (only tracks in a parallel group can be run concurrently)
* Difficult to extend with domain decomposition due to increased communication between nodes

For primarily the last reason (domain decomposition) we move to a Point Jacobi iteration scheme throughout the entire code in this PR. With this decision we do loose a bit of performance. First, we require more iterations to converge the solution due to a lagged update of the boundary angular fluxes. However with CMFD acceleration this effect is minimal (often only one or two iterations). Second, we double the storage requirements for boundary angular fluxes since we need to store a copy for each boundary angular flux which is transferred at the end of the iteration. Since we expect boundary angular flues to dominate the storage requirements in large MOC problems (particularly 3D MOC) this doubles the entire memory footprint of the problem. However, we have decided the reduction in complexity is worth taking this penalty. 

This PR mainly focuses on the GPU and implements the Point Jacobi scheme on the GPU.

### DIFF
--- a/src/TrackGenerator.cpp
+++ b/src/TrackGenerator.cpp
@@ -38,13 +38,12 @@ TrackGenerator::~TrackGenerator() {
     delete [] _num_tracks;
     delete [] _num_x;
     delete [] _num_y;
-    delete [] _num_tracks_by_parallel_group;
 
     for (int i = 0; i < _num_azim_2; i++)
       delete [] _tracks[i];
 
     delete [] _tracks;
-    delete [] _tracks_by_parallel_group;
+    delete [] _tracks_array;
   }
 
   if (_FSR_locks != NULL)
@@ -194,8 +193,7 @@ Track **TrackGenerator::getTracks() {
 
 /**
  * @brief Returns a 1D array of Track pointers.
- * @details The tracks in the _tracks_by_parallel_group array are organized
- *          by UID.
+ * @details The tracks in the tracks array are organized by UID.
  * @return The 1D array of Track pointers
  */
 Track **TrackGenerator::getTracksArray() {
@@ -663,10 +661,9 @@ void TrackGenerator::generateTracks(bool store, bool neighbor_cells) {
   /* Deletes Tracks arrays if Tracks have been generated */
   if (_contains_tracks) {
     delete [] _num_tracks;
-    delete [] _num_tracks_by_parallel_group;
     delete [] _num_x;
     delete [] _num_y;
-    delete [] _tracks_by_parallel_group;
+    delete [] _tracks_array;
 
     for (int i = 0; i < _num_azim_2; i++)
       delete [] _tracks[i];
--- a/src/TrackGenerator.cpp
+++ b/src/TrackGenerator.cpp
@@ -1106,7 +1106,6 @@ void TrackGenerator::initializeTrackUids() {
       Track* track = &_tracks[a][i];
       track->setUid(uid);
       uid++;
-      }
     }
   }
 }
--- a/src/TrackGenerator.cpp
+++ b/src/TrackGenerator.cpp
@@ -127,20 +127,6 @@ int TrackGenerator::getNumTracks() {
 
 
 /**
- * @brief Return the number of parallel track groups.
- * @return The number of parallel track groups
- */
-int TrackGenerator::getNumParallelTrackGroups() {
-
-  if (!_contains_tracks)
-    log_printf(ERROR, "Unable to return the number of parallel track groups "
-               "since Tracks have not yet been generated.");
-
-  return _num_parallel_track_groups;
-}
-
-
-/**
  * @brief Return the number of tracks on the x-axis for a given azimuthal angle.
  * @param azim An azimuthal angle index
  * @return The number of Tracks on the x-axis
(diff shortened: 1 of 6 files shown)


src/TrackGenerator.cpp
```diff
@@ -1131,104 +1092,22 @@ void TrackGenerator::initializeTrackCycleIndices(boundaryType bc) {
...
-        track = &_tracks[a][i];
-
-        /* Get the track azim group id */
-        track_azim_group_id = a / (_num_azim_2 / 2);
-
-        /* Get the track periodic group id */
-        if (periodic) {
-          track_periodic_index = track->getPeriodicTrackIndex();
+  /* Allocate memory for the 1D tracks array */
+  _tracks_array = new Track*[getNumTracks()];
```
wbinventor: Should this deallocate `_tracks_array` if it was previously allocated (*i.e.*, in a multi-sim use case)?

src/accel/cuda/GPUSolver.cu
```diff
@@ -1484,25 +1490,24 @@ void GPUSolver::transportSweep() {
...
   /* Initialize flux in each FSR to zero */
   flattenFSRFluxes(0.0);
 
-  /* Loop over the parallel track groups and perform transport sweep on tracks
-   * in that group */
-  for (int g=0; g < _num_parallel_track_groups; g++) {
-    tid_offset = tid_max * _num_groups;
-    tid_max += _track_generator->getNumTracksByParallelGroup(g);
+  /* Copy starting flux to current flux */
+  cudaMemcpy((void*)boundary_flux, (void*)start_flux, _tot_num_tracks
```
wbinventor: Should the size be `_tot_num_tracks * _polar_times_groups`?

src/accel/cuda/GPUSolver.h
```diff
@@ -53,6 +53,7 @@
 /** Indexing macro for the angular fluxes for each polar angle and energy
  *  group for a given Track */
 #define boundary_flux(t,pe2) (boundary_flux[2*(t)*(*polar_times_groups)+(pe2)])
+#define start_flux(t,pe2) (start_flux[2*(t)*(*polar_times_groups)+(pe2)])
```
wbinventor: Can you add a doxygen comment above `start_flux` akin to those for the other macros?

wbinventor:
@geogunow this looks pretty straightforward. Since I'm not set up to launch a GPU node using StarCluster atm, I'll consider your tests to be sufficient. It would be nice to see a table of eigenvalues and iteration counts for C5G7 `CPUSolver` and `GPUSolver` before/after the move to Jacobi iteration, for both forward and adjoint calculations.

src/TrackGenerator.cpp
geogunow: @wbinventor we didn't do that before. Right now it gets deleted (along with other things) when you call ``TrackGenerator::generateTracks``. Though I could be convinced to restructure it to be deleted closer to where it's allocated.

src/accel/cuda/GPUSolver.cu
geogunow: yep

src/TrackGenerator.cpp
wbinventor: Well I know we have at least one or two memory leaks lying around since the memory footprint grows over time when I run multi-sim cases with ray tracing 100s of times. Perhaps this is part of the reason.

src/TrackGenerator.cpp
```diff
@@ -38,13 +38,12 @@ TrackGenerator::~TrackGenerator() {
...
     delete [] _num_x;
     delete [] _num_y;
-    delete [] _num_tracks_by_parallel_group;
 
     for (int i = 0; i < _num_azim_2; i++)
       delete [] _tracks[i];
 
     delete [] _tracks;
-    delete [] _tracks_by_parallel_group;
+    delete [] _tracks_array;
```
samuelshaner: Could we just have one tracks array? It seems a little unnecessary to have two. Perhaps we could then have a macro that allows the tracks to be indexed using the azimuthal and xy index.

src/TrackGenerator.cpp
```diff
@@ -1131,104 +1092,22 @@ void TrackGenerator::initializeTrackCycleIndices(boundaryType bc) {
...
-        /* Check if track has current azim_group_id and periodic_group_id */
-        if (azim_group_id == track_azim_group_id &&
-            periodic_group_id == track_periodic_group_id) {
-          track->setUid(uid);
-          _tracks_by_parallel_group[uid] = track;
-          uid++;
-          num_tracks++;
-        }
-      }
+  /* Loop over all tracks and assigns UIDs */
```
samuelshaner: Either `Loop..assign` or `Loops...assigns`

src/TrackGenerator.cpp
geogunow: It shouldn't be.  This function is only called from within ``TrackGenerator::generateTracks``.

wbinventor: Yeah, but I generate tracks 100s of times for different clustered
geometries. Should only be a two line change to deallocate if necessary.

geogunow:
For C5G7 (forward), here are the results:

| Iteration Scheme | Eigenvalue | Iteration Count  | Time per Integration |  Total time to solution |
| ------------- | ------------- | ------------- | ---------------------------------------- | ---------------------------------------- |
| "Gauss-Seidel"  | 1.185592  |  627 | 1.20 ns  | 18.8 s |
| Point Jacobi  | 1.185586  | 641  | 1.09 ns  | 17.5 s |

wbinventor:
Thanks for confirming the accuracy @geogunow.

PR #314, opened 2016-06-01
MC Solver

ljeure:
I added the source files for the Monte Carlo Solver to the src directory. I created a profile for the Monte Carlo Solver called monte_carlo. This profile calculates the eigenvalue, but does not make any plots because SWIG has not been implemented yet.


src/Universe.cpp
```diff
@@ -729,6 +729,7 @@ void Universe::subdivideCells(double max_radius) {
       else
         static_cast<Lattice*>(fill)->subdivideCells(max_radius);
     }
+
```
geogunow: Remove

geogunow:
Remove all the binary files and .swo files from begin tracked in the PR (i.e. git rm <filename>)

geogunow:
@ljeure Do we never use ``Geometry::findNextCell``? And if so how come?

profile/Makefile
```diff
@@ -34,14 +34,25 @@ Timer.cpp \
 Track.cpp \
 TrackGenerator.cpp \
 Universe.cpp \
-Vector.cpp
+Vector.cpp \
+\
```
geogunow: Remove

profile/Makefile
```diff
@@ -34,14 +34,25 @@ Timer.cpp \
...
+TraverseTracks.cpp \
+TrackTraversingAlgorithms.cpp \
+MOCKernel.cpp \
+\
```
geogunow: Remove

profile/Makefile
```diff
@@ -34,14 +34,25 @@ Timer.cpp \
...
 
 cases = \
 gradients/one-directional/one-directional-gradient.cpp \
 gradients/two-directional/two-directional-gradient.cpp \
 homogeneous/homogeneous-one-group.cpp \
 c5g7/c5g7.cpp \
-c5g7/c5g7-cmfd.cpp
+c5g7/c5g7-cmfd.cpp \
+monte_carlo/monte_carlo.cpp
+
```
geogunow: Remove line

src/Neutron.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+/*
+ @brief     returns a pseudo-random number using the seed between 0 and 1
+ @return    a psuedo-random number between 0 and 1
+*/
+double Neutron::arand() {
+    double r = rand_r(&_seed);
+    return (double) r / (double) RAND_MAX;
+}
+
```
geogunow: Put two empty lines between all methods

src/MCSolver.cpp
```diff
@@ -0,0 +1,545 @@
...
+            crow distance
+ @param     n_histories number of neutron histories to run
+ @param     mat a Material object containing information
+            about the material
+ @param     flux a Flux object containing information about the flux
+ @param     num_batches the number of batches to be tested
+ @param     num_groups the number of neutron energy groups
+*/
+void MCSolver::computeEigenValue(int n_histories, int num_batches,
+        int num_groups) {
```
geogunow: We should be able to know the number of groups from the ``Geometry`` which was set during ``MCSolver::setGeometry(...)`` so we shouldn't have it as a parameter here

src/MCSolver.cpp
```diff
@@ -0,0 +1,545 @@
...
+    std::vector <Tally> tallies(5);
+    Fission fission_banks;
+    
+    bool first_round = true;
+
+    for (int batch=1; batch <= num_batches; ++batch) {
+
+        // clear flux data
+        for (int i=0; i<_num_groups; ++i) {
+            for (int j=0; j<_geometry->getNumFSRs(); ++j) {
```
geogunow: We should probably reverse the ordering of these loops and to be consistent with the rest of OpenMOC use ``i`` for the fsr number and ``g`` for group number

profile/models/monte_carlo/monte_carlo.cpp
```diff
@@ -0,0 +1,153 @@
...
+    geometry->setRootUniverse(root_universe);
+    
+    // simulate neutron histories
+    int num_neutrons = 100000;
+    int num_batches = 3;
+    
+    // create solver
+    MCSolver solver;
+    solver.setGeometry(geometry);
+    solver.setRootCell(root_cell);
```
geogunow: The ``Geometry`` should know the root cell so this should be done inside ``setGeometry`` rather than in the input

profile/models/monte_carlo/monte_carlo.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+    // initialize solver function
+    solver.initializeFSRs(lattice);
+    solver.initializeFlux();
```
geogunow: I feel like there should just be one ``MCSoler::initialize(...)`` function rather than two

src/MCSolver.cpp
```diff
@@ -0,0 +1,545 @@
...
+void MCSolver::storeFSRFluxes() {}
+void MCSolver::computeKeff() {}
+double MCSolver::computeResidual(residualType res_type){
+    return 0;
+}
+
+/*
+ @brief initialize the fsrs
+*/
+void MCSolver::initializeFSRs(Lattice* lattice) {
```
geogunow: I totally forgot about this....hmmm...we'll need to make this more general. I'll take some time to think about this.

profile/models/monte_carlo/monte_carlo.cpp
ljeure: How do I get the cell from the Geometry? I tried doing _geometry->getRootUniverse()->getCell(0), because 0 is the ID number of the Root Cell but it says that the Root Universe doesn't contain that cell.

profile/Makefile
ljeure: @geogunow We use Geometry::findNextCell in MCSolver.cpp in the code I copied from the segmentize routine. It's used in finding the next cell the neutron will hit along its path.

profile/models/monte_carlo/monte_carlo.cpp
geogunow: This is definitely possible....but we need to do something altogether different in order to treat arbitrary geometries. I'm working on a patch right now.

geogunow:
I'm going to close this PR because:
1) MCSolver still has some bugs
2) More importantly, this PR should be directed to a new branch (not develop) since we don't really plan on making this a main addition to OpenMOC. Rather it is an optional add-on that we should sync with develop but not be an essential part of it.

PR #317, opened 2016-07-18
Created local variables to store Universe extrema

ljeure:
Previously when the Universe functions getMinX(), getMaxX(), ... , getMaxZ() were called, the min or max in question was calculated, then returned. I added a function calculateExtrema() which calculates each min and max and stores them in local variables. When a getMin or getMax function is called, the extrema are recalculated only if cells have been added to or removed from the geometry since the extrema were last called.

### DIFF
--- a/src/Universe.cpp
+++ b/src/Universe.cpp
@@ -1656,7 +1656,7 @@ void Universe::calculateExtrema() {
 
       if (surf->getSurfaceType() == XPLANE && halfspace == +1 &&
           surf->getBoundaryType() != BOUNDARY_NONE)
-        return surf->getMinX(halfspace);
+        min_x = surf->getMinX(halfspace);
     }
   }
 
@@ -1680,7 +1680,7 @@ void Universe::calculateExtrema() {
       halfspace = s_iter->second->_halfspace;
       if (surf->getSurfaceType() == XPLANE && halfspace == -1 &&
           surf->getBoundaryType() != BOUNDARY_NONE)
-        return surf->getMaxX(halfspace);
+        max_x = surf->getMaxX(halfspace);
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/Universe.cpp
```diff
@@ -1720,3 +1631,130 @@ int Lattice::getLatticeSurface(int cell, Point* point) {
...
+  }
+
+  /* If a y-max boundary was not found, get the y-max from the bounding boxes
+   * of the cells */
+  for (c_iter = _cells.begin(); c_iter != _cells.end(); ++c_iter)
+    max_y = std::max(max_y, c_iter->second->getMaxY());
+
+  _max_y = max_y;
+
+  //min z
```
geogunow: Let's keep consistent with the comment style. as the previous comment has the /* ... */ style, we should do that here as well. Also let's make this comment more descriptive.

src/Universe.cpp
```diff
@@ -1720,3 +1631,130 @@ int Lattice::getLatticeSurface(int cell, Point* point) {
...
+  double min_z = -std::numeric_limits<double>::infinity();
+  std::map<int, Cell*>::iterator iter;
+
+  for (iter = _cells.begin(); iter != _cells.end(); ++iter)
+    min_z = std::min(min_z, iter->second->getMinZ());
+
+  _min_z = min_z;
+
+  // max z
```
geogunow: ditto, for all min x, max x, min y, max y, etc

geogunow:
@wbinventor does this seem alright to you? I seem to remember we previously had a ``findBoundingBox`` routine that I think did something like this and we opted to remove it for searching every time. I believe the reasoning was that determining boundaries was only done as a pre-processing step for ``Tracks`` in OpenMOC. However with Luke's Monte Carlo work, it makes his code run incredibly slow since it needs to do a full geometry traversal each time it looks for an intersection.

I forget the reason why we made the change to remove ``findBoundingBox`` in favor of searching. Do you remember? Also if you do, does this change handle the issue we were trying to address?

PR #319, opened 2016-07-27
Gauss Legendre quadrature for arbitrary number of angles

ljeure:
Previously the abscissa and weights used in Gauss Legendre quadrature were hardcoded into Quadrature.cpp. I added functions that calculate the roots of Legendre equations and the weights to be used in GL quadrature.

The abscissa used in GL quadrature are the roots to the Legendre polynomials. getLegendreRoots() calculates all the roots simultaneously using an Alberth-Householder-n method. The algorithm was designed by Pablo Ducru. The use of recursive functions causes the runtime of the function to skyrocket for polynomials of degree > 40.

The weights are caclulated by the function getGLWeights().

Running a c5g7 simulation with the new method for GL quadrature yielded a k-eff that was 4 PCM different from the k-eff produced a c5g7 simulation with the default settings.

### DIFF
--- a/profile/Makefile
+++ b/profile/Makefile
@@ -13,22 +13,22 @@ PRECISION   = single
 # Source Code List
 #===============================================================================
 
-case = test_quad/test_quad.cpp
+case = c5g7/c5g7-cmfd.cpp
 
 source = \
-Quadrature.cpp \
-log.cpp 
-#Cell.cpp \
+Cell.cpp \
 Cmfd.cpp \
 CPUSolver.cpp \
 ExpEvaluator.cpp \
 Geometry.cpp \
 LocalCoords.cpp \
 linalg.cpp \
+log.cpp \
 Material.cpp \
 Matrix.cpp \
 MOCKernel.cpp \
 Point.cpp \
+Quadrature.cpp \
 Solver.cpp \
 Surface.cpp \
 Timer.cpp \
@@ -44,7 +44,6 @@ gradients/one-directional/one-directional-gradient.cpp \
 gradients/two-directional/two-directional-gradient.cpp \
 homogeneous/homogeneous-one-group.cpp \
 c5g7/c5g7.cpp \
-test_quad/test_quad.cpp \
 c5g7/c5g7-cmfd.cpp
 
 #===============================================================================
--- a/profile/Makefile
+++ b/profile/Makefile
@@ -13,22 +13,22 @@ PRECISION   = single
 # Source Code List
 #===============================================================================
 
-case = c5g7/c5g7-cmfd.cpp
+case = test_quad/test_quad.cpp
 
 source = \
-Cell.cpp \
+Quadrature.cpp \
+log.cpp 
+#Cell.cpp \
 Cmfd.cpp \
 CPUSolver.cpp \
 ExpEvaluator.cpp \
 Geometry.cpp \
 LocalCoords.cpp \
 linalg.cpp \
-log.cpp \
 Material.cpp \
 Matrix.cpp \
 MOCKernel.cpp \
 Point.cpp \
-Quadrature.cpp \
 Solver.cpp \
 Surface.cpp \
 Timer.cpp \
@@ -44,6 +44,7 @@ gradients/one-directional/one-directional-gradient.cpp \
 gradients/two-directional/two-directional-gradient.cpp \
 homogeneous/homogeneous-one-group.cpp \
 c5g7/c5g7.cpp \
+test_quad/test_quad.cpp \
 c5g7/c5g7-cmfd.cpp
 
 #===============================================================================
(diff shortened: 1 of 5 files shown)


profile/Makefile
```diff
@@ -13,22 +13,22 @@ PRECISION   = single
 # Source Code List
 #===============================================================================
 
-case = c5g7/c5g7-cmfd.cpp
+case = test_quad/test_quad.cpp
```
geogunow: Change this back to c5g7

profile/Makefile
```diff
@@ -13,22 +13,22 @@ PRECISION   = single
...
 
 source = \
-Cell.cpp \
+Quadrature.cpp \
+log.cpp 
+#Cell.cpp \
```
geogunow: Remove these comments

src/Quadrature.cpp
```diff
@@ -1184,6 +1110,171 @@ void GLPolarQuad::precomputeWeights(bool solve_3D) {
...
+
+
+/**
+ * @brief    calculates the weights to be used in Gauss-Legendre Quadrature
+ * @param    n the order of the Legendre Polynomial
+ * @param    roots a vector containing the roots of the Legendre polynomial
+ * @return   a vector of weights matched by index to the vector of roots
+ */
+std::vector <double> GLPolarQuad::getGLWeights(std::vector <double> roots,
+    int n) {
```
geogunow: alight ``int`` with ``std::vector``

src/Quadrature.h
```diff
@@ -178,12 +182,20 @@ class LeonardPolarQuad: public Quadrature {
 class GLPolarQuad: public Quadrature {
 
 private:
+  /** the roots to the Legendre polynomial of degree _num_polar */
+  std::vector <double> _roots;
```
geogunow: Why does this class variable need to be added?

ljeure: The roots are needed to calculate the weights in getGLWeights

geogunow: I see what you're saying - but it seems like we can get the roots direction from the ``_thetas`` array. Also does this PR incorporate the weight correction?

ljeure: I don't think we can get the exacat Gauss Legendre abscissae from _thetas since the thetas are adajusted. 
It does incorporate weight correction. If we only use the corrected weights, (which converge k-eff better and likely yield a more accurate answer) then there's no need for the class variable

geogunow: If we can prove that correcting the weights is advantageous there's no reason to keep the other option around. But let's put off merging that until we convince Ben that is the case.

geogunow:
Do we know why the checks are failing? The output seems to imply that all tests passed. Let's try running these tests locally.

src/Quadrature.cpp
```diff
@@ -1184,6 +1110,171 @@ void GLPolarQuad::precomputeWeights(bool solve_3D) {
...
+
+  /* use the Alberth-Housholder_n method to nudge guesses towards roots */
+  while (not all_roots_converged) {
+    
+    /* set S tildes */
+    for (int i=0; i < (n+1)/2; ++i) {
+      if (not converged[i]) {
+        double sum1 = 0;
+        double sum2 = 0;
+        for (int j=0; j<= (n+1)/2; ++j) {
```
geogunow: There is an inconsistency here. Your vectors were only made to of size `(n+1)/2`. Here you go to element `(n+1)/2` (ie assuming that the vector has size `(n+1)/2+1`. This is the reason why your tests are failing. It's picking up a random value in that last position. By the nature of how `std::vector` works, that value is often zero so you weren't seeing the issue. But that value is not guaranteed to be zero. And in the test being run, that memory is being reused so it essentially picks up a random value. If you access with `.at(...)` instead of `[]` the bounds error will show up right away.

sample-input/pin-cell/pin-cell.py
```diff
@@ -83,8 +83,7 @@
 
 openmoc.log.py_printf('NORMAL', 'Initializing the track generator...')
 
-track_generator = openmoc.TrackGenerator(geometry, opts.num_azim,
-                                         opts.azim_spacing)
+track_generator = openmoc.TrackGenerator(geometry, 128, .05)
```
geogunow: Let's change this back to the default

src/Quadrature.cpp
```diff
@@ -503,6 +503,7 @@ void Quadrature::setPolarWeights(FP_PRECISION* weights,
                " in each octant",
                num_azim_times_polar, _num_polar/2, _num_azim/4);
 
+
```
geogunow: And remove this line

geogunow:
@ljeure This PR is still failing tests. I believe this is because it's not converging the legendre roots (there's a while loop instead of a for loop so it can go on indefinitely). Can you look into it and a) make sure that is indeed the problem and b) come up with a solution? I think if you refer to what was working before, that should help. You might need to have that extra zero in the array to make it work.

ljeure:
@geogunow definitely! Where can I find the test_quadrature.py file? The input file that's failing

geogunow:
It's in tests/test_quadrature/

PR #320, opened 2016-08-16
Fix for performance problem in transport sweeps

geogunow:
This PR addresses a problem that caused the transport sweeps to be very slow after the TraverseTracks abstraction update. Previously the number of groups was calculated by looping through all materials in order to allocate a temporary array. This would be computed for every Track during the transport sweep. Now that array has been made a class variable so the temporary allocation only occurs once per sweep.


wbinventor:
Thanks for the quick fix @geogunow!

PR #321, opened 2016-08-18
An assortment of small fixes and optimizations

wbinventor:
This short PR adds an assortment of small fixes and optimizations that I've made to make it possible to model the fully detailed 2D BEAVRS core with 10,000s of materials. These include:

* Improve performance of `load_openmc_mgxs_lib(...)` by caching the number of materials in the geometry
* Improve performance of `get_openmoc_geometry(...)` from an OpenCG geometry by eliminating compatibility checks unless necessary
* Optimized and parallelized CMFD centroid generation
* Added a new `Material::copy(Material* material)` method to make it easy to design algorithms to populate large geometries with lots of different materials
* Added an optional user-defined track filename suffix to make it easy to save/load track files for different geometries without them being overwritten

### DIFF
--- a/openmoc/opencg_compatible.py
+++ b/openmoc/opencg_compatible.py
@@ -992,13 +992,17 @@ def get_opencg_geometry(openmoc_geometry):
     return opencg_geometry
 
 
-def get_openmoc_geometry(opencg_geometry):
+def get_openmoc_geometry(opencg_geometry, compatible=True):
     """Return an OpenMOC geometry corresponding to an OpenCG geometry.
 
     Parameters
     ----------
     opencg_geometry : opencg.Geometry
         OpenCG geometry
+    compatible : bool
+        Whether the OpenCG geometry is compatible with OpenMOC's geometric
+        primitives. This should be set to False if the OpenCG geometry
+        uses SquarePrism surfaces. True by default as an optimization.
 
     Returns
     -------
@@ -1018,9 +1022,6 @@ def get_openmoc_geometry(opencg_geometry):
 
     print('starting OpenCG-to-OpenMOC conversion #2')
 
-    # Update Cell bounding boxes in Geometry
-#    opencg_geometry.update_bounding_boxes()
-
     # Clear dictionaries and auto-generated IDs
     OPENMOC_MATERIALS.clear()
     OPENCG_MATERIALS.clear()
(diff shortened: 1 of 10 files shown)


openmoc/opencg_compatible.py
```diff
@@ -1014,9 +1018,6 @@ def get_openmoc_geometry(opencg_geometry):
     opencg_geometry.assign_auto_ids()
     opencg_geometry = copy.deepcopy(opencg_geometry)
 
-    # Update Cell bounding boxes in Geometry
-    opencg_geometry.update_bounding_boxes()
```
samuelshaner: Why was this removed?

wbinventor: To be honest, I don't recall why I even had it in here, and I don't see any reason to perform this calculation before building an OpenMOC geometry. It was creating a bottleneck for a version of the BEAVRS core with over 50,000+ each of materials, cells and universes that I didn't want to wait on, so I removed it. Do you need it?

src/Material.cpp
```diff
@@ -1039,6 +1039,37 @@ Material* Material::clone() {
 
 
 /**
+ * @brief Copy another Material's cross section data into this Material.
+ * @details This method performs a deep copy of the other Material's
+ *          cross section data arrays for use by this Material.
+ * @param material a pointer to the Material with data to copy
+ */
+void Material::copy(Material* material) {
```
samuelshaner: Since this method doesn't return a copy (what one might expect from a copy method) but instead copies *from* the input material to the current material, I would recommend naming this method `copyFrom(...)`?

wbinventor:
By the way, I removed the optimizations / parallelization that I put into `Cmfd::generateKNearest()` since I discovered that they resulted in a segmentation fault when compiled in double precision.

openmoc/opencg_compatible.py
samuelshaner: Nope, just curious if it was removed by mistake.

src/TrackGenerator.h
```diff
@@ -79,6 +79,9 @@ class TrackGenerator {
   /** Filename for the *.tracks input / output file */
   std::string _tracks_filename;
 
+  /** Filename suffix for the *.tracks input / output file */
+  std::string _tracks_filename_suffix;
```
samuelshaner: Do you just have this so you can put in custom names for track files and run different geometries with the same parameters and not overwrite the tracks file?

samuelshaner:
This looks good to go once the items are addressed.

src/TrackGenerator.h
wbinventor: Yes - if I want to model the different clustered versions of the same geometry, with the same CMFD mesh and track laydown, I want the track files to be stored separately so I can actually take advantage of them for subsequent simulations. Perhaps even more important, I've noticed issues when running multiple simulations which write/load tracks files at the same time - it can cause one or more simulations to just hang.

src/Material.cpp
wbinventor: Agreed, I like this name change and just pushed it.

PR #322, opened 2016-08-22
Fixed bug in CMFD for geometries not centered at the origin

samuelshaner:
This PR fixes a bug in computing the k-nearest weights where the equation in Cmfd::getDistanceToCentroid(...) assumes the geometry is centered at the origin. Now, the min-x and min-y values of the geometry are given to CMFD so the proper weights can be computed.

### DIFF
--- a/src/Cmfd.cpp
+++ b/src/Cmfd.cpp
@@ -19,6 +19,8 @@ Cmfd::Cmfd() {
   _num_y = 1;
   _width_x = 0.;
   _width_y = 0.;
+  _x_min = 0.;
+  _y_min = 0.;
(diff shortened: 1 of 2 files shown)


wbinventor:
Looking good @samuelshaner! Thanks f or putting this fix in so quickly. With this patch I've been able to successfully converge the BEAVRS problem with 31 iterations with a quarter pin-wise CMFD mesh for a variety of MGXS energy group structures and spatial homogenization schemes. We are getting close!

See the pin-wise energy-integrated fission rates and U-238 capture rates below for some eye candy:

**Fission**


**U-238 Capture**

PR #323, opened 2016-08-24
Fix and documentation for Krylov subspace methods

cjosey:
This pull request makes 4 changes. 
1. It fixes the memory allocation issues with the Krylov solver.
2. It ensures that the Krylov solver outputs plots.
3. It extracts the operator initialize tasks into a separate function, so that users can play with the operators themselves.
4. It adds documentation on how the method was derived.

The documentation is fairly terse and just kinda tacked on into the middle of nowhere, so any comments would be great.

### DIFF
--- a/docs/source/methods/index.rst
+++ b/docs/source/methods/index.rst
@@ -14,3 +14,4 @@ Theory and Methodology
     constructive_solid_geometry
     track_generation
     cmfd
+    krylov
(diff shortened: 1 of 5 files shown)


wbinventor:
Thanks for the fix and writeup @cjosey! Everything looks good to me. Just to confirm, did you build the documentation and see that your writeup shows up as expected?

cjosey:
Yes, I built the documentation and it appeared as anticipated.

PR #328, opened 2017-03-21
Complex cell regions

wbinventor:
This PR adds in support for complex cell regions like those supported in OpenMC (introduced in mit-crpg/openmc#463). Currently, OpenMOC only supports the intersection of one or more surface halfspaces to define a `Cell`'s spatial domain. This PR makes it possible to recursively define arbitrarily complex intersections, unions and complements of surface halfspaces to define a `Cell`'s spatial domain.

In particular, this PR implements the following new classes:

* `Region` (abstract)
* `Intersection` (subclass of `Region`)
* `Union` (subclass of `Region`)
* `Complement` (subclass of `Region`)
* `Halfspace` (subclass of `Intersection`)
* `RectangularPrism` (subclass of `Intersection`)

The `Cell` class now has a `_region` attribute which can be any one of the classes which implement the abstract `Region` interface. A user may either create a `Region` directly and assign it to a `Cell`, and/or add surface halfspaces using the original `Cell::addSurface(...)` method (which will populate an `Intersection` object under-the-hood). In addition, the `Region` class now implements the functionality to find min/max x, y or z bounds and BCs, and to query whether a `Point` is in the region (as was previously implemented directly by the `Cell` class).

This PR makes it possible once again to model the BEAVRS benchmark constructed by the mit-crpg/PWR_benchmarks BEAVRS builder script, which uses unions and complements to model grid spacers and assembly sleeves (introduced in mit-crpg/PWR_benchmarks#49). I will open a complementary PR to update the `openmc.openmoc_compatible` module to support complex cell region shortly.

@geogunow @samuelshaner let me know if either of you want or have the time to review this. If not, I'll review it myself and merge it in.


src/Universe.cpp
```diff
@@ -610,6 +648,24 @@ void Universe::removeCell(Cell* cell) {
...
+ * @returns
+ */
+bool Universe::containsPoint(Point* point) {
+
+  /* Loop over all Cells */
+  std::map<int, Cell*>::iterator iter;
+  for (iter = _cells.begin(); iter != _cells.end(); ++iter) {
+    if (!iter->second->containsPoint(point))
+      return false;
+  }
```
samuelshaner: It seems to me like this should be the other way around - when looping over the cells in a universe check to see if any of them contain the point. If one does, then return true. If none are found to contain the point, return false.

wbinventor:
Thanks for the swift review @samuelshaner - as you can see, I still need to add in doxygen comments, as well as new documentation for this feature. I also want to give more consideration to the SWIG memory ownership rules for the new `Region` subclasses to avoid memory leaks. I may have opened this PR a little prematurely after *finally* getting BEAVRS to work with OpenMOC again.

src/Cell.cpp
```diff
@@ -764,6 +615,16 @@ void Cell::setName(const char* name) {
 
 
 /**
+ * @brief Sets the name of the Cell
+ * @param name the Cell name string
+ */
+void Cell::setRegion(Region* region) {
+  // FIXME: this is a memory leak
+  _region = region;
```
wbinventor: Yeah, I'm really not sure how to deal with this, aside from specifying (and documenting) a rule that a `Region` can be assigned to one and only one `Cell`.

wbinventor: An alternative option might be to clone any `Region` assigned to a `Cell` and leave the Python garbage collector to clean up any `Region`s instantiated in Python.

wbinventor: I've opted for the latter option. The one downside of this is that the `Region` object given to a `Cell` cannot be change or manipulated after the assignment, since any changes made to it will not be reflected in the copied `Region` stored by the `Cell`. I've documented this though.

src/Universe.cpp
wbinventor: Oh yes of course, I'm not sure what I was thinking!

wbinventor:
I've updated this PR with the following:

* Wrote doxygen comments for all new classes and methods
* Wrote Sphinx documentation on complex regions
* Added "grid-spacer" example input file to showcase
* Revised many example inputs to use `RectangularPrism` class for boundaries
* Added "test_complex_regions" to simultaneously test `Complement`, `Intersection` and `Union` classes 

This PR is ready for review - I have no further planned changes to make. Let me know @samuelshaner @geogunow if you would prefer I just deal with this on my own.

geogunow:
Is this still backwards compatible with old inputs?

openmoc/thisown.i
```diff
@@ -2,10 +2,44 @@
 
 %module thisown
 
-/* A Cell owns the memory for each Surface it contains */
-%pythonappend Cell::addSurface %{
+/* A Region owns the memory for each Region it contains */
```
geogunow: Should this be Cell?

sample-input/simple-lattice/simple-lattice.py
```diff
@@ -24,18 +24,13 @@
 
 openmoc.log.py_printf('NORMAL', 'Creating surfaces...')
 
-left = openmoc.XPlane(x=-2.0, name='left')
-right = openmoc.XPlane(x=2.0, name='right')
-bottom = openmoc.YPlane(y=-2.0, name='bottom')
-top = openmoc.YPlane(y=2.0, name='top')
-boundaries = [left, right, bottom, top]
+boundary = openmoc.RectangularPrism(4., 4.)
+boundary.setBoundaryType(openmoc.PERIODIC)
```
geogunow: Does this mean we would no longer support bounding boxes with different boundary types?

src/Cell.cpp
```diff
@@ -353,146 +359,74 @@ int Cell::getNumSectors() {
...
-    surface = iter->second->_surface;
-    halfspace = iter->second->_halfspace;
-    min_x = std::max(min_x, surface->getMinX(halfspace));
-  }
-
-  return min_x;
+  if (_region == NULL)
+    return -INFINITY;
+  else
+    return _region->getMinX();
```
geogunow: Oh no. This might be incompatible with ClosedMOC.

src/Cell.cpp
```diff
@@ -958,71 +821,58 @@ void Cell::setParent(Cell* parent) {
...
- * @param surface a pointer to the Surface to remove
- */
-void Cell::removeSurface(Surface* surface) {
-
-  if (_surfaces.find(surface->getId()) != _surfaces.end()) {
-    delete _surfaces[surface->getId()];
-    _surfaces.erase(surface->getId());
+  /* Assign the Halfspace as the Cell's Region if it has none */
+  if (_region == NULL)
+    _region = new_halfspace;
```
geogunow: Why is a region a halfspace?

wbinventor:
@geogunow the changes made here are still compatible with old inputs. Has it been decided that ClosedMOC will indeed "bulldoze" what we have here? Do you plan to update all of the documentation on the website, integrate the test suite, etc.? If not, then it might be worthwhile to consider alternative plans (let's chat offline).

wbinventor:
@geogunow might you consider merging this in here to support others in the group who could use complex cells in 2D MOC calculations? If you plan to bulldoze OpenMOC with your closed 3D version later, it won't even matter if the two versions of the code are consistent, so I suggest merging this in to aid those who may continue to use this version in the near to intermediate term.

geogunow:
Why are the checks still failing?

wbinventor:
I'm not sure, the new test passes on my machine. I'll spend the time to debug it if there is a plan to merge this in.

wbinventor:
So what's the word on this @geogunow?

geogunow:
@wbinventor I'm fine merging this in once the checks pass

wbinventor:
Alright I will dig into what is holding up the test now. We can talk about bulldozing later.

wbinventor:
I couldn't reproduce the test failure on four different machines, so I instead removed the complex region test. This test was essentially the same as the new "sample-input/grid-spacer" example file, which uses `Complement` and `Union` objects to simply represent grid spacer around a fuel pin cell.

PR #335, opened 2017-06-28
Fix for true division in Python 3

tjlaboss:
`plotter.py` occasionally uses division to calculate values which need to be integers. The division operator performs integer division in Python 2, but true division in Python 3.

This PR ensures that three of these quotients are integers. There may be other instances, but these are the ones I encountered running the test suite.


samuelshaner:
@tjlaboss - thanks for this PR on this issue! Instead of typecasting I feel it might be cleaner to just use floor division (`//`) in the locations you identified. Could you make that edit?

tjlaboss:
Done.

samuelshaner:
Thanks! As soon as the tests finish I'll merge it in.

djsegal:
sorry for being nosy :P 

you could have just squashed all three commits into one using:

+ `git rebase HEAD~10 -i`

PR #336, opened 2017-06-28
fixed issue in Cell::clone() for trying to clone a NULL region

samuelshaner:


tjlaboss:
This appears to have solved my SEGFAULT issue when generating tracks on the TREAT geometry.

PR #337, opened 2017-07-30
Support for reading OpenMC MGXS Libraries which use anisotropic scattering matrices

nelsonag:
This simple PR allows OpenMOC to read OpenMC MGXS Libraries which include scattering matrices produced using anisotropic data.  Specifically, only the 0th order is obtained from the library.

This method does not use the higher order data to perform transport c orrection, it simply uses the 0th order data as is.


nelsonag:
Any interest in this noe?

geogunow:
Looks good to me

PR #391, opened 2019-02-20
fix compilation on python>3.5

gridley:
Hi,

#390 found that compiles fail on python 3.6. Apparently @wuwenbin2006 had similar issues. It seems that Pool.imap causes a deadlock, but the ever so slightly slower Pool.map does not. These are used in setup.py for parallel compilation. Maybe this is worth reporting to the python bug trackers.

Oddly enough, I'm still experiencing unit test failures with subtle differences in k_eff. For instance on test_1d_gradient I get results that are 2 pcm higher than the reference case. Similar small discrepancies exist around the sixth decimal place on other unit tests like test_tracking_pin_cell. Sounds like single floats are being used somewhere where they used to not, or vice-versa.

Anyways, it compiles now!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -317,7 +317,7 @@ def parallel_compile(self, sources, output_dir=None, macros=None,
       self._compile(obj, src, ext, cc_args, extra_postargs, pp_opts)
 
   # Convert thread mapping to C/C++/CUDA objects to a list and return
-  list(pool.ThreadPool(num_cpus).imap(_single_compile, objects))
+  list(pool.ThreadPool(num_cpus).map(_single_compile, objects))
   return objects
 
 


wuwenbin2006:
Good catch. Are you working on develop branch? I and Guil came across these same problems on 3D-MOC branch. The cause is that the track is parallel with a division edge between meshes. 
The solution is changing the MOC parameter a little bit, like in this commit 29168d5b5dfb0ae7ab2ac7daed61bb08305cfaa6

GiudGiud:
Great that you found the bug ! 
Another possible cause is from the way you are running the tests
Are you running them one by one or using run_tests.py ?
If you are running them one by one, you need to make sure you compiled the code in double precision (single is default), using the flag --fp=double.

gridley:
Ah, if that’s the default, then that’s for sure the issue. Thanks!

Gavin Ridley

PR #440, opened 2019-09-18
Fix CUDA code

gridley:
Greetings,

Here is a start on getting gloriously speedy CUDA to work once again. Under this commit, I have updated several method names referred to within the GPUSolver class, which have as of three years ago been changed in the parent class, Solver. Currently, actually running the code causes a hang once the simulation gets to the eigenvalue calculation iterations. Working on fixing that! I went ahead and made the PR in case I'm making any glaring errors.


src/accel/cuda/GPUSolver.cu
```diff
@@ -715,6 +715,10 @@ GPUSolver::GPUSolver(TrackGenerator* track_generator) :
   _materials = NULL;
   _dev_tracks = NULL;
   _FSR_materials = NULL;
+  _dev_chi_spectrum_material = NULL;
+
+  // TODO, take chi_spectrum_material from solver, and translate it
+  // to a pointer on the GPU. But it might not be copied there yet?
```
GiudGiud: Dont worry about chi spectrum too much, uniform starting guess is fine

src/accel/cuda/GPUSolver.cu
```diff
@@ -1033,34 +1037,46 @@ void GPUSolver::copyQuadrature() {
...
+  // unsigned outer_size = sinThetas.size();
+  // unsigned inner_size = sinThetas[0].size();
+  // Make an array of FP_PRECISION, which is what the old CUDA code took.
+  // It may be possible to just pass in a pointer to the first element of
+  // getSinThetas()... TODO
+  //   FP_PRECISION** sin_theta_array = (FP_PRECISION**)malloc(outer_size * sizeof(FP_PRECISION*));
+  //   for (unsigned j=0; j<outer_size; ++j)
+  //     sin_theta_array[j] = (FP_PRECISION*)malloc(inner_size * sizeof(FP_PRECISION));
+  //   
+  cudaMemcpyToSymbol(sin_thetas, (void*)&(_quad->getSinThetas()[0]),
```
GiudGiud: _sin_thetas for the array name since it's global ?

src/accel/cuda/GPUSolver.cu
```diff
@@ -1227,11 +1243,11 @@ void GPUSolver::initializeTracks() {
       clone_track(_tracks[i], &_dev_tracks[i], _material_IDs_to_indices);
 
       /* Get indices to next tracks along "forward" and "reverse" directions */
-      index = _tracks[i]->getTrackIn()->getUid();
+      index = _tracks[i]->getTrackNextBwd();
```
GiudGiud: Flipped bwd and fwd I think. 
We should rename _track_in and _track_out with _track_fwd and _track_bwd or something the other solvers use.
What do you think ?

src/accel/cuda/GPUSolver.cu
```diff
@@ -1831,8 +1874,9 @@ double GPUSolver::computeResidual(residualType res_type) {
  *                      passed in as a NumPy array from Python)
  * @param num_FSRs the number of FSRs passed in from Python
  */
-void GPUSolver::computeFSRFissionRates(double* fission_rates, int num_FSRs) {
+void GPUSolver::computeFSRFissionRates(double* fission_rates, long num_FSRs, bool nu) {
 
+  /* TODO what the heck is nu for? */
```
GiudGiud: see docstring in CPUSolver

src/accel/cuda/clone.cu
```diff
@@ -91,11 +91,13 @@ void clone_track(Track* track_h, dev_track* track_d,
...
-  new_track._next_out = track_h->isNextOut();
-  new_track._transfer_flux_in = track_h->getTransferFluxIn();
-  new_track._transfer_flux_out = track_h->getTransferFluxOut();
+  new_track._azim_angle_index = track_h->getAzimIndex();
+
+  // TODO unsure on these things:
+  new_track._next_in = track_h->getNextFwdFwd();
+  new_track._next_out = track_h->getNextBwdFwd();
+  new_track._transfer_flux_in = track_h->getBCFwd();
+  new_track._transfer_flux_out = track_h->getBCBwd();
```
GiudGiud: transfer_flux_in/out are supposed to be booleans to indicate whether the flux should be transferred to the next track

Basically they're NOT(BC==VACUUM).
I think what you wrote works (VACUUM = 0)

src/accel/cuda/GPUSolver.cu
```diff
@@ -397,24 +397,24 @@ __global__ void computeFSRScatterSourcesOnDevice(int* FSR_materials,
...
+__global__ void flattenFSRFluxesChiSpectrumOnDevice(dev_material* chi_material,
+                                                    FP_PRECISION* scalar_flux) {
+  /* NOTE, if this needs to go faster, copy Chi spectrum to each
+     thread block, then fill fluxes from there. Now, memory contention
+     may arise when many threads access elements of chi here at the
+     ~same~ time. */
 
-  return __longlong_as_double(old);
+  int tid = threadIdx.x + blockIdx.x * blockDim.x;
+  while (tid < *num_FSRs) {
```
GiudGiud: num_FSRs

sample-input/benchmarks/c5g7/c5g7-2d-gpu.py
```diff
@@ -0,0 +1,100 @@
...
+
+cells['Root'].setFill(lattices['Root'])
+
+###############################################################################
+##########################     Creating Cmfd mesh    ##########################
+###############################################################################
+
+log.py_printf('NORMAL', 'Creating Cmfd mesh...')
+
+cmfd = openmoc.Cmfd()
```
GiudGiud: Is there routines to tally Cmfd currents in GPUSolver ?

GiudGiud (commented):
Thanks for the PR! Looks like the structures have diverged quite a bit since then
Good luck with getting the rest of the solver to work

src/accel/cuda/GPUSolver.cu
gridley: Wait, you're preceding global variables with underscores? The other variables in constant memory also have no preceding underscore, which are at the top of this file.

src/accel/cuda/GPUSolver.cu
gridley: Ah, thanks! Guess I should have read the documentation rather than just the code xD

gridley: OK, this should do what it's supposed to now.

src/accel/cuda/GPUSolver.cu
gridley: Are you sure here? num_FSRs is a length 1 array here so that it could be placed in GPU constant memory, which is why you access it like a pointer. Although, with this tiny example program, it seems that maybe in Will Boyd's day, __constant__ memory had to consist of arrays or pointers but now it can be values? I'm going to try changing these all to values rather than length one arrays.

```cuda
__constant__ unsigned asdf = 5;

__global__ void printSecretMessage()
{
  printf("Waddup %i\n", asdf);
}

int main()
{
  printSecretMessage<<<1,32>>>();
  cudaDeviceSynchronize();
}
```

sample-input/benchmarks/c5g7/c5g7-2d-gpu.py
gridley: Uh, definitely not. Removing that...

src/accel/cuda/GPUSolver.cu
GiudGiud: oh I meant "global" as in a class attribute as opposed to a local function-scope variable

src/accel/cuda/GPUSolver.cu
GiudGiud: Well if it's an array you're right!
But let's make it a value.

gridley: Your wish is my command! (done) I imagine this has to do with some old version of CUDA. I checked cudaMemcpyToSymbol with plain old values and it seems to work fine.

gridley:
So, it seems that the main thing holding this back from working right now is Thrust not doing its job, see [this](https://groups.google.com/forum/#!topic/thrust-users/LneZ66OZ_yg) post I made to their mailing list for an explanation. I'm thinking about just implementing our own device_vector data structure since it's not crazy hard...

config.py
```diff
@@ -380,7 +380,6 @@ def setup_extension_modules(self):
         if self.debug_mode:
             for k in self.compiler_flags:
                 self.compiler_flags[k].append('-g')
-                self.compiler_flags[k].append('-fno-omit-frame-pointer')
```
GiudGiud, line 384: It's not a nvcc flag, it's a compiler flag to help with debugging large functions. I read about it there : [link](https://stackoverflow.com/questions/14666665/trying-to-understand-gcc-option-fomit-frame-pointer
) and decided to add it.

src/accel/cuda/GPUSolver.cu
```diff
@@ -663,28 +662,36 @@ __global__ void computeFSRFissionRatesOnDevice(FP_PRECISION* FSR_volumes,
                                                int* FSR_materials,
                                                dev_material* materials,
                                                FP_PRECISION* scalar_flux,
-                                               FP_PRECISION* fission) {
+                                               FP_PRECISION* fission,
+                                               bool nu = false) {
```
GiudGiud: add nu to the docstring too

src/accel/cuda/GPUSolver.cu
```diff
@@ -1185,6 +1193,8 @@ void GPUSolver::initializeMaterials(solverMode mode) {
   log_printf(INFO, "Initializing materials on the GPU...");
 
   /* Copy the number of energy groups to constant memory on the GPU */
+  printf("LOOK HERE <----------------------------\n");
+  printf("num groups on host = %i\n", _num_groups);
```
GiudGiud: remove coments

src/accel/cuda/GPUSolver.cu
```diff
@@ -1487,6 +1496,8 @@ void GPUSolver::computeFSRSources(int iteration) {
  */
 void GPUSolver::computeFSRFissionSources() {
 
+  printf("compute FSR fission sources\n");
```
GiudGiud: use log_printf

src/accel/cuda/GPUSolver.cu
```diff
@@ -1504,6 +1515,8 @@ void GPUSolver::computeFSRFissionSources() {
  */
 void GPUSolver::computeFSRScatterSources() {
 
+  printf("compute fsr scatter sources\n");
```
GiudGiud: log_printf

src/accel/cuda/GPUSolver.h
```diff
@@ -123,23 +123,23 @@ class GPUSolver : public Solver {
    * @return the number of threads per block
    */
   int getNumThreadsPerBlock();
-  FP_PRECISION getFSRSource(int fsr_id, int group);
-  FP_PRECISION getFlux(int fsr_id, int group);
-  void getFluxes(FP_PRECISION* out_fluxes, int num_fluxes);
+  double getFSRSource(long fsr_id, int group) override;
+  double getFlux(long fsr_id, int group) override;
```
GiudGiud, line 128: any technical reason you're switching to double precision ?
It doesnt make much of a difference here

src/accel/cuda/GPUSolver.cu
```diff
@@ -1223,6 +1223,7 @@ void GPUSolver::initializeMaterials(solverMode mode) {
     }
   }
   catch(std::exception &e) {
+    log_printf(DEBUG, e.what());
```
GiudGiud, line 1238: great idea !

src/accel/cuda/GPUSolver.cu
```diff
@@ -1107,6 +1119,14 @@ void GPUSolver::initializeExpEvaluators() {
...
   cudaMemcpyToSymbol(exp_evaluator, (void*)dev_exp_evaluator,
                      sizeof(GPUExpEvaluator), 0, cudaMemcpyDeviceToDevice);
+
+}
+
+
+/**
+ * @brief Explicitly disallow construction of CMFD, for now.
+ */
+void GPUSolver::initializeCmfd() {
```
GiudGiud, line 1100: put an error log for now

src/accel/cuda/GPUSolver.cu
GiudGiud: ok done in a later commit

GiudGiud:
Looks good!
Let me know when it's ready for merging.
I ll test it for accuracy/correctness then.

src/accel/cuda/GPUSolver.cu
GiudGiud: Ok looking at the VectorizedSolver which also uses the previous format, getTrackIn is for the Bwd direction

GiudGiud: And looking at DeviceTrack.h, which disagrees,
** Index of the next Track when traveling along this Track in the "forward" direction. */
  int _track_in;

GiudGiud: looking at GPUSolver.cu line 506, you can see that _track_in is used in the backward direction.

SO
- I was wrong to ask for a change here
- deviceTrack docstrings are wrong and need to be modified

src/accel/cuda/clone.cu
GiudGiud: In and Out are flipped. In is backward. 

This whole mess shows we should change notation to something more explicit

config.py
gridley: Ah, OK, thanks for catching that. It seems that nvcc doesn't allow this option, so, I've changed the script to now only apply it if nvcc isn't used.

src/accel/cuda/GPUSolver.cu
gridley: aight

src/accel/cuda/GPUSolver.cu
gridley: done

src/accel/cuda/GPUSolver.h
gridley: Yes, actually. double now has to be used here. Because GPUSolver inherits from Solver, these have to be returning doubles, since that what Solver.h has specified. Otherwise, they won't override their parent class's methods. I had an issue with these not overriding at first, which is why I added the override keywords to give errors in case anyone changes the return value and unknowingly starts making GPUSolver call methods from Solver, which would make it not work.

gridley:
OK, cool, so, the last commit addresses all those comments. It will probably be a few more days (hopefully, do time estimates on coding ever work out?) before this works.

src/accel/cuda/clone.cu
gridley: Hm, that is some confusing notation. So, I just changed _next_in to _next_bwd and vice versa for the other. Looks far easier to understand now.

src/accel/cuda/GPUSolver.h
GiudGiud: Ok works for me! We may want to change all the solvers to return FP_PRECISION since OpenMOC should be run in single precision most of the time.

gridley:
Hey, so, does initializeSolver not get called now for some reason @GiudGiud? This is creating some issues with GPUSolver, which assumes the initializeSolver method will get called. It seems that you can attempt to run GPUSolver's computeEigenvalue without initializing it, which has confused me very much because I for some reason assumed initializeSolver would be called in the constructor.

gridley:
OK, so, anyways, now running the computeEigenvalue method in GPUSolver now actually runs and does transport sweeps, but seems to numerically blow up in a few iterations. This will take some effort to fix.

GiudGiud:
InitializeSolver is not run on the C++ side by the computeFlux/Source/Eigenvalue routines or the constructor. It was added on the Python side, it was added when Colin Josey created the Krylov solver I think, and he needed a bit more control on the solver from Python

In initializeSolver you essentially have a couple routines for each part of the solver. These are called separately in the computeFlux/Source/Eig routines.

Good luck with the bug search. I'd set track angular fluxes to 0 / do a vacuum boundary case and see if it works fine there first.

gridley:
OK, cool. So, I think the issue is in GPUExpEvaluator. In particular, it seems that in the past, there was not a matrix of ExpEvaluators in solver for each of the polar and azimuthal angles. What is the purpose of this?

GiudGiud:
A sin(theta) term comes up in the exponential terms (tau = XSt * length_2D / sin(theta)). 
In 3D MOC the polar angles are slightly adjusted for each azimuthal angle to make the tracks cyclic. 
So it's better to use a different ExpEval per azimuthal angle. 

Note that we use rational fractions not expEvaluators with interpolation tables now. Look at tallyScalarFlux in CPUSolver.cpp and the call to expF1Fractional(tau, &exp). Making that change for GPU would also work and may be easier than fixing interpolation tables.

gridley:
It is with great pleasure I present the following table for results on C5G7 with 32 azimuthal angles, and everything else set to default values from the c5g7-2d.py input. This is on my laptop with an  Intel Core i7-8750H and Nvidia Quadro P1000.

```
      Time      Keff
     -------------------
CPU  | 648 s  | 1.186703
GPU  | 35  s  | 1.186700
```

The PR is ready after review, aside from the fact that some regression test is failing about IHM multisim, and I have absolutely no clue why. By the way, turning on Cmfd seems to give my CPU solver a different result, about 100 pcm different from the case here. That's why I was unable to get GPU to match CPU.

sample-input/benchmarks/c5g7/c5g7-2d.py
```diff
@@ -4,6 +4,11 @@
 from openmoc.options import Options
 from lattices import lattices, universes, cells, surfaces
 
+use_gpu = True
```
GiudGiud: default to False

src/Solver.cpp
```diff
@@ -779,6 +780,9 @@ void Solver::initializeExpEvaluators() {
  */
 void Solver::initializeMaterials(solverMode mode) {
 
+  // Don't reinitialize.
+  if (_materials_initialized) return;
```
GiudGiud: no longer needed with proposed change, see c5g7-2d.py comments

sample-input/cosine-source/cosine-source.py
```diff
@@ -106,8 +108,13 @@
...
 # initialize the solver
-solver = openmoc.CPUSolver(track_generator)
-solver.setNumThreads(opts.num_omp_threads)
+if use_gpu:
+    from openmoc.cuda import GPUSolver
+    solver= GPUSolver(track_generator)
+else:
+    solver = openmoc.CPUSolver(track_generator)
+    solver.setNumThreads(opts.num_omp_threads)
+solver.initializeSolver(0)
```
GiudGiud, line 117: same

sample-input/benchmarks/c5g7/c5g7-2d.py
```diff
@@ -78,9 +85,14 @@
 ###########################   Running a Simulation   ##########################
 ###############################################################################
 
-solver = openmoc.CPUSolver(track_generator)
+if use_gpu:
+    solver = GPUSolver(track_generator)
+else:
+    solver = openmoc.CPUSolver(track_generator)
+solver.initializeSolver(0)
```
GiudGiud: Get rid of this line and swap line 1370 and 1371 in Solver.cpp so that FSRs are initialized before materials.

src/Solver.h
```diff
@@ -426,11 +427,14 @@ class Solver {
   /* Whether to ray-trace and propagate fluxes at the same time */
   bool _OTF_transport;
 
+  /* Whether materials have been initialized (sometimes called twice unnecessarily) */
+  bool _materials_initialized;
```
GiudGiud: remove

src/Solver.cpp
GiudGiud: also it may be what s causing the test failure for multi-simulations

src/Solver.cpp
```diff
@@ -825,6 +834,14 @@ void Solver::initializeFSRs() {
   _num_groups = _geometry->getNumEnergyGroups();
   _num_materials = _geometry->getNumMaterials();
 
+  /* Sanity check */
+  if (_num_FSRs == 0)
+    log_printf(ERROR, "0 flat source regions found!");
+  if (_num_groups == 0)
+    log_printf(ERROR, "0 energy groups found!");
+  if (_num_materials == 0)
+    log_printf(ERROR, "0 materials found!");
```
GiudGiud: Remove those, this is already checked in the call to geometry.getNumEnergyGroups()

src/accel/cuda/GPUSolver.h
```diff
@@ -82,6 +83,9 @@ class GPUSolver : public Solver {
   /** A pointer to an array of the Materials on the device */
   dev_material* _materials;
 
+  /** Pointer to chi spectrum material, on device! */
```
GiudGiud: , on device! -> on the device

src/accel/cuda/GPUSolver.cu
```diff
@@ -1,22 +1,19 @@
 #include "GPUSolver.h"
 
-/** The number of azimuthal angles */
-__constant__ int num_azim[1];
+/** The number of FSRs */
+__constant__ long num_FSRs;
```
GiudGiud, line 4: any reason all of these arent in the header ?

src/accel/cuda/GPUSolver.cu
```diff
@@ -25,11 +22,14 @@ __constant__ FP_PRECISION sin_thetas[MAX_POLAR_ANGLES_GPU];
...
-/** An GPUExpEvaluator object to compute exponentials */
-__constant__ GPUExpEvaluator exp_evaluator;
+__constant__ long tot_num_tracks;
 
+__global__ void printmateriald(dev_material* matd)
+{
+  printf("Printing the material %i\n", matd->_id);
+  for (int g=0; g<num_groups; ++g)
+    printf("    sigmaf(%i) = %f\n", g, matd->_sigma_f[g]);
+}
```
GiudGiud, line 32: use log_printf
Also this should be a class function of dev_material

src/accel/cuda/GPUSolver.cu
```diff
@@ -255,22 +254,24 @@ __global__ void computeFSRSourcesOnDevice(int* FSR_materials,
   FP_PRECISION* fiss_mat;
 
   /* Iterate over all FSRs */
-  while (tid < *num_FSRs) {
+  while (tid < num_FSRs) {
 
+    // TODO: does taking the pointer really make sense here?
+    // Why not take it by value? Shouldn't that have better cache efficiency?
```
GiudGiud: most probably negligible effect to use the value rather than the pointer.

src/accel/cuda/GPUSolver.cu
```diff
@@ -397,24 +398,24 @@ __global__ void computeFSRScatterSourcesOnDevice(int* FSR_materials,
...
+ * @param chi_material pointer to device material's chi spectrum to use
+ * @param scalar_flux the array of FSR scalar fluxes
  */
-__device__ double atomicAdd(double* address, double val) {
-
-  unsigned long long int* address_as_ull = (unsigned long long int*)address;
-  unsigned long long int old = *address_as_ull, assumed;
+__global__ void flattenFSRFluxesChiSpectrumOnDevice(dev_material* chi_material,
+                                                    FP_PRECISION* scalar_flux) {
+  /* NOTE, if this needs to go faster, copy Chi spectrum to each
```
GiudGiud: delete comment, this an initialization routine, that we dont use because taking the Chi spectrum as a first guess isnt actually a good idea (will crash your CMFD), so it wont need to go faster

src/accel/cuda/GPUSolver.cu
```diff
@@ -493,24 +494,33 @@ __device__ void transferBoundaryFlux(dev_track* curr_track,
   bool transfer_flux;
   int track_out_id;
 
+  /* TODO possible micro-optimization: go back to storing next_fwd_is_bwd
+     rather than next_fwd_is_bwd. More confusing that way, but avoids one
+     "!" operation. If it gives more than a 0.5% speed increase, I'd say
+     it's worth it. */
```
GiudGiud: A 0.5% speedup is going to be very very hard to measure
If you want to keep the comment, fix next_fwd_is_bwd that appears twice

src/accel/cuda/GPUSolver.cu
```diff
@@ -493,24 +494,33 @@ __device__ void transferBoundaryFlux(dev_track* curr_track,
...
   }
 
   FP_PRECISION* track_out_flux = &boundary_flux(track_out_id,start);
 
   /* Put Track's flux in the shared memory temporary flux array */
-  for (int p=0; p < *num_polar_2; p++)
+  /* TODO check if just writing if(transfer_flux) will be usable here and
+     not incur a performance penalty. The original intention of this code
+     must have been not to get thread divergence, but the speed should be
+     exactly the same since using an if statement just leaves threads idling. */
```
GiudGiud: Here ? Not sure what you mean

src/accel/cuda/GPUSolver.cu
```diff
@@ -658,43 +669,59 @@ __global__ void addSourceToScalarFluxOnDevice(FP_PRECISION* scalar_flux,
  * @param materials an array of the dev_material pointers
  * @param scalar_flux an array of FSR scalar fluxes
  * @param fission an array of FSR nu-fission rates
+ * @param nu whether total neutron production rate should be calculated
  */
```
GiudGiud, line 659: add computeKeff argument in param

src/accel/cuda/GPUSolver.cu
```diff
@@ -715,6 +742,10 @@ GPUSolver::GPUSolver(TrackGenerator* track_generator) :
   _materials = NULL;
   _dev_tracks = NULL;
   _FSR_materials = NULL;
+  _dev_chi_spectrum_material = NULL;
+
+  // TODO, take chi_spectrum_material from solver, and translate it
```
GiudGiud: remove

src/accel/cuda/GPUSolver.cu
```diff
@@ -658,43 +669,59 @@ __global__ void addSourceToScalarFluxOnDevice(FP_PRECISION* scalar_flux,
...
+ * @param nu whether total neutron production rate should be calculated
  */
 __global__ void computeFSRFissionRatesOnDevice(FP_PRECISION* FSR_volumes,
                                                int* FSR_materials,
                                                dev_material* materials,
                                                FP_PRECISION* scalar_flux,
-                                               FP_PRECISION* fission) {
+                                               FP_PRECISION* fission,
+                                               bool nu = false,
+                                               bool computekeff = false) {
```
GiudGiud: computekeff -> computing_fission_norm is a more explicit name

src/accel/cuda/GPUSolver.cu
```diff
@@ -1118,27 +1155,31 @@ void GPUSolver::initializeFSRs() {
     FP_PRECISION* host_FSR_volumes = _FSR_volumes;
     int* host_FSR_materials = _FSR_materials;
 
-    /* Allocate memory on device for FSR volumes and Material indices */
-    cudaMalloc((void**)&_FSR_volumes, _num_FSRs * sizeof(FP_PRECISION));
-    cudaMalloc((void**)&_FSR_materials, _num_FSRs * sizeof(int));
+    cudaMalloc(&_FSR_volumes, _num_FSRs * sizeof(FP_PRECISION));
+    getLastCudaError();
+    cudaMalloc(&_FSR_materials, _num_FSRs * sizeof(int));
+    getLastCudaError();
```
GiudGiud, line 1144: do we want to keep these getLastCudaError calls everywhere ?

src/accel/cuda/GPUSolver.cu
```diff
@@ -1164,22 +1206,35 @@ void GPUSolver::initializeFSRs() {
  */
 void GPUSolver::initializeMaterials(solverMode mode) {
 
+  /* Don't double-initialize */
+  if(_materials_initialized) return;
```
GiudGiud: remove

src/accel/cuda/GPUSolver.cu
```diff
@@ -1190,13 +1245,25 @@ void GPUSolver::initializeMaterials(solverMode mode) {
...
     for (iter=host_materials.begin(); iter != host_materials.end(); ++iter) {
       clone_material(iter->second, &_materials[material_index]);
+
+      // DBG remove, spot check materials copied correctly
+      // printf("Host material %i:\n", iter->second->getId());
+      // FP_PRECISION* sigmaff = iter->second->getSigmaF();
+      // for (int g=0; g<_num_groups; ++g)
+      //   printf("    sigmaf(%i)=%f\n", g, sigmaff[g]);
+      // printmateriald<<<1,1>>>(&_materials[material_index]);
+      // cudaDeviceSynchronize();
```
GiudGiud: remove

src/accel/cuda/GPUSolver.cu
```diff
@@ -1495,20 +1583,30 @@ void GPUSolver::transportSweep() {
   FP_PRECISION* reduced_sources =
        thrust::raw_pointer_cast(&_reduced_sources[0]);
 
+  log_printf(DEBUG, "Obtained device pointers to thrust vectors.\n");
+
   /* Initialize flux in each FSR to zero */
   flattenFSRFluxes(0.0);
 
+  log_printf(DEBUG, "Flattened FSR fluxes.");
```
GiudGiud: remove

src/accel/cuda/dev_exponential.cu
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * @brief Computes the exponential term for a optical length and polar angle.
+ * @details This method computes \f$ 1 - exp(-x) \f$
+ *          for some optical path length and polar angle. Uses the rational
+            approximation of Josey from exponentials.h, but using device constants.
+ * @param tau the optical path length (e.g., sigma_t times length)
+ * @param polar the polar angle index
+ * @return the evaluated exponential
+ */
+__device__ FP_PRECISION dev_exponential(FP_PRECISION x)
```
GiudGiud: this function call should probably be inlined. Could you please look if that gets us anything ?

src/accel/cuda/dev_exponential.cu
```diff
@@ -0,0 +1,37 @@
...
+  den = den*x + d0;
+  den = d0 / den; 
+ 
+  num = p5*x + p4;
+  num = num*x + p3;
+  num = num*x + p2;
+  num = num*x + p1;
+  num = num*x + p0;
+
+  exponential = num * den * x;
```
GiudGiud: Add comment
//TODO Redo fractional approximation for 1-exp(-x) rather than (1-exp(-x)) / x

GiudGiud: Right now, the appxomiation error is non-uniform because it is multiplied by x, so smaller error below 1 and bigger above

GiudGiud (changes requested):
Great PR!
It runs well and pretty fast. Thanks for taking care of this

I'll rework the timer to show stats for the GPU once this is merged.

GiudGiud:
For the disagreement (100pcm) with the solution using CMFD, that's because that one is typically much more converged. You want to run a very tight convergence criterion to get that level of convergence without CMFD. For this case, 1e-6 gets you 1 pcm away

src/Solver.cpp
gridley: Gee, I hope! Let's see...

sample-input/benchmarks/c5g7/c5g7-2d.py
gridley: Hm. It seems that if I remove initializeSolver when the GPUSolver is being used, the number of groups used in the calculation doesn't get communicated to the GPU. Luckily that error now gets caught in GPUSolver before havoc ensues.

src/accel/cuda/dev_exponential.cu
gridley: Ah dang, silly me for not doing that. So, slapping on a __forceinline__ takes C5G7 with 64 azimuthals from 50s to 34s total run time without plotting. Shoo wee, that is fast. I've read the nvcc compiler automatically inlines stuff, so I'm amazed it wasn't doing that before.

sample-input/cosine-source/cosine-source.py
gridley: see next comment

src/accel/cuda/dev_exponential.cu
GiudGiud: nice, 30% performance increase in 5 minutes of work, the way to go
When it's in different source files you need some link time optimization to get inlining. I dont know much about nvcc but it s probably not done by default

GiudGiud (commented):
Couple comments were folded by github in last review. Should show up now that the others were resolved.

src/accel/cuda/GPUSolver.cu
gridley: Hm, I think it would be good to have them. Does any compiler macro variable get defined when --debug is used? Then I could only enable them with ifdef DEBUG or the like.

src/accel/cuda/GPUSolver.cu
gridley: nope, that's just the way I found it. I agree they should be in the header.

gridley: Actually, it seems that for CUDA, you have to put __constant__ variables in the implementation file and not the header, otherwise you get multiple definition errors.

src/accel/cuda/GPUSolver.cu
gridley: removed!

src/accel/cuda/GPUSolver.cu
gridley: removed

src/accel/cuda/GPUSolver.cu
gridley: So, it's a question of code expressiveness and not really a big deal, so I removed the comment. It's there because you end up multiplying something by a boolean rather than just using an if statement. I think just using the boolean as is is fine, so the comment is removed.

src/accel/cuda/GPUSolver.cu
gridley: done

gridley:
Ah, OK, stuff should be fixed now.

GiudGiud (approved):

src/accel/cuda/GPUSolver.cu
GiudGiud: Ok. I agree it should work either way. Thanks for the insight though

src/accel/cuda/GPUSolver.cu
GiudGiud: Ok good to know. Thanks for checking!

src/accel/cuda/GPUSolver.cu
GiudGiud: It doesnt. We'll have to think about this. Current abstraction is using those log_printf(DEBUG)

GiudGiud (approved):
Merging! Works great, is really fast, thanks for the PR!

PR #473, opened 2021-06-23
Remove repetetive sense in CMFD that cancels out

loganharbour:
After working with @GiudGiud for a bit, we came to the conclusion that this term cancels out.

Hopefully the test suite will agree?﻿

### DIFF
--- a/src/Cmfd.cpp
+++ b/src/Cmfd.cpp
@@ -1064,8 +1064,6 @@ void Cmfd::getSurfaceDiffusionCoefficient(int cmfd_cell, int surface,
   if (global_cmfd_cell_next != -1)
     delta_next = getPerpendicularSurfaceWidth(surface, global_cmfd_cell_next);
 
-  int sense = getSense(surface);
-
   /* Correct the diffusion coefficient with Larsen's effective diffusion
    * coefficient correction factor */
   if (!_linear_source)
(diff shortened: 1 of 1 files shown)


GiudGiud:
Looks like CI suite is down. I ll just check locally

src/Cmfd.cpp
```diff
@@ -1140,23 +1138,23 @@ void Cmfd::getSurfaceDiffusionCoefficient(int cmfd_cell, int surface,
...
-    dif_surf_corr = -(sense * dif_surf * (flux_next - flux) + current)
+    dif_surf_corr = -(dif_surf * (flux_next - flux) + current)
         / (flux_next + flux);
 
     /* Flux limiting condition */
     if (_flux_limiting && _moc_iteration > 0) {
       double ratio = dif_surf_corr / dif_surf;
       if (std::abs(ratio) > 1.0) {
 
-        if (sense * current > 0.0)
```
GiudGiud, line 1154: well there s the reason sense is here
This seems to be a test to see if it's outgoing or not.

dont think we can change that line.

GiudGiud (changes requested):

src/Cmfd.cpp
loganharbour: Oh duh - yeah that makes sense. I'll add back.

src/Cmfd.cpp
```diff
@@ -1458,12 +1456,12 @@ void Cmfd::constructMatrices() {
...
-          value = (dif_surf - sense * dif_surf_corr) * delta;
+          value = (dif_surf - dif_surf_corr) * delta;
           _A->incrementValue(i, e, i, e, value);
 
           /* Set the off diagonal term */
           if (getCellNext(i, s, false, false) != -1) {
-            value = - (dif_surf + sense * dif_surf_corr) * delta;
+            value = - (dif_surf + dif_surf_corr) * delta;
             _A->incrementValue(getCellNext(i, s, false, false), e, i, e, value);
           }
```
GiudGiud, line 1466: there's another sense to remove ten lines below, it s for the off-domain term

src/Cmfd.cpp
GiudGiud: nah you're right it cancels out here too.

loganharbour:
@GiudGiud should be good to go

GiudGiud (approved):

PR #491, opened 2026-06-22
Fixups for M4 macs

GiudGiud:
Note the WIP commit introduces its own problem, we need to run:
```
install_name_tool -delete_rpath /Users/giudgl/miniforge/envs/moose-old/lib     /Users/giudgl/.local/lib/python3.11/site-packages/_openmoc.cpython-311-darwin.so
```
twice (and NOT 3 times, you'd naturally get a missing symbol) to clean up our library

other items:

- need to use swig 4, not the latest!
- need some working compilers with openmp. On mac I used the ones coming with this package

```
moose-dev                               2024.10.01       openmpi                  https://conda.software.inl.gov/public
```


config.py
```diff
@@ -451,6 +458,9 @@ def setup_extension_modules(self):
         if self.cc == 'mpicc':
             self.swig_flags += ['-DMPIx']
 
+        # Dedup rpath
```
GiudGiud, line 461: (try to)

